PR #62, opened 2026-02-04
api: add MCP server endpoint

snormore:
## Summary of Changes
- Add Model Context Protocol (MCP) server at `/api/mcp` with tools for querying ClickHouse (`execute_sql`), Neo4j (`execute_cypher`), fetching database schema (`get_schema`), and reading documentation (`read_docs`)
- Add MCP resources for schema, SQL context, and Cypher context, plus an `analyze-data` prompt for guided data analysis
- Add per-IP rate limiting (100 queries/min, burst of 20) for all database query endpoints including MCP tools
- Replace hardcoded `read_docs` page enum with regex validation in both MCP and agent tools to prevent path traversal while allowing new doc pages without code changes
- Add `.mcp.json` config for local Claude Code development

## Testing Verification
- Added comprehensive MCP handler tests covering tool initialization, SQL/Cypher execution, schema fetching, and docs reading (598 lines)
- Added rate limiter unit tests for token bucket behavior and HTTP middleware responses (104 lines)
- Connect your AI of choice to https://pr-62.data.malbeclabs.com/api/mcp and try it out ([docs](https://pr-62.data.malbeclabs.com/docs/mcp))

### DIFF
--- a/agent/pkg/workflow/v3/tools.go
+++ b/agent/pkg/workflow/v3/tools.go
@@ -78,7 +78,7 @@ var (
 			"properties": {
 				"page": {
 					"type": "string",
-					"pattern": "^[a-zA-Z0-9][a-zA-Z0-9\\- ]*$",
+					"pattern": "^[a-zA-Z0-9][a-zA-Z0-9\\-]*$",
 					"description": "The documentation page to read (e.g., 'index', 'architecture', 'setup', 'troubleshooting')"
 				}
 			},
@@ -87,7 +87,8 @@ var (
 	}
 
 	// validPageNameRegex validates documentation page names to prevent path traversal.
-	validPageNameRegex = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9\- ]*$`)
+	// Allow alphanumeric and hyphens only (docs use slug format).
+	validPageNameRegex = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9\-]*$`)
 )
 
 // DefaultTools returns the default set of tools for the v3 workflow.


ben-dz (commented):
## Findings
1. High: `doublezero://schema` reads bypass the new per-IP rate limiter. The resource handler fetches schema directly from ClickHouse without calling `CheckRateLimit`, so an MCP client can spam `resources/read` to avoid the tool-level limits applied to `get_schema`/`execute_sql`. Consider applying the same limiter to the resource handler (capture IP from the request like the tools) or add a global rate limit for `/api/mcp` reads. (`api/handlers/mcp.go:408-437`)
2. Medium: `read_docs` allows spaces in page names but builds the URL via string concatenation without URL-encoding. Requests like `"DZ RPC-Connection"` will fail with `invalid control character in URL` or 404. Use `url.PathEscape` for the page segment or normalize to the slug format you expect. (`api/handlers/mcp.go:323-332`)
3. Low: The MCP prompt always instructs clients to use `execute_cypher` and read the cypher context even when the tool is not registered (non-mainnet). This can produce invalid tool calls in dev/test environments. Consider registering the prompt/resource conditionally or tailoring the prompt based on Neo4j availability. (`api/handlers/mcp.go:32-58`, `api/handlers/mcp.go:466-498`, `api/handlers/mcp.go:500-535`)

## Questions
1. Are doc pages with spaces part of the supported set? If not, should the regex be tightened to disallow spaces and align with the actual docs slug format?

snormore (commented):
Addressed all three in d4e6d950:

1. **Rate limit bypass** — Added `CheckRateLimit` to the schema resource handler, matching the `get_schema` tool.

2. **URL encoding / spaces** — Tightened the regex to `^[a-zA-Z0-9][a-zA-Z0-9\-]*$` (removed space). Docs use slug format so spaces shouldn't be allowed.

3. **Cypher prompt on non-mainnet** — `registerAnalyzeDataPrompt` now takes `cypherAvailable bool` and only includes the cypher step when Neo4j is available.

> Are doc pages with spaces part of the supported set?

No — all docs use hyphenated slugs. The space allowance was unintentional.

snormore:
Fixed in 35bf06c9 — aligned agent `read_docs` regex to slug-only format (`^[a-zA-Z0-9][a-zA-Z0-9\-]*$`) matching the MCP handler.

ben-dz:
Codex used the MCP to get the right answer (took awhile). Build the chat interface on the site got a very wrong answer. 
image

PR #92, opened 2026-02-13
web: add 3D globe view to topology page

snormore:
## Summary of Changes
- Add 3D globe view at `/topology/globe` using `react-globe.gl` with all overlays, analysis modes, and panels
- Auto-rotating "living demo" default with latency-proportional animated links, bandwidth-scaled thickness, and vibrant color scheme
- Animation and rotation auto-pause during analysis, with manual toggles in the control bar
- Inter-metro aggregate arcs when metros are collapsed
- Light/dark theme support (day/night Earth textures, shared starfield background)
- View-specific overlay defaults that reset when switching between map, graph, and globe

## Testing Verification
- Manually verified all overlays, analysis modes, collapsed metros, view switching, and theme rendering on the globe


ben-dz:
It's a pretty low-res globe. Can we find a higher resolution globe?

PR #97, opened 2026-02-16
web: add fullscreen toggle to topology views

snormore:
## Summary of Changes
- Add a fullscreen button to the topology control bar that uses the browser Fullscreen API to expand the view to fill the entire screen
- Available on all topology views (map, graph, globe) with icon/label that updates to reflect current state

## Testing Verification
- N/A - UI-only change, manually verified fullscreen toggle works on globe view


snormore:
Oh yeah actually I need to update the PR description - I only included the fullscreen button in the globe. Do you think it's useful to have in the others too?

PR #106, opened 2026-02-18
controlcenter: add service management dashboard

thijsvanemmerik:
## Summary of Changes
- Adds `controlcenter/` — a self-contained Go HTTP server with embedded React UI for managing platform services locally, replacing running indexer/api/web in separate terminals
- Dashboard shows live service status (running/stopped/failed), uptime, and PID with per-service and bulk start/stop controls
- Logs page streams real-time output from all services via SSE with filtering by service, log level, and keyword; includes a time-bucketed activity histogram and virtualized list for performance
- Settings page configures a daily start/stop schedule; times are stored in UTC and converted to/from the browser's local timezone for display
- Scheduler uses a watermark approach (fires whenever the clock passes the scheduled time and hasn't already fired that cycle) rather than a narrow 1-minute window, making it reliable regardless of when the process starts
- CORS is restricted to localhost origins only — safe when bound to `0.0.0.0` for network access
- `BindHost` config field is validated to `127.0.0.1` or `0.0.0.0`; API log query params are capped (`limit` ≤ 10 000, `interval` 1–86 400 s); config and state files written with `0600` permissions
- Frontend log buffer is capped at 10 000 entries to prevent unbounded memory growth during long sessions
- Sidebar, nav active state, theme toggle, and logo size match the data app's visual style

## Testing Verification
- Verified service start/stop/start-all/stop-all from the dashboard
- Verified log streaming reconnects automatically after network interruption
- Verified schedule triggers services at the configured local time
- Verified `PUT /api/config` with an arbitrary `bindHost` value returns 400
- Verified `GET /api/logs?limit=999999` is capped at 10 000 results in the response


snormore:
Can you put this in `dev/controlcenter` just to make it obvious that it's for local dev?

PR #124, opened 2026-02-24
rewards: add Shapley value simulation

xentoshi:
Summary of Changes

  - Adds a rewards simulation page at /topology/rewards that computes Shapley value-based reward shares for network contributors using the network topology from ClickHouse
  - Implements three modes: simulate (full network), compare (baseline vs hypothetical modified network), and link value estimate (per-link breakdown for a single contributor)
  - Adds shapley-cli, a Rust binary wrapping the network-shapley-rs crate, which solves the LP problems for each operator coalition
  - Operators with fewer than 5 devices are collapsed into "Others" by default to keep computation tractable (2^11 to ~2^6 coalitions); full simulation is opt-in via checkbox
  - Contributor, device, and link counts in the summary table link to their respective filtered pages
  - Fixes link value estimation for contributors with non-numeric device names and links stored unidirectionally in the DB
  - Documents shapley-cli build instructions in CLAUDE.md

  Known issues / what I need feedback on

  - Demand data is currently synthetic; the query pulling real metro-to-metro traffic from fact_dz_device_interface_counters returns empty results, likely because link_pk is not populated in that table.
  - Because demands are synthetic, the simulation results are mathematically correct relative to the inputs but don't reflect actual traffic patterns yet. The numbers will change once real demand data is wired in.
  - Link value estimation can be slow for contributors with many links. On the live network it took ~10 minutes for stakefac (17 links). Contributors with more than 15 links fall back to a leave-one-out approximation, but the exact path still runs a full Shapley simulation per link. This needs optimization before it's usable interactively.
  - The collapsing logic hasn't been validated end-to-end on the live network yet. Happy to do that once the demand data issue is resolved, or if someone has a pointer on the link_pk question.
  - Open question: is collapsing small contributors into "Others" acceptable for the reward calculation, or would the team prefer exact results even if it means longer computation times?

  Testing Verification

  - Verified simulate, compare, and link value modes manually in the browser against the live network on staging
  - Unit tests for CollapseSmallOperators cover threshold boundary conditions
  - Integration tests verify Shapley output against the upstream network-shapley-rs test suite (cross-validated against Python reference implementation)
  
simulate
add-link
baseline-vs-modified
link-value


xentoshi:
Closing this for now, really appreciate @snormore and @vihu taking the time to review and give feedback earlier. 

The blocker is exact Shapley computation not scaling past a modest number of operators (one LP solve per coalition). 

Approximate/Monte Carlo values could be a way through this, though that's a call for the team, not something I'd want to decide alone. 

Pushed a final commit so it builds and tests clean before closing. 

Thanks again, and happy to pick this back up if there's ever interest in the approximate-values direction :)

PR #223, opened 2026-03-12
web: differentiate soft-drained and hard-drained with stripe patterns

martinsander00:
## Summary of Changes
- Differentiate soft-drained (light stripes) and hard-drained (dark stripes) visually in the existing status timeline instead of a separate component
- Carry `drain_status` through existing link/device history API responses instead of dedicated state-timeline endpoints
- Add drain status tracking to single-link and single-device history endpoints with carry-forward logic from `dim_dz_links_history` / `dim_dz_devices_history`
- Show "Soft Drained" / "Hard Drained" badge text on the status page link list
- Align single-link detail page status classification with the multi-link status page (use `classifyLinkStatus` and interface counter thresholds consistently)
- Remove the separate `OperationalStateTimeline` component and its API endpoints

## Testing Verification
- Inserted test drain transitions into local ClickHouse and verified stripe patterns render correctly on both the main status page and individual detail pages
- Verified single-link and multi-link status classification now produce consistent results

image
image

### DIFF
--- a/api/handlers/status.go
+++ b/api/handlers/status.go
@@ -1086,7 +1086,7 @@ func fetchStatusData(ctx context.Context) *StatusResponse {
 // Link history types for status timeline
 type LinkHourStatus struct {
 	Hour         string  `json:"hour"`
-	Status       string  `json:"status"` // "healthy", "degraded", "unhealthy", "no_data"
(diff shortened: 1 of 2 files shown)


martinsander00:
I like that. Will do

PR #284, opened 2026-03-23
web: nest Publisher Check under Multicast Groups in sidebar

### LINKED ISSUE #188: Move Multicast "Publisher Check" & "Scoreboard" underneath "Multicast Groups"
The Lake sidebar is growing unwieldily with the number of top-level items. We should start nesting new sidebar pages.

 I'd propose that "Multicast"/DZ Edge stuff goes into the "Multicast Groups", similar to how Traffic has "overview" and "interfaces" that only show when you click on Traffic, and "Performance" has "DZ vs Internet" and "Path Latency"

xentoshi:
## Summary of Changes

- Publisher Check is now nested under Multicast Groups in the sidebar, matching the pattern used by Traffic and Performance sections
- Multicast Groups expands to show Groups and Publisher Check sub-items when active
- Scoreboard remains a top-level DZ item (it is edge performance, not multicast-specific)

## Testing Verification

- Verified visually in the browser: sub-items appear correctly when navigating to Multicast Groups or Publisher Check
- Build passes with no TypeScript errors

Closes #188

### DIFF
--- a/web/src/components/sidebar.tsx
+++ b/web/src/components/sidebar.tsx
@@ -93,6 +93,7 @@ export function Sidebar() {
   const isMulticastGroupsRoute = location.pathname.startsWith('/dz/multicast-groups')
   const isPublisherCheckRoute = location.pathname === '/dz/publisher-check'
   const isScoreboardRoute = location.pathname === '/dz/edge/scoreboard'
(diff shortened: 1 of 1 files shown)


ben-dz:
Resolved by PR #433

PR #470, opened 2026-04-16
api: default shredder db to shredder_qa

Jared-dz:
## Summary of Changes
- Changes the default `shredderDB` from `shredder` to `shredder_qa` so the API points at the QA shredder dataset out of the box

## Diff Breakdown
| Category   | Files | Lines (+/-) | Net |
|------------|-------|-------------|-----|
| Core logic | 1     | +2 / -2     | 0   |

## Testing Verification
- Verified locally against ClickHouse Cloud — shreds scoreboard loads correctly with `shredder_qa` as the default database


Jared-dz:
data source stale

PR #472, opened 2026-04-16
web: fix win rate label to show normalized share on shreds scoreboard

thijsvanemmerik:
## Summary of Changes
- The win rate percentage displayed above each row's bar and in the hero card was showing the raw \`dz_edge.win_rate_pct\` value from the API, which uses a different denominator than the bar segments (which are normalized so DZ Edge + Jito + Turbine sum to 100%). This caused the label to show a significantly lower number (e.g. 56%) than what the bar and tooltip were displaying (e.g. 87%).
- Per-row label and tooltip header now use \`dz_edge.win_rate_pct\` normalized against the non-granular total (DZ Edge + Jito + Turbine), making label, bar, and tooltip consistent. The label is also stable across the granular toggle since it always uses the same non-granular basis.
- Hero bar \`winRate\` (label and gauge) is now derived from the already-normalized \`feedRatesGranular\` DZ keys instead of a simple raw average, so the hero and row figures are computed on the same basis. The now-unused raw accumulator variables (\`dzShredsWon\`, \`dzTotalShreds\`, \`edgeFirstArrival\`) are removed.

## Diff Breakdown
| Category   | Files | Lines (+/-) | Net |
|------------|-------|-------------|-----|
| Core logic | 1     | +16 / -11   |  +5 |

Pure display fix — no data, API, or test changes.

## Testing Verification
- Verified label, bar fill, and tooltip hover all show the same percentage for a given row
- Confirmed the label value is stable when toggling the granular layers button

### DIFF
--- a/web/src/components/edge-scoreboard-page.tsx
+++ b/web/src/components/edge-scoreboard-page.tsx
@@ -2168,7 +2168,6 @@ function NodeRow({ node, label, granular }: { node: EdgeScoreboardNode; label: s
   const cellRef = useRef<HTMLDivElement>(null)
   const dz = node.feeds['dz']
   const dzEdge = node.feeds['dz_edge']
-  const edgeFirstArrival = dzEdge?.win_rate_pct ?? 0
 
   // Build lead time lookup: loser_feed -> { p50, p95 }.
   // Prefer dz_edge (dz + dz_rebop combined, matches the win-rate framing);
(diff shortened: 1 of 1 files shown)


thijsvanemmerik:
Superseded by #473, which fixes the root cause in the API query rather than patching the frontend.

PR #541, opened 2026-04-28
indexer: fix isis_down false positive after device hostname renames

### LINKED ISSUE #540: indexer: isis_down false positive after device hostname renames
## Summary

Links show IS-IS down on the status page after a device hostname rename, even though IS-IS adjacencies are stable and UP on the routers. The false positive persists until the fix is deployed — after which it clears automatically within one rollup cycle (~30 seconds).

## Root cause

When a hostname is renamed, the IS-IS telemetry reconciliation writes two records at the same `snapshot_ts` for the same `link_pk`:
- `is_deleted=1` for the old hostname entity (cleanup)
- `is_deleted=0` for the new hostname entity (active)

`resolveISISAdjacency()` in `indexer/pkg/rollup/activities.go` uses a single `current bool` variable per link, overwriting it for each entry ordered by `(link_pk, snapshot_ts)`. When timestamps are equal, ClickHouse does not guarantee row order. If the deletion entry lands last, `current = true` (deleted), and every subsequent bucket inherits `isis_down = true` through carry-forward — indefinitely.

This is a latent bug that only surfaces after a hostname rename, and only if ClickHouse's internal storage order (which can shift after MergeTree merges) puts the deletion entry last.

## Fix

Switch from a single carry-forward variable to per-entity state tracking. Every link has exactly two sides reporting IS-IS telemetry (verified across all 166 links in the database). A link is only UP when **at least 2 entities** carry a non-deleted state — meaning both sides independently confirm the adjacency is healthy.

This removes the dependency on ClickHouse row ordering entirely. After a rename, 2 new UP entities exist alongside 1–2 stale deleted entities, and the count of active entities is still 2, so the link correctly shows UP.

## Why it clears automatically

`link_rollup_5m` uses `ReplacingMergeTree(ingested_at)`. The live rollup runs every 30 seconds and rewrites the last 30 minutes of buckets. An incident is only marked ongoing if the last `isis_down = true` bucket is within 15 minutes (`ended_at >= now() - 15min` in `link_incidents_v`). Once the fix is deployed, the next rollup cycle writes `isis_down = false` for all recent buckets, and ongoing incidents close on their own.

## Affected links (incident on 2026-04-27)

Triggered by hostname updates to match onchain names: `marseille` → `dz-mrs-01`, `madrid` → `dz-mad-01`, `strasbourg` → `dz-sxb-01`, `frankfurt` → `dz-fra-01`.

- `dz-mrs-01:dz-mad-01`
- `dz-fra-01:dzd-fra-01`
- `dz-sxb-01:dz-mad-01`
- `dz-fra-01:dz-sxb-01`
- `dz-fra-01:dz-muc-01`
- `dgt-dzd-mrs-mrs1:dz-mrs-01`

thijsvanemmerik:
Resolves: #540

## Summary of Changes
- Replace single carry-forward `bool` per link in `resolveISISAdjacency` with per-entity state tracking (`map[entity_id]bool`). A link is only UP when ≥ 2 entities are non-deleted — both sides of the link must independently confirm the adjacency is healthy.
- Evaluate state after each full same-timestamp group rather than after each individual row, so a rename's simultaneous `is_deleted=1` (old entity) and `is_deleted=0` (new entity) at the same `snapshot_ts` is resolved atomically.
- Add regression test `TestComputeLinkRollup_ISISHostnameRename` that seeds a simultaneous delete+create at the same timestamp and asserts the link stays UP.

## Diff Breakdown
| Category   | Files | Lines (+/-)  | Net  |
|------------|-------|--------------|------|
| Core logic |     1 | +60 / -21    |  +39 |
| Tests      |     1 | +53 / -1     |  +52 |
| **Total**  |     2 | +113 / -22   |  +91 |

Small, focused algorithm fix with targeted regression test.

<details>

- `indexer/pkg/rollup/activities.go` — rewrites `resolveISISAdjacency` to use per-entity carry-forward and timestamp-group evaluation
- `indexer/pkg/rollup/activities_test.go` — seeds both sides of the IS-IS adjacency and adds a dedicated rename regression test

</details>

## Testing Verification
- `TestComputeLinkRollup_ISISHostnameRename`: seeds a simultaneous `is_deleted=1` (old entity) + `is_deleted=0` (new entity) at the same `snapshot_ts`, confirms `isis_down=false`
- All rollup tests pass (`go test ./indexer/pkg/rollup/...`) and full test suite passes (`make test`)
- Fix self-heals automatically on deploy: `link_rollup_5m` uses `ReplacingMergeTree` and the live rollup rewrites the last 30 minutes every 30 seconds; once recent buckets get `isis_down=false`, the ongoing incident in `link_incidents_v` closes on its own

### DIFF
--- a/indexer/pkg/rollup/activities.go
+++ b/indexer/pkg/rollup/activities.go
@@ -355,10 +355,18 @@ func (a *Activities) resolveLinkState(ctx context.Context, linkPKs map[string]st
 // resolveISISAdjacency queries dim_isis_adjacencies_history to determine if each
 // (link, bucket) pair has an ISIS adjacency. For each bucket, it finds the latest
 // adjacency state as of the bucket's end time (bucket + 5min).
+//
+// A link is UP only when both sides independently confirm the adjacency is healthy
+// (activeCount >= 2). This guards against false positives from hostname renames,
+// where the old entity gets is_deleted=1 alongside a new is_deleted=0 entity for
+// the same link_pk at the same snapshot_ts. Because ClickHouse row ordering for
+// same-timestamp entries is nondeterministic, a single carry-forward variable could
+// flip to "deleted" and never recover. Per-entity tracking and evaluating state
+// after each timestamp group (not after each individual row) avoids this.
 func (a *Activities) resolveISISAdjacency(ctx context.Context, linkPKs map[string]struct{}, bucketTimes map[time.Time]struct{}, _, windowEnd time.Time, sourceDB string) (map[bucketKey]bool, error) {
-	// Query all adjacency entries up to windowEnd
+	// Include entity_id so we can carry forward state per entity independently.
 	query := `
-		SELECT link_pk, snapshot_ts, is_deleted
+		SELECT link_pk, entity_id, snapshot_ts, is_deleted
 		FROM ` + tableRef(sourceDB, "dim_isis_adjacencies_history") + `
 		WHERE snapshot_ts <= $1 AND link_pk != ''
 		ORDER BY link_pk, snapshot_ts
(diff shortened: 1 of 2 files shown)


ben-dz (commented):
I think it would be worth adding a regression test to the test pattern to make sure the simultaneous deleted/not deleted behavior stays fixed?

This also now requires two sides to be up for a link to be up, before only one of them had to be. That's probably actually a good thing, but it is a change.

thijsvanemmerik (commented):
Added a regression test for this in the latest commit: `TestComputeLinkRollup_ISISHostnameRename`.

ben-dz (approved):
Looks good to me.

PR #606, opened 2026-05-27
feat(api): add gNMI network state endpoint

vihu:
# Summary

Adds a read-only `/api/dz/network-state` endpoint that summarizes existing gNMI telemetry for the selected DoubleZero environment.

## Changes

- Add env-aware network state API handler backed by telemetry ClickHouse tables.
- Summarize telemetry freshness across interface, BGP, ISIS, and transceiver tables.
- Aggregate latest interface state by interface family, including `Switch*` fabric interfaces.
- Aggregate latest BGP neighbors and ISIS adjacencies by raw OpenConfig state.
- Add optics volume summary from latest transceiver state and threshold telemetry.
- Include known-gap flags for empty telemetry, mainnet pilot gaps, and empty ISIS overload-bit data.
- Register `GET /api/dz/network-state` route.
- Add handler tests for empty telemetry, missing telemetry schema, and populated aggregate responses.

## Validation

- `go test ./api/handlers -run 'TestGetNetworkState|TestGetDeviceOptics' -count=1`
- `go test ./api/handlers -count=1`
- `make fmt`
- `make lint`

### DIFF
--- a/api/handlers/network_state.go
+++ b/api/handlers/network_state.go
@@ -165,36 +165,56 @@ func (a *API) GetNetworkState(w http.ResponseWriter, r *http.Request) {
 }
 
 func queryNetworkFreshness(ctx context.Context, conn driver.Conn, tdb string) ([]TelemetryFreshness, error) {
-	freshness := make([]TelemetryFreshness, 0, len(networkStateFreshnessTables))
-	for _, table := range networkStateFreshnessTables {
-		query := fmt.Sprintf(`
+	parts := make([]string, 0, len(networkStateFreshnessTables))
+	for i, table := range networkStateFreshnessTables {
+		parts = append(parts, fmt.Sprintf(`
 			SELECT
+				%[1]d AS table_order,
+				'%[2]s' AS table_name,
 				count() AS rows,
 				uniqExact(device_pubkey) AS devices,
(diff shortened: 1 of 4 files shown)


api/handlers/network_state.go
```diff
@@ -0,0 +1,431 @@
...
+
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(resp); err != nil {
+		logError("network state: failed to encode response", "error", err)
+	}
+}
+
+func queryNetworkFreshness(ctx context.Context, conn driver.Conn, tdb string) ([]TelemetryFreshness, error) {
+	freshness := make([]TelemetryFreshness, 0, len(networkStateFreshnessTables))
+	for _, table := range networkStateFreshnessTables {
```
ben-dz: Could this loop make a single query or parallelize rather than so many sequential ones all inside the same context timeout within GetNetworkState?

vihu: Right, good call. Could get away with a `union all` query here I think. I'll make that change shortly.

vihu: Updated: [e23b8d6](https://github.com/malbeclabs/lake/pull/606/commits/e23b8d61c5055354b8613b8a33866bae55c81478)

ben-dz (approved):
lgtm

PR #607, opened 2026-05-27
feat(lake): add network state page

vihu:
# Summary

Adds a read-only DoubleZero Network State page backed by the existing gNMI telemetry summary endpoint, with API-side caching to make the page usable against large telemetry tables.

## Changes

- Add per-environment in-memory caching for `GET /api/dz/network-state`
- Add singleflight refresh behavior to avoid duplicate expensive ClickHouse queries
- Serve stale cached network-state data when refreshes fail
- Return `X-Cache: MISS | HIT | STALE` and vary responses by `X-DZ-Env`
- Add typed frontend API client support via `fetchNetworkState()`
- Add `/dz/network-state` route and sidebar navigation entry
- Add Network State UI for telemetry freshness, interfaces, BGP, ISIS, optics, and known gaps
- Add tests covering empty telemetry, missing tables, summaries, and per-env cache behavior

## Validation

- `go test ./api/...`
- `go test -race ./api/handlers -run 'TestGetNetworkState' -count=1`
- `golangci-lint run -c ./.golangci.yaml ./api/...`
- `cd web && bun run build`
- `cd web && bun run test:run`
- `cd web && bun run lint`
- `git diff --check`

### Devnet

image

### Testnet

image

### DIFF
--- a/api/handlers/api.go
+++ b/api/handlers/api.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"sync"
 
 	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
 	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/malbeclabs/lake/indexer/pkg/neo4j"
+	"golang.org/x/sync/singleflight"
 )
 
 var errNoPgPool = errors.New("postgres not configured")
@@ -61,6 +63,10 @@ type API struct {
 
 	// OnSlackInstallationChange is called when a Slack installation changes.
 	OnSlackInstallationChange func(teamID string)
+
+	networkStateCacheMu sync.RWMutex
+	networkStateCache   map[DZEnv]networkStateCacheEntry
+	networkStateGroup   singleflight.Group
 }
 
 // envDB returns the ClickHouse connection for the environment in the context.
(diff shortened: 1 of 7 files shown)


vihu:
Note: This is a raw telemetry diagnostic page. Future operator/contributor telemetry surfaces will use Lake `enriched_*` views.

ben-malbeclabs:
@vihu I am just wondering whether having Network State, at least with the counters shown, belongs so prominently on the website.

As mentioned yesterday, the telemetry freshness is only interesting to those supporting the gNMI stack (internal), so I am not sure if even belongs in data.malbeclabs.com at all.

Also, the counters as shown are not really useful to me, and I am not sure how we would use them.  

It seems like this belongs in a Grafana dashboard used internally for ops.  I can't see how a contributor would use what is there currently.

vihu:
> @vihu I am just wondering whether having Network State, at least with the counters shown, belongs so prominently on the website.
> 
> As mentioned yesterday, the telemetry freshness is only interesting to those supporting the gNMI stack (internal), so I am not sure if even belongs in data.malbeclabs.com at all.
> 
> Also, the counters as shown are not really useful to me, and I am not sure how we would use them.
> 
> It seems like this belongs in a Grafana dashboard used internally for ops. I can't see how a contributor would use what is there currently.

That's fair, I'll close this PR and move the counters to our grafana instead.

PR #608, opened 2026-05-28
feat(api): add controller-call status timeline

vihu:
# Summary

Adds a controller GetConfig call timeline for devices so operators can correlate device pages and topology sidebars with “device stopped calling controller” alerts.

## Changes

- Adds `GET /api/dz/devices/{pk}/controller-calls`
- Classifies controller-call state as calling, stopped, recovered, or no data
- Handles missing/inaccessible controller-call source tables without breaking the UI
- Adds a Controller calls chart to the device detail page
- Adds the same chart to the topology device details sidebar
- Adds API types/fetcher and backend tests for not found, unavailable source, stopped/recovered classification, and invalid ranges

### Devices page

image

### Topology

image

### DIFF
--- a/api/handlers/device_controller_calls.go
+++ b/api/handlers/device_controller_calls.go
@@ -106,7 +106,8 @@ func (a *API) GetDeviceControllerCalls(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	sourceDB, sourceAvailable, err := resolveControllerCallsSourceDB(ctx, conn, EnvFromContext(ctx))
+	sourceDB := TelemetryDatabaseForEnv(EnvFromContext(ctx))
+	sourceAvailable, err := controllerCallsSourceAvailable(ctx, conn, sourceDB)
 	if err != nil {
 		logError("device controller calls: source check failed", "error", err, "pk", pk, "database", sourceDB)
 		http.Error(w, dberror.UserMessage(err), http.StatusInternalServerError)
@@ -217,20 +218,6 @@ func queryDeviceControllerCallMeta(ctx context.Context, conn driver.Conn, pk str
 	err := conn.QueryRow(ctx, query, pk).Scan(&meta.Code, &meta.Status)
 	metrics.RecordClickHouseQuery("device_controller_calls_device", time.Since(start), err)
 	return meta, err
-}
-
-func resolveControllerCallsSourceDB(ctx context.Context, conn driver.Conn, env DZEnv) (string, bool, error) {
-	candidates := []string{TelemetryDatabaseForEnv(env), string(env)}
-	for _, database := range candidates {
-		available, err := controllerCallsSourceAvailable(ctx, conn, database)
-		if err != nil {
-			return database, false, err
-		}
-		if available {
-			return database, true, nil
-		}
-	}
-	return candidates[0], false, nil
 }
 
 func controllerCallsSourceAvailable(ctx context.Context, conn driver.Conn, database string) (bool, error) {
--- a/api/handlers/device_controller_calls.go
+++ b/api/handlers/device_controller_calls.go
@@ -106,8 +106,7 @@ func (a *API) GetDeviceControllerCalls(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	sourceDB := string(EnvFromContext(ctx))
-	sourceAvailable, err := controllerCallsSourceAvailable(ctx, conn, sourceDB)
+	sourceDB, sourceAvailable, err := resolveControllerCallsSourceDB(ctx, conn, EnvFromContext(ctx))
 	if err != nil {
 		logError("device controller calls: source check failed", "error", err, "pk", pk, "database", sourceDB)
 		http.Error(w, dberror.UserMessage(err), http.StatusInternalServerError)
@@ -218,6 +217,20 @@ func queryDeviceControllerCallMeta(ctx context.Context, conn driver.Conn, pk str
 	err := conn.QueryRow(ctx, query, pk).Scan(&meta.Code, &meta.Status)
 	metrics.RecordClickHouseQuery("device_controller_calls_device", time.Since(start), err)
 	return meta, err
+}
+
+func resolveControllerCallsSourceDB(ctx context.Context, conn driver.Conn, env DZEnv) (string, bool, error) {
+	candidates := []string{TelemetryDatabaseForEnv(env), string(env)}
+	for _, database := range candidates {
+		available, err := controllerCallsSourceAvailable(ctx, conn, database)
+		if err != nil {
+			return database, false, err
+		}
+		if available {
+			return database, true, nil
+		}
+	}
+	return candidates[0], false, nil
 }
 
 func controllerCallsSourceAvailable(ctx context.Context, conn driver.Conn, database string) (bool, error) {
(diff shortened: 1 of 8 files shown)


ben-malbeclabs (commented):
Three findings — the env→database wiring (#1) is the load-bearing one; the state-machine items are smaller and worth a look before merge.

api/handlers/device_controller_calls.go
```diff
@@ -0,0 +1,482 @@
...
+		bucketSecs = 300
+	}
+
+	if !sourceAvailable {
+		resp := emptyDeviceControllerCallsResponse(pk, meta, params, bucketSecs, false)
+		writeJSON(w, resp)
+		return
+	}
+
+	bucketAggs, err := queryControllerCallBucketAggs(ctx, conn, sourceDB, pk, params, bucketSecs)
```
ben-malbeclabs, line 128: **Env value is used as the database name; handler silently returns the source-unavailable payload in prod.**

`EnvFromContext(ctx)` returns the env tag (`"mainnet-beta"`, `"testnet"`, `"devnet"`), not a database name. The source-availability check on `system.tables` (line 251) then looks for a database literally named `mainnet-beta`, which doesn't exist anywhere — every prod call falls into the "source unavailable" branch and returns an empty response.

Other handlers in this repo go through `TelemetryDatabaseForEnv(env)`, which maps to `telemetry_mainnet_beta` / `telemetry_testnet` / `telemetry_devnet`. See `api/handlers/network_state.go:213` and `api/handlers/device_optics.go:93`. The tests here pass only because the setup creates a database literally named `` `mainnet-beta` ``.

Fix:
```go
sourceDB := TelemetryDatabaseForEnv(EnvFromContext(ctx))
```

Also worth confirming with the telemetry team that the upstream `controller_grpc_getconfig_success` table actually lands in `telemetry_<env>` rather than `lake_<env>` — couldn't find a writer for it anywhere in lake/, so this PR is effectively gated on whichever pipeline produces that table.

api/handlers/device_controller_calls.go
```diff
@@ -0,0 +1,482 @@
...
+				gapSeconds = &gap
+			}
+		}
+
+		lastStatus = status
+		buckets = append(buckets, DeviceControllerCallsBucket{
+			TS:               bucketStart.Format(time.RFC3339),
+			Calls:            agg.Calls,
+			MinutesWithCalls: agg.MinutesWithCalls,
+			Status:           status,
```
ben-malbeclabs, line 401: **`stoppedOpen` stays latched after the prior-history floor decays.**

Once `stoppedOpen = true`, this branch keeps firing on every empty bucket because of the `stoppedOpen ||` term — irrespective of whether `sumControllerCalls(prefixTimes, prefixCalls, bucketEnd.Add(-controllerCallHistoryWindow), bucketEnd)` still clears `controllerCallPriorHistoryMinimum`. A device that's been quiet long enough for the original history window to age out will still render as `stopped` forever instead of falling back to `no_data`.

I think the intent is "stay stopped as long as the prior-history floor is still met" — so the fix is to drop the `stoppedOpen ||` and let each empty bucket re-evaluate the floor:

```go
} else if sumControllerCalls(prefixTimes, prefixCalls, bucketEnd.Add(-controllerCallHistoryWindow), bucketEnd) > controllerCallPriorHistoryMinimum {
    status = ControllerCallStatusStopped
    stoppedOpen = true
}
```

As-is, the latched state outlives the data that justified it. Concrete failure mode: a device that was deleted or fully decommissioned >72h ago appears as `stopped` indefinitely; the operator can't tell "actually broken" from "not interesting anymore" from the timeline alone.

api/handlers/device_controller_calls.go
```diff
@@ -0,0 +1,482 @@
...
+	lo, hi := 0, len(times)
+	for lo < hi {
+		mid := int(uint(lo+hi) >> 1)
+		if times[mid].Before(target) {
+			lo = mid + 1
+		} else {
+			hi = mid
+		}
+	}
+	return lo
```
ben-malbeclabs, line 474: **`latestTimes[idx]` can be later than `before` when `before` falls mid-minute.**

`sortSearchTimes` finds the smallest minute-start `≥` target; `idx := result - 1` then gives the largest minute-start strictly less than `before`. But `latestTimes[idx]` is the *latest call observation within that minute*, which can be after `before` if `before` is sub-minute aligned. The caller computes `gap := bucketEnd.Sub(*last)` and drops it via `if gap >= 0` — so `current_gap_seconds` silently goes `nil` even when calls are happening normally.

Only triggers when bucket boundaries fall mid-minute, which the default 300s `bucket` avoids. So low-severity — but easy to harden:

```go
func latestControllerCallAt(times, latestTimes []time.Time, before time.Time) *time.Time {
    idx := sortSearchTimes(times, before) - 1
    for idx >= 0 && !latestTimes[idx].Before(before) {
        idx--
    }
    if idx < 0 {
        return nil
    }
    ts := latestTimes[idx]
    return &ts
}
```

Or: have the caller pass a tight upper bound and have this function return nil when no observation strictly precedes it. Either way, the silent `nil` from a negative gap is masking the real value.

vihu:
@ben-malbeclabs addressed your comments here: [d7188de](https://github.com/malbeclabs/lake/pull/608/commits/d7188de42e2121b09ac21531ac3664796641a7d7)

ben-malbeclabs:
> @ben-malbeclabs addressed your comments here: [d7188de](https://github.com/malbeclabs/lake/pull/608/commits/d7188de42e2121b09ac21531ac3664796641a7d7)

@vihu:

The fix works, and the two new tests are good — thanks for adding the explicit fallback path coverage.

  One concern with the fallback design itself: in prod the `controller_grpc_getconfig_success` table only ever lands in `telemetry_<env>`. If it's missing there, "source
  unavailable" is the correct answer, not "look in a different database." The fallback adds two failure modes:

  1. **Masks misconfig** — a missing `telemetry_devnet` should be loud; instead it quietly searches a parallel namespace.
  2. **Wrong-DB risk** — if anyone ever creates a database literally named `devnet` / `mainnet-beta` / `testnet` (the new test setup does this: `api.EnvDatabases["devnet"]
  = api.Database`), the handler will silently serve data from it.

vihu:
> > @ben-malbeclabs addressed your comments here: [d7188de](https://github.com/malbeclabs/lake/pull/608/commits/d7188de42e2121b09ac21531ac3664796641a7d7)
> 
> @vihu:
> 
> The fix works, and the two new tests are good — thanks for adding the explicit fallback path coverage.
> 
> One concern with the fallback design itself: in prod the `controller_grpc_getconfig_success` table only ever lands in `telemetry_<env>`. If it's missing there, "source unavailable" is the correct answer, not "look in a different database." The fallback adds two failure modes:
> 
>     1. **Masks misconfig** — a missing `telemetry_devnet` should be loud; instead it quietly searches a parallel namespace.
> 
>     2. **Wrong-DB risk** — if anyone ever creates a database literally named `devnet` / `mainnet-beta` / `testnet` (the new test setup does this: `api.EnvDatabases["devnet"]   = api.Database`), the handler will silently serve data from it.

updated [fae9ce8](https://github.com/malbeclabs/lake/pull/608/commits/fae9ce8e8df2072ae927319ae7da6aa383b88c8b)

ben-malbeclabs (approved):

PR #609, opened 2026-05-29
indexer: ingest mroute state-collect snapshots into clickhouse

ben-malbeclabs:
## Summary

- Add `lake/indexer/pkg/dz/mroute` package that parses `show ip mroute | json` snapshots written by `doublezero-telemetry --state-collect-enable` and writes them to ClickHouse as a type-2 dimension (`dim_dz_ip_mroute_entries_history` + `dz_ip_mroute_entries_current` view).
- Wire a new `SyncIPMroute` activity into the `dzingest` Temporal workflow so per-device PIM forwarding state — flags, RPF interface/neighbor, OIF list, creation time — flows into ClickHouse alongside ISIS without a separate worker.
- Mroute ingest is gated by `--mroute-enabled` / `MROUTE_ENABLED` (off by default); when enabled, the source enumerates `snapshots/ip-mroute/device=<pubkey>/` prefixes and pulls the latest snapshot per device over the previous 3 days.

## Why

This is the lake half of malbeclabs/infra#1416. doublezero/PR #3801 (already open) adds the corresponding `ip-mroute` and `ip-mroute-count` default state-collect kinds on the device side. Together they put live OIL state in ClickHouse for the multicast demo dashboard (#1428) and for any future per-(S,G) debugging.

The type-2 storage model means a steady-state cluster produces zero writes between snapshots — only OIL/RPF/flag changes per (device, vrf, mode, group, source) tuple turn into rows — so the table stays cheap to query and supports "state at time T" via the existing `GetAsOfRows` API.

## Testing Verification

- `go test ./indexer/pkg/dz/mroute/...` — parser exercises the prod-derived fixture (ME/SMP/SBNP/E/SP flag variants, missing `rpf{}` block on `Null0`, multi-element OIF, multi-source group, empty `bidirectional.groups`); source_s3 unit tests cover `snapshotsPrefix` with and without `BucketPathPrefix`, snapshot-timestamp parsing from the state-ingest filename format, and the Source interface assertions.
- `go vet ./indexer/...` clean.
- Manually walked the lake-side S3 key format against the state-ingest server (`telemetry/state-ingest/pkg/server/handler.go:302`) so the source pulls from `snapshots/ip-mroute/device=<pk>/date=YYYY-MM-DD/hour=HH/YYYYMMDDTHHMMSSZ.json` — the path the device-side PR writes.

## Notes

Closes malbeclabs/infra#1416 (paired with doublezero PR #3801, which lands the device-side state-collect kinds).

PR is ~700 lines of new non-test Go + 125 lines of migration SQL + ~120 lines of wiring, which is above the 500-line guideline; the package mirrors the ISIS one-to-one (types/parser/schema/source/store/sync split + S3 source + migration), so I kept it as a single review rather than splitting since each file is small and the shape only makes sense as a whole. Happy to split the migration out separately if the reviewer prefers.

### DIFF
--- a/indexer/pkg/dz/mroute/source_s3.go
+++ b/indexer/pkg/dz/mroute/source_s3.go
@@ -1,0 +1,233 @@
+package mroute
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+)
+
+// SnapshotKind is the state-collect kind name for mroute dumps. It matches the
+// key used by the telemetry state-ingest server when it presigns the
+// device-side PUT.
+const SnapshotKind = "ip-mroute"
+
+// S3SourceConfig configures the state-collect mroute snapshot source.
+type S3SourceConfig struct {
+	Bucket      string // S3 bucket name written to by the state-ingest server
+	Region      string // AWS region
+	KeyPrefix   string // Optional path prefix matching the state-ingest server's BucketPathPrefix
+	EndpointURL string // Optional custom endpoint (for MinIO in tests)
+
+	// LookbackDays controls how many UTC days back to scan for snapshots when
+	// finding the latest per device. Three covers clock skew and the
+	// midnight-UTC boundary; tests can shorten this.
+	LookbackDays int
+}
+
+// S3Source implements Source against an S3 bucket populated by the telemetry
+// state-ingest server. Snapshot keys follow the pattern
+//
+//	<KeyPrefix>/snapshots/<SnapshotKind>/device=<pubkey>/date=YYYY-MM-DD/hour=HH/<ts>.json
+type S3Source struct {
+	client       *s3.Client
+	bucket       string
+	keyPrefix    string // already stripped of trailing slash; empty when no prefix
+	lookbackDays int
+}
+
+// NewS3Source constructs a state-collect S3 source using the default AWS
+// credential chain (env / IAM role / profile). The bucket is expected to be
+// the same one the telemetry state-ingest server writes to.
+func NewS3Source(ctx context.Context, cfg S3SourceConfig) (*S3Source, error) {
+	if cfg.Bucket == "" {
+		return nil, fmt.Errorf("mroute: bucket is required")
+	}
+	if cfg.Region == "" {
+		cfg.Region = "us-east-1"
+	}
+	if cfg.LookbackDays <= 0 {
+		cfg.LookbackDays = 3
+	}
+
+	awsCfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(cfg.Region))
+	if err != nil {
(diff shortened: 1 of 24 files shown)


indexer/pkg/dz/mroute/source_s3.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+		body, err := s.getObject(ctx, key)
+		if err != nil {
+			return nil, err
+		}
+		dumps = append(dumps, &Dump{
+			FetchedAt:    time.Now().UTC(),
+			SnapshotTS:   snapTS,
+			DevicePubkey: pubkey,
+			RawJSON:      body,
```
vihu: I think the S3 body shape here does not match what doublezero-telemetry uploads.

`mroute.Parse` expects top-level raw Arista JSON (`{"vrfs": ...}`), and `S3Source` passes the full object body directly into `Parse`. But the state collector uploads a `StateSnapshot` envelope:

```go
snap := StateSnapshot{
    Metadata: StateSnapshotMetadata{Kind: kind, Timestamp: ..., Device: ...},
    Data: json.RawMessage(data),
}
snapJSON, _ := json.Marshal(snap)
```

indexer/pkg/dz/mroute/sync.go
```diff
@@ -0,0 +1,40 @@
...
+			continue
+		}
+		entries, err := Parse(d.RawJSON)
+		if err != nil {
+			s.log.Warn("mroute/sync: failed to parse dump, skipping device",
+				"device_pubkey", d.DevicePubkey,
+				"file", d.FileName,
+				"err", err,
+			)
+			continue
```
vihu, line 28: `Sync` skips parse failures, but the store writes with `MissingMeansDeleted=true`.

That means a single malformed device dump is not just ignored; its previously-current rows are missing from the batch and will be tombstoned by the type-2 writer. This seems contrary to the comment that a bad dump “should not prevent the rest of the fleet from being written” - with the current global replacement semantics, skipping is destructive for that device.

Can we either fail the sync before writing on any parse error, or make the delete semantics scoped to devices that were successfully parsed?

Also perhaps a test here would be useful too. So like seed current rows for device B, feed one malformed dump for B plus valid dumps for A/C, and assert B is not deleted unless that is explicitly intended.

ben-malbeclabs:
Thanks — both real bugs. Fixed in `97a7678`.

**Envelope unwrap.** Added `pkg/dz/statecollect.Envelope` mirroring `controlplane/telemetry/internal/state/snapshot.go`. `S3Source.FetchLatest` now unmarshals the envelope, returns the inner `Data` as `Dump.RawJSON`, and prefers `Metadata.Device` / `Metadata.Timestamp` over the values parsed from the key path. The MinIO layer-C test now uploads envelope-wrapped objects matching what the collector writes — that's the test-fidelity bug that masked this originally, fixed.

**Fail-fast Sync.** Switched to your first option: on any parse failure, `Sync` returns an error and writes nothing, preserving prior state for all devices under `MissingMeansDeleted=true`. Added `TestStore_Sync_FailsFastOnParseError` that seeds a row, calls `Sync` with one good + one corrupt dump, and asserts the seed survives.

Updated comment on `Sync` explains why fail-fast is the safe default here.

vihu (approved):

PR #614, opened 2026-06-01
indexer: parse flat shape from show ip mroute | json

ben-malbeclabs:
## Summary
- Arista's `show ip mroute | json` returns a flat `{sparseMode, bidirectional}` object for the default VRF — only `vrf all` produces the `{"vrfs": {<name>: ...}}` wrapper. The parser only handled the wrapper, so every prod upload parsed to zero entries.
- Make the parser accept both shapes; treat flat input as VRF `"default"`.
- Parsing a real mainnet S3 envelope now yields 612 entries (was 0).

## Symptom
`SyncIPMroute` activity reported `success` with `entries=0` for all three networks despite devices having live mroute state (verified by `show ip mroute | json` on `frk-dz001` testnet and `sea10-dz001` mainnet). The bucket received envelopes; the parser dropped them.

## Testing Verification
- Added `TestParse_FlatShape` covering the flat envelope shape pulled from a real S3 object.
- Parsed an actual mainnet S3 envelope (`s3://malbeclabs-device-telemetry-state-snapshots-mainnet-beta/snapshots/ip-mroute/.../20260530T033834Z.json`) — 612 entries, correct VRF/group/source/flags on the first row.
- Existing `parser_test.go` and full mroute suite still green.

Closes malbeclabs/infra#1416.

### DIFF
--- a/indexer/pkg/dz/mroute/parser.go
+++ b/indexer/pkg/dz/mroute/parser.go
@@ -8,8 +8,13 @@ import (
 )
 
 // rawDump matches the top-level shape of `show ip mroute | json` on Arista EOS.
+// Arista emits two shapes:
+//   - Wrapped (used by `show ip mroute vrf all | json`): {"vrfs": {<name>: rawVRF}}
+//   - Flat   (used by `show ip mroute | json`):          rawVRF at top level, default VRF
 type rawDump struct {
-	VRFs map[string]rawVRF `json:"vrfs"`
+	VRFs          map[string]rawVRF `json:"vrfs"`
+	SparseMode    *rawModeGroups    `json:"sparseMode,omitempty"`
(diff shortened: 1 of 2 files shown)


bgm-malbeclabs (approved):
The new test leaves the flat bidirectional path uncovered; the `bidirectional.groups` is empty, so the `appendEntries(out, "default", ModeBidirectional, *d.Bidirectional)` line never produces an asserted entry. Minor thing you can fix or leave

ben-malbeclabs:
> The new test leaves the flat bidirectional path uncovered; the `bidirectional.groups` is empty, so the `appendEntries(out, "default", ModeBidirectional, *d.Bidirectional)` line never produces an asserted entry. Minor thing you can fix or leave

Addressed in [e58947a](https://github.com/malbeclabs/lake/pull/614/commits/e58947a63b2c556c6761e16e6f37fcee46b2efce)

PR #617, opened 2026-06-02
feat(api): add multicast delivery-state endpoint

vihu:
# Summary

Adds a semantic multicast delivery-state API for inspecting observed forwarding state, expected topology, subscriber outcomes, MSDP state, and delivery anomalies for a multicast group.

## Changes

- Add `GET /api/dz/multicast-groups/{pk}/delivery-state`
- Enrich mroute rows with publisher/group context, freshness, RPF details, route IDs, and source-match status
- Resolve OIFs into physical links, subscriber tunnels, local interfaces, special interfaces, or unknown OIFs
- Include MSDP peers, MSDP PIM SA cache, and MSDP SA cache in the response
- Build observed segments, expected segments, subscriber outcomes, and anomaly summaries
- Use `log_ingestion_runs` freshness for mroute/MSDP sources so unchanged current rows are not incorrectly marked stale
- Expose `latest_ingested_at` separately from `latest_snapshot_ts`
- Add handler tests covering not-found, no-source, observed-state, stale-state, and ingestion-freshness behavior

## Validation

- `go test ./api/... -count=1`
- `golangci-lint run -c ./.golangci.yaml ./api/...`
- `make build`
- `make lint`


ben-malbeclabs:
Continuing from the review discussion about #1418 overlap — rather than abandon either approach, I've split the join logic into reusable ClickHouse views (#619 / #620 / #621) and prepared a refactor of this PR's handlers to read from those views instead of doing the joins inline.

**Refactor commit:** [`8721062`](https://github.com/malbeclabs/lake/commit/87210627a98badd10bf20f8755f143328161a6c9) — that's the focused diff. (The branch itself also contains merge commits for #619 / #620 / #621 so the new views exist for the refactor to reference; ignore those in review.)

### What it changes

- `multicast_delivery_routes.go` → `multicast_delivery_mroutes.go` (renamed; reads from `enriched_ip_mroute`).
- `multicast_delivery_oifs.go` reads from `enriched_ip_mroute_oifs`. The view does OIF expansion + classification, so `classifyMulticastOIF` in `_build.go` is removed.
- `multicast_delivery_msdp.go` reads from the three `enriched_ip_msdp_*` views. The two SA-cache variants share one helper.
- `MulticastDeliveryMSDPPeer` and `MulticastDeliveryMSDPSA` extended with `peer_device_*` and `remote_device_*` enrichment — newly resolvable via `dz_device_interface_ips` (parses `dz_devices_current.interfaces`). Verified against current mainnet data: 96/96 MSDP peers and 617/616 SA-cache remote_addresses match a device.
- `sa.status` renamed to `accept_status` (avoids ambiguity with `multicast_group_status`).
- `LIMIT 5000` / `LIMIT 20000` silent caps removed — streams all rows. Optional handler-side Limit/Offset stays.

Net: -84 lines across 6 files; build + vet clean.

### How to consume

Pick whatever's useful — cherry-pick `8721062`, or pull the branch wholesale, or treat as reference. View migrations are PRs #619 / #620 / #621 (stacked: #619 first, then #620, then #621 independent); this refactor depends on all three.

Deliberately scoped out of the commit so the diff stays focused:

- `{pk}` URL param → `{group}` rename in `main.go` (handler already accepts both).
- `MulticastDeliveryPIMNeighbor struct{}` placeholder removal.
- Test fixtures reusing `attrs_hash = 1` (minor — would mask SCD2 bugs in real ingestion).

Happy to fold those in if useful or leave to a follow-up — your call.

ben-malbeclabs (approved):

PR #618, opened 2026-06-02
api,web: add service-aware path visualization (RFC-18)

### LINKED ISSUE #3358: Service-aware path visualization
When tracing a path between two nodes in lake, resolve via the correct topology for the selected traffic type: unicast uses algo 128 (tagged links only), multicast uses algo 0 (all links). Paths between the same pair may differ.

**Milestone:** RFC-18: Observability
**RFC:** https://github.com/malbeclabs/doublezero/pull/3288

---

**Repo:** lake
**Size:** Medium

**Tests:**
- Unicast path excludes untagged links
- Multicast path includes all links
- Paths between same node pair differ when untagged links exist on the route

**Blockers:** #3356 (topology view filtering)

ben-malbeclabs:
## Summary

- Add `service` query parameter (`unicast` | `multicast`) to all path-finding API endpoints (`/api/topology/paths`, metro path latency, metro path detail, metro paths, metro device paths) — unicast excludes links with empty `link_topologies` and drained links (flex-algo 128 constrained topology), multicast includes all links (algo 0)
- Add unicast/multicast toggle to the path calculator page, topology map, globe, and graph path panels
- Default to unicast, which reflects the real forwarding plane for VPN traffic post RFC-18

Closes https://github.com/malbeclabs/doublezero/issues/3358

## Testing Verification

- `go build ./api/...` and `go vet ./api/...` pass
- `bun tsc --noEmit` passes for web (pre-existing radix-ui errors unrelated)
- Verified production ClickHouse has 165 links tagged `UNICAST-DEFAULT` and 2 untagged — unicast mode will exclude the 2 untagged links, producing different paths when those links are on the route
- URL params (`service`, `path_service`) persist across navigation and "View in graph" deep links


api/handlers/isis.go
```diff
@@ -1731,7 +1739,8 @@ func (a *API) GetMetroPathLatency(w http.ResponseWriter, r *http.Request) {
 	ctx, cancel := context.WithTimeout(r.Context(), 30*time.Second)
 	defer cancel()
 
-	response, err := a.FetchMetroPathLatencyData(ctx, optimize)
+	service := r.URL.Query().Get("service")
```
ben-dz: `service` is read but not validated here (nor in GetMetroPathDetail:1909, GetMetroPaths:2038, GetMetroDevicePaths:2957). Only GetISISPaths validates `service ∈ {unicast, multicast, ""}`. Any other value (e.g. `Unicast`, `foo`, trailing space) silently falls through to all-links behavior with a 200. Reachable today only via manual URL editing, but the inconsistency across five sibling handlers is worth closing — hoist the validation into a shared helper and apply it to all five.

web/src/components/path-calculator-page.tsx
```diff
@@ -396,6 +411,38 @@ export function PathCalculatorPage() {
...
+            <button
+              onClick={() => updateService('multicast')}
+              className={`px-3 py-1 text-sm rounded-sm transition-colors ${
+                service === 'multicast'
+                  ? 'bg-background text-foreground shadow-sm'
+                  : 'text-muted-foreground hover:text-foreground'
+              }`}
+            >
+              Multicast
+            </button>
```
ben-dz: Copy says "Flex-algo 128 — topology-tagged links only", but the backend filter (isis_ksp.go:56-60) matches any non-empty `link_topologies` membership, not algo 128 specifically. If more than one flex-algo topology ever exists this is misleading; consider "topology-tagged, non-drained links".

ben-dz (commented):
Service-aware path filtering is a clean, backward-compatible mechanism, and security is clean (no SQL injection, encodeURIComponent on the frontend). Given the API is UI-only and not a stable external contract, no findings block merge. Remaining items are consistency and coverage: validation exists on only 1 of 5 endpoints, three backend endpoints thread `service` with no frontend caller, URL persistence differs across the three canvases, `service` is echoed on only 1 of 5 responses, and there is no test coverage for the new filter. One latent trap: the metro-path-latency page cache would return stale all-links data if a `service=unicast` caller were added.

- Medium — Backend threads `service` through GetMetroPathLatency/GetMetroPathDetail/GetMetroPaths, but the frontend clients fetchMetroPathLatency (web/src/lib/api.ts:4453), fetchMetroPathDetail (:4485), and fetchMetroPaths (:4519) were not updated to send it. Three of five threaded endpoints have no UI caller — wire the param through or drop the unused threading. Wiring fetchMetroPathLatency is also what makes the latent cache trap go live.
- Medium — Response echoes `service` on only 1 of 5 endpoints: MultiPathResponse gained a `Service` field (isis.go:906) but MetroPathLatencyResponse/MetroPathDetailResponse/MetroPathsResponse/MetroDevicePathsResponse do not, despite all accepting the param. Echo it consistently or not at all.
- Medium — No test coverage: api/handlers/isis_test.go is untouched; nothing exercises the unicast filter, the cache interaction, or the validation. The cache-HIT and validation logic are testable with existing fixtures.
- Low — The unicast/multicast toggle JSX is duplicated four ways (PathModePanel.tsx:124, MetroPathModePanel.tsx:115, path-calculator-page.tsx:414) and four components each redeclare a `'unicast' | 'multicast'` useState. Extract a shared <ServiceToggle> component and a `PathService` type in api.ts.
- Low — Three names for one concept: `service` (API/calculator), `path_service` (map/graph URL params), `pathService` (React state), labeled "Traffic type" in the UI. Standardize the param name or document the mapping.

- `web/src/components/topology-graph.tsx:1482` — medium: This URL-writing effect never writes `path_service` (it's only read on restore at line 1547). So toggling traffic type inside the graph doesn't persist to the URL, and a "View in graph" deep-link loses the param on the next URL rewrite. topology-map does a full round-trip and topology-globe uses local state only — pick one persistence contract and apply it consistently.
- `api/handlers/isis.go:1730` — low: Latent stale-cache trap: the page cache is read here (keyed only on `optimize`) before `service` is read at line 1742, and the cache is populated with `service=""` (all links) via `FetchMetroPathLatencyData(ctx, strategy, "")` in worker/workflow.go:180. No caller passes `service` to this endpoint today, so it's not currently reachable — but it becomes a live silent-wrong-data bug (`?service=unicast` served cached all-links with `X-Cache: HIT` on mainnet) the moment fetchMetroPathLatency is wired for service. Fix: read `service` first and bypass the cache when `service == "unicast"`, or make the cache key service-aware and populate both variants.

api/handlers/isis.go
ben-malbeclabs: Good call. Option A removed `service` from `GetMetroPathLatency` / `GetMetroPathDetail` / `GetMetroPaths`, so the only remaining handler that read `service` without validating was `GetMetroDevicePaths`. Fixed in e53f593: extracted a shared `validPathService` helper and applied it to both handlers that still accept the param (`GetISISPaths` and `GetMetroDevicePaths`), so a bad value returns an error response instead of silently falling through to all-links. Added a `GetMetroDevicePaths` invalid-service test alongside the existing `GetISISPaths` one.

web/src/components/path-calculator-page.tsx
ben-malbeclabs: Fixed in e53f593. Changed the unicast description to "Topology-tagged, non-drained links only" (and multicast to "All links"), which matches the actual filter (`link_topologies` non-empty AND `unicast_drained = 0`) rather than implying flex-algo 128 specifically. The text now lives in the shared `<ServiceToggle>` component after the refactor.

ben-malbeclabs:
Thanks for the thorough pass. Addressed in `2a7a946` — went with trimming scope to exactly what's wired end-to-end rather than expanding the frontend. Point by point:

**Medium — `service` threaded through 3 endpoints with no frontend caller**
Dropped the param from `GetMetroPathLatency`, `GetMetroPathDetail`, and `GetMetroPaths` (plus the `FetchMetroPathLatencyData` signature and the `worker/workflow.go` cache call). `service` now lives only on the two endpoints wired end-to-end: `GetISISPaths` and `GetMetroDevicePaths`. This keeps the PR scoped to #3358 rather than shipping speculative plumbing.

**Medium (latent stale-cache trap on metro-path-latency)**
Resolved as a consequence of the above: `GetMetroPathLatency` no longer accepts `service`, so there's no path by which a `?service=unicast` request can be served cached all-links data. The cache stays keyed on `optimize` only, which is now correct because that's the only dimension the endpoint varies on. Good catch — this was the real footgun.

**Validation across handlers** (from your summary + inline note)
After trimming, only `GetISISPaths` and `GetMetroDevicePaths` accept `service`. Both now validate through a shared `validPathService` helper (`e53f593`), so a bad value returns an error response instead of silently serving all-links. Details on the inline thread.

**Medium — response echoes `service` on only 1 of 5**
Removed the `Service` field from `MultiPathResponse` (Go + the TS type). Went with "not at all" — the client already knows what it asked for, so echoing it added no value.

**Medium — no test coverage**
Added `api/handlers/isis_service_filter_test.go`. Table-driven test through `GetISISPaths` that seeds a 4-device topology and asserts: unicast excludes both untagged **and** `unicast_drained` links, multicast includes them, unset `service` defaults to all-links, and an invalid `service` value is rejected. Verified the graph loads 2 nodes/1 edge under unicast vs 4/3 under multicast.

**Low — toggle JSX duplicated 4×**
Extracted a shared `<ServiceToggle>` (`size` `sm`/`md`) and a `PathService` type in `api.ts`, replacing the inline toggles in `PathModePanel`, `MetroPathModePanel`, and `path-calculator-page`. State across the components now uses the shared `PathService` type.

**Low — three names for one concept**
Kept the layering intentionally: `service` (API query param), `path_service` (URL param, namespaced to avoid collision with other topology params), `pathService` (React state, camelCase). The one thing that was genuinely loose — the TS type — is now unified as `PathService`.

**Medium (not anchored) — topology-graph URL persistence inconsistency**
Acknowledged and left as-is in this PR. The three canvases already diverged before this change (map round-trips, graph reads-only-on-restore, globe is local-only), and `path_service` just inherits whatever each canvas does. It's low impact (a deep-link can lose the param on the next URL rewrite in the graph view) and orthogonal to service-aware paths. Happy to align all three to one persistence contract as a follow-up if you'd like — I'd rather not fold a cross-canvas refactor into this PR.

ben-malbeclabs:
@ben-dz ready for another review when you have a moment, thanks.

ben-dz (approved):

PR #620, opened 2026-06-02
indexer: enriched MSDP views (peers, pim_sa_cache, sa_cache)

ben-malbeclabs:
## Summary

Adds three ClickHouse views enriching `dz_ip_msdp_*_current` with device, group, and publisher attribution:

- **`enriched_ip_msdp_peers`** — joins each MSDP peer session to the local device and (via `dz_device_interface_ips` from the parent PR) the remote peer device. The 172.16.x.x mesh-space `peer_address` resolves to whichever device has that IP on one of its interfaces, exposing `peer_device_pk` + `peer_device_code` + `peer_interface_name`. Verified empirically: 96/96 production mainnet MSDP peers map to a device this way.
- **`enriched_ip_msdp_pim_sa_cache`** — joins each PIM SA cache entry to local device, multicast group, and publisher user (via `source_address` → user's `dz_ip` when the group is in the publishers list). `rp_address` left raw — production uses anycast `10.0.0.0` which doesn't map to a specific device.
- **`enriched_ip_msdp_sa_cache`** — same enrichments as PIM SA cache, plus remote device resolution via `dz_device_interface_ips` (mapping the 172.16.x.x `remote_address` to a device pubkey + code). `sa.status` is aliased to `accept_status` to disambiguate from `multicast_group_status`. Verified empirically: 617/616 production mainnet SA cache rows match a device this way (one source matched two devices, handled by LEFT ANY JOIN).

## Stacking

PR 2 of 3 for infra#1418. **Base: `bc/enriched-mroute`** (PR 1) — this branch depends on the `dz_device_interface_ips` view introduced there. Merge PR 1 first; this should be rebased / merged onto main afterwards.

## Testing Verification

- `TestMigration_Applies` extended to cover all three new views.
- `TestEnrichedView_MSDPPeers` — inserts two devices (sea + nyc, nyc has the peer address on a Loopback interface), then an MSDP peer row. Asserts the peer_address resolves to nyc's device pubkey + code + interface name.
- `TestEnrichedView_MSDPSACache` — full round-trip exercising remote-device resolution + group join + publisher attribution, asserts `accept_status = 'accepted'`, `source_match_status = 'publisher_matched'`, plus all peer/publisher codes are populated.

Refs malbeclabs/infra#1418.

### DIFF
--- a/indexer/db/clickhouse/migrations/20260602000002_enriched_ip_msdp_views.sql
+++ b/indexer/db/clickhouse/migrations/20260602000002_enriched_ip_msdp_views.sql
@@ -1,0 +1,183 @@
+-- +goose Up
+
+-- +goose StatementBegin
+-- enriched_ip_msdp_peers
+-- Joins dz_ip_msdp_peers_current with the local device and (via
+-- dz_device_interface_ips) the peer device. Mesh-space peer addresses
+-- (172.16.x.x) map back to whichever device has that IP on one of its
+-- interfaces. Every entity pubkey is paired with its human-readable code.
+CREATE OR REPLACE VIEW enriched_ip_msdp_peers
+AS
+SELECT
+    p.entity_id AS msdp_peer_entity_id,
+    p.snapshot_ts,
+    p.ingested_at,
+
+    -- local device
+    p.device_pubkey AS device_pk,
+    d.code AS device_code,
+    d.status AS device_status,
+    d.metro_pk,
+    m.code AS metro_code,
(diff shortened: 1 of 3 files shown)


vihu (approved):
Just a nitpick, `enriched_ip_msdp_pim_sa_cache` has migration-existence coverage, but peers and SA-cache have direct round-trip tests while PIM-SA does not. A small PIM-SA round-trip test would make the coverage symmetrical, but I don't think that's blocking.

PR #622, opened 2026-06-03
indexer: convert dz_device_interface_ips to a refreshable materialized view

ben-malbeclabs:
## Summary

Converts `dz_device_interface_ips` from a plain VIEW to a REFRESHABLE MATERIALIZED VIEW. Fixes the JOIN explosion that crashes `enriched_ip_msdp_peers` and `enriched_ip_msdp_sa_cache` on real production data.

## Symptom

Querying either of the two affected views — for any column that references the peer/remote device resolution — OOMs with a `128 TiB allocation attempted` error. Observed against `lake_testnet` on lake-staging:

```
SELECT device_code, peer_address, peer_device_code FROM lake_testnet.enriched_ip_msdp_peers LIMIT 1
→ Code: 241. Query memory limit exceeded: would use 128.00 TiB (attempt to allocate chunk of 128.00 TiB)
```

The basic `count()` succeeds (returns 56) but any projection through the JOIN explodes. The other four enriched views are unaffected and return data normally.

## Root cause

ClickHouse 25.12.11's query optimizer fails to estimate result sizes when a JOIN's right side is derived from `ARRAY JOIN JSONExtractArrayRaw(...)`. `dz_device_interface_ips` was defined as:

```sql
CREATE OR REPLACE VIEW dz_device_interface_ips AS
SELECT ... , splitByChar('/', JSONExtractString(iface, 'ip'))[1] AS ip_address
FROM dz_devices_current d
ARRAY JOIN JSONExtractArrayRaw(d.interfaces) AS iface
WHERE ...
```

The view itself works for direct `SELECT`. The bug surfaces only when something `LEFT ANY JOIN`s against the `ip_address` column. CH allocates a max-int sized hash table and the query OOMs.

Reproduced and isolated: same explosion happens with the JOIN inlined, with explicit `toString()` casts on the join key, with `extract()` instead of `splitByChar()`, and with `join_algorithm='hash'` forced.

## Fix

Replace the plain VIEW with a REFRESHABLE MATERIALIZED VIEW. The JSON parsing + ARRAY JOIN runs once per refresh and the result lands in a real `MergeTree` table. Downstream JOINs see a normal table with proper size statistics and plan correctly.

Refresh cadence: 60 seconds. Device interface lists change rarely (only when devices are added/removed/reconfigured), so the freshness lag is acceptable.

The view's name and column shape are unchanged, so `enriched_ip_msdp_peers` and `enriched_ip_msdp_sa_cache` need no modification.

## Testing Verification

- Reproduced the OOM on `lake_testnet.enriched_ip_msdp_peers` against staging's CH (25.12.11). Created a test refreshable MV with the same definition; `LEFT ANY JOIN` against it returned three rows correctly (`172.16.0.17 → fra-dz-001-x`, `172.16.0.39 → lon-dz001`).
- `TestMigration_Applies` extended to accept the new `mv` kind for `dz_device_interface_ips` (engine reports as `MaterializedView` in `system.tables`).

## Why this blocks the rollout

Promoting the current `:edge` to prod without this fix means mainnet's `enriched_ip_msdp_peers` and `enriched_ip_msdp_sa_cache` would crash on every query. Both are intended consumers for the per-multicast-group dashboard (infra#1428).

Refs malbeclabs/infra#1418.

### DIFF
--- a/api/testing/clickhouse.go
+++ b/api/testing/clickhouse.go
@@ -274,6 +274,14 @@ func SetupClickHouseWithMigrationsForTest(t *testing.T, db *ClickHouseDB) (drive
 	require.NoError(t, err, "failed to list template tables")
 	defer rows.Close()
 
+	// Collect all clone queries first, then apply with dependency-aware retries.
+	// system.tables returns rows in arbitrary order; plain VIEWs are created
+	// lazily so order doesn't matter for them, but MATERIALIZED VIEWs eagerly
+	// validate their AS SELECT against the source table at CREATE time. A
+	// repeat-until-no-progress loop resolves dependency ordering without
+	// having to parse SQL.
+	type pending struct{ name, query string }
+	var queries []pending
 	for rows.Next() {
 		var name, engineFull, createQuery string
 		require.NoError(t, rows.Scan(&name, &engineFull, &createQuery))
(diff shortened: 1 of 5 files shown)


vihu:
A few things:

1. So we should fix tests by forcing refresh after device-interface inserts:

```sql
  SYSTEM REFRESH VIEW dz_device_interface_ips
```

2. Add some sort of regression test that selects peer/remote columns from both affected views, not just migration existence.

3. Maybe change the order key:

Current:

```sql
  ORDER BY (device_pk, interface_name)
```

Since the join key is ip_address, per schema-pk-prioritize-filters, I’d prefer:

```sql
  ORDER BY (ip_address, device_pk, interface_name)
```

ben-malbeclabs:
> A few things:
> 
> 1. So we should fix tests by forcing refresh after device-interface inserts:
> 
> ```sql
>   SYSTEM REFRESH VIEW dz_device_interface_ips
> ```
> 
> 2. Add some sort of regression test that selects peer/remote columns from both affected views, not just migration existence.
> 3. Maybe change the order key:
> 
> Current:
> 
> ```sql
>   ORDER BY (device_pk, interface_name)
> ```
> 
> Since the join key is ip_address, per schema-pk-prioritize-filters, I’d prefer:
> 
> ```sql
>   ORDER BY (ip_address, device_pk, interface_name)
> ```

Points addressed: [aa7534e](https://github.com/malbeclabs/lake/pull/622/commits/aa7534e5285e01a8f7c1e699f5f1de007c428dad) and [66858fc](https://github.com/malbeclabs/lake/pull/622/commits/66858fcc4a105a4188b002249dcf69cf60e8e828)

vihu (approved):

PR #626, opened 2026-06-03
indexer: health_multicast_user view (per-user onchain↔dataplane reconciliation)

ben-malbeclabs:
## Summary

Second view of Track 1 (infra#1501). Adds `health_multicast_user` — the per-user onchain↔dataplane reconciliation view.

For every activated multicast user × group they belong to, produces one row with:

| Column | Meaning |
| --- | --- |
| `mode` | `P` / `S` / `P+S` derived from `dz_users_current.publishers` / `subscribers` JSON arrays |
| `expected_tunnel_position` | `iif` for publishers, `oif` for subscribers, `iif\|oif` for both |
| `publisher_iif_observed` | `Tunnel<user.tunnel_id>` appears as `rpf_interface` on an mroute for this group on the user's device |
| `subscriber_oif_observed` | `Tunnel<user.tunnel_id>` appears in `oif_list` of an mroute for this group on the user's device |
| `reconciled` | every expected position observed |
| `health_status` | `healthy` / `degraded` / `unhealthy` |
| `mismatch_reason` | specific free-text explanation when not reconciled |

Directly answers the cross-check the operator called out in the planning discussion: "If a user is connected as tunnel505 on frankry as a multicast publisher, there is a tunnel505 in the Incoming interface on frankry. Similarly, if the user is connected as a subscriber, their tunnel appears in the Outgoing interface list on the device it is connected to."

## Stacking

Stacks on `bc/mcast-health-mroute-view` (#625) — same migration file ordering. Builds on the same `enriched_*` + onchain dim substrate, no dependency on #625's views.

## What's left in Track 1

The third view, `health_publisher_subscriber_path` (strict per-pair path verification), lands as a follow-up. CH SQL's lack of full recursive CTE support means the path-traversal design needs more thought — fixed-depth joins, API-layer Dijkstra over the topology graph, or endpoint-only verification with a documented limitation. Worth its own focused PR.

## Testing Verification

- `TestMigration_Applies` extended to cover `health_multicast_user`.
- `TestHealthMulticastUser_Reconciliation` covers all three paths via synthetic data:
  - `u-pub-healthy`: publisher whose Tunnel appears as RPF interface → `mode='P'`, `reconciled=true`, `health_status=healthy`.
  - `u-pub-silent`: publisher whose Tunnel is absent from every RPF position → `reconciled=false`, `health_status=unhealthy`, `mismatch_reason` mentions "Tunnel504" and "RPF interface".
  - `u-sub-healthy`: subscriber whose Tunnel appears in OIF list → `mode='S'`, `reconciled=true`, `health_status=healthy`.

Refs malbeclabs/infra#1501.

### DIFF
--- a/indexer/db/clickhouse/migrations/20260603000003_health_multicast_user_view.sql
+++ b/indexer/db/clickhouse/migrations/20260603000003_health_multicast_user_view.sql
@@ -1,0 +1,168 @@
+-- +goose Up
+
+-- +goose StatementBegin
+-- health_multicast_user
+--
+-- One row per (multicast user, group) reconciling onchain membership
+-- (dz_users_current.publishers / .subscribers) against the observed
+-- dataplane mroute state on the user's device.
+--
+-- A user is healthy for a group iff their Tunnel<tunnel_id> appears in
+-- the expected position(s) on the device's mroute entries for that group:
+--   - mode 'P' (publisher):   Tunnel<N> appears as rpf_interface (IIF)
+--   - mode 'S' (subscriber):  Tunnel<N> appears in oif_list (OIF)
+--   - mode 'P+S' (both):      both of the above
+--
+-- Second track of malbeclabs/infra#1501. The third view —
+-- health_publisher_subscriber_path — lands in a follow-up.
+CREATE OR REPLACE VIEW health_multicast_user
+AS
+WITH
+multicast_users AS (
+    SELECT
+        u.pk AS mu_user_pk,
+        u.dz_ip AS mu_user_dz_ip,
+        u.tunnel_id AS mu_user_tunnel_id,
+        u.device_pk AS mu_user_device_pk,
+        u.owner_pubkey AS mu_user_owner_pubkey,
+        d.code AS mu_user_device_code,
+        JSONExtract(u.publishers, 'Array(String)') AS mu_publisher_group_pks,
+        JSONExtract(u.subscribers, 'Array(String)') AS mu_subscriber_group_pks
+    FROM dz_users_current u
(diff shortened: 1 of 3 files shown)


vihu:
One thing to clarify before this becomes the per-user health source of truth: the IIF/OIF tunnel checks appear aggregated by device and group, not by source_address. Could a subscriber have its tunnel present for one source and missing for another, but still look reconciled here?

Is this intentionally a coarse tunnel-membership or control-plane reconciliation view? If yes, I think the API/UI copy should say that and rely on health_publisher_subscriber_path for source-specific delivery. If not, could we carry source_address through or add a multi-publisher partial-delivery test?

ben-malbeclabs:
> One thing to clarify before this becomes the per-user health source of truth: the IIF/OIF tunnel checks appear aggregated by device and group, not by source_address. Could a subscriber have its tunnel present for one source and missing for another, but still look reconciled here?
> 
> Is this intentionally a coarse tunnel-membership or control-plane reconciliation view? If yes, I think the API/UI copy should say that and rely on health_publisher_subscriber_path for source-specific delivery. If not, could we carry source_address through or add a multi-publisher partial-delivery test?

@vihu fixed in [ea93e2f](https://github.com/malbeclabs/lake/commit/ea93e2f872b22c698ec3bfe51588f471dcf998c3). The view now tracks at (device, group, source_address): a subscriber's tunnel must appear in OIF for every (S, G) at its LHR, not just one.

Partial coverage drops the verdict to degraded with a 1 of 2 publisher sources reason, and there's a new TestHealthMulticastUser_PartialDelivery covering the multi-publisher case.

vihu (approved):

PR #628, opened 2026-06-04
api: multicast health endpoints (infra#1501 Track 2)

ben-malbeclabs:
## Summary

Track 2 of infra#1501. Adds four read-only HTTP endpoints that expose the Track 1 health views to the web app and any other consumer.

| Endpoint | Backing view | Notes |
| --- | --- | --- |
| `GET /api/dz/multicast-groups/{pk}/health` | UNION across `health_mroute` / `health_multicast_user` / `health_publisher_subscriber_path` | Per-group summary: counts per status across all three granularities |
| `GET /api/dz/multicast-groups/{pk}/health/users` | `health_multicast_user` | Per-user reconciliation rows for the group, paginated |
| `GET /api/dz/multicast-groups/{pk}/health/paths` | `health_publisher_subscriber_path` | Per `(publisher, subscriber)` rows for the group, paginated |
| `GET /api/dz/users/{pk}/health` | `health_multicast_user` filtered by user_pk | Per-group reconciliation for one user |

Conventions match `lake#617`:
- Pubkey ↔ code paired on every entity field
- Optional `limit` / `offset`; stream-all by default (no silent caps)
- Pure SQL handlers — no joining logic in Go, all enrichment lives in the views

## Stacking

`bc/mcast-health-api` → `bc/mcast-health-path` (#627) → `bc/mcast-health-user-path` (#626) → `bc/mcast-health-mroute-view` (#625). Each rebases onto main as the chain merges.

## Testing Verification

Five tests, all green against the testcontainer:

- `TestGetMulticastGroupHealth_Summary` — full roll-up: with 1 publisher + 1 subscriber + 2 mroute rows, returns `Users.Total=2, Healthy=2`, `Paths.Total=1, Healthy=1`, `Mroutes.Total=2, Healthy=2`.
- `TestGetMulticastGroupHealth_NotFound` — unknown group → 404.
- `TestGetMulticastGroupHealth_Users` — paginated per-user response shape, both users reconciled.
- `TestGetMulticastGroupHealth_Paths` — single `(pub, sub)` row returned with `endpoints_reconciled=true`, `verification_method=endpoints_only`.
- `TestGetUserHealth_PerGroupRows` — per-user endpoint returns membership rows.
- `TestGetUserHealth_NoMemberships` — unknown user returns 200 with empty `items` (not 404, since "no multicast memberships" is a valid steady state).

## What's next

- **Track 1 follow-up**: recursive OIF chain walk to flip `verification_method` from `endpoints_only` to `full_path` on the paths view.
- **Track 3**: lake/web UI consuming these endpoints — Health tab on `/dz/multicast-groups/{group}`, health card on `/dz/users/{user}`.

Refs malbeclabs/infra#1501.


vihu:
The users and paths endpoints appear to fetch all matching rows from ClickHouse and then apply limit/offset in Go. Paths can be publisher x subscriber per group, so this could become expensive quickly, especially with UI refetch. Could we push LIMIT/OFFSET into SQL and use a separate count query or bounded total strategy?

ben-malbeclabs:
> The users and paths endpoints appear to fetch all matching rows from ClickHouse and then apply limit/offset in Go. Paths can be publisher x subscriber per group, so this could become expensive quickly, especially with UI refetch. Could we push LIMIT/OFFSET into SQL and use a separate count query or bounded total strategy?

Pagination silently excludes potentially the rows operators most need to see, so we're sticking with stream-all (per the #617 convention). Pushed [01098e3](https://github.com/malbeclabs/lake/pull/628/commits/01098e365217fc3d3ef0d48d92e0c7a33800924e) instead: ORDER BY severity (unhealthy → degraded → unknown → healthy) so any consumer truncating downstream still lands on the actionable rows; payload is bounded by group membership.

vihu (approved):

PR #629, opened 2026-06-04
web: multicast health tab + per-user health card (Track 3 of infra#1501)

ben-malbeclabs:
## Summary

Track 3 of infra#1501. Surfaces the Track 1 health views and Track 2 API endpoints in the lake/web data app — same UX patterns the existing detail pages use, no new top-level routes.

| Page | Change |
| --- | --- |
| `/dz/multicast-groups/{group}?tab=health` | New **Health** tab. Summary counts across three granularities (users, paths, mroutes), a per-user reconciliation table, and a per-path table with verification method. |
| `/dz/users/{user}` | The existing Multicast Groups card gains a health badge per (group, mode) row. The group link now deep-links into the Health tab. |

Status colors: green (healthy) / amber (degraded) / red (unhealthy) / muted (unknown). All classification comes from the views — the web app only renders.

## Stacking

`bc/mcast-health-web` (this) → `bc/mcast-health-api` (#628) → `bc/mcast-health-path` (#627) → `bc/mcast-health-user-path` (#626) → `bc/mcast-health-mroute-view` (#625).

## Testing Verification

- `bun run build` — production tsc + vite build clean.
- `bun run test:run` — 180/180 vitest unit tests pass.
- Browser smoke against `data.malbeclabs.com` via local dev server:
  - `/dz/multicast-groups/tiredsolid?tab=health` — three tabs render (Publishers / Subscribers / Health). On the Health tab, the "Failed to load health data" banner displays cleanly because the prod API hasn't yet been deployed with #628 — confirms error handling.
  - `/dz/users/{publisher_pk}` — Multicast Groups card renders, group link points at `?tab=health`, no health badges yet (expected — depends on #628).

Live happy-path verification against staging once #628 is deployed is the remaining step.

## Out of scope (follow-up)

- Per-row red/yellow/green indicators on the existing Publishers / Subscribers tabs. Currently only the dedicated Health tab carries per-row status. Separable — adds a column to the existing table and another query.
- Track 1 follow-up: recursive OIF chain walk to flip `verification_method` from `endpoints_only` to `full_path`.

Refs malbeclabs/infra#1501.

PR #630, opened 2026-06-04
web: multicast health tab polish — legend + tooltips + nav arrows

ben-malbeclabs:
## Summary

Three small UX additions on top of #629 that make the Health tab discoverable without prior context.

- **Status legend** beneath the Reconciliation summary card — four pills, each with a one-line definition next to its colored dot. Visible without hover.
- **Info (?) icons** next to each section header (Summary / Per-user / Per-path) opening a tooltip with the per-section definition. Surfaces nuances like "paths is endpoints-only verification right now."
- **Chevron-right markers** next to every entity link in both tables — user, device, publisher, subscriber, FHR, LHR. Makes the drill-down affordance obvious without changing the click target.

No API contract change. One file touched.

## Stacking

`bc/mcast-health-web-polish` → `bc/mcast-health-web` (#629).

## Testing Verification

- `bunx tsc -b` clean, `bun run test:run` 180/180.
- Browser smoke against the local demo-mock backend:
  - `/dz/multicast-groups/demo-degraded?tab=health` — 3 info icons present, legend renders 4 dots with definitions, 8 chevron-right markers across the two tables (2 links × 2 user-rows + 4 links × 1 path-row).
  - Tooltip content verified by focus-trigger on the summary info icon — opens with the section help text.

Refs malbeclabs/infra#1501.


ben-malbeclabs:
Folded into #633

PR #631, opened 2026-06-04
indexer: add health_multicast_user_rate view (rate reconciliation, B1 of infra#1501)

ben-malbeclabs:
## Summary

First PR of the rate-reconciliation stack (B1 of infra#1501). Adds a ClickHouse view that layers a 5-min data-plane rate dimension on top of the existing control-plane health classification.

After this lands, "healthy" means both:
- onchain ↔ mroute reconciliation passes (existing `health_multicast_user`), **and**
- every subscriber's TX rate matches the sum of its group's publishers' RX rates (new).

The new view, `health_multicast_user_rate`, wraps the existing view — additive, not a replacement.

## What the view exposes

| Column | Description |
| --- | --- |
| `rate_bucket_ts` | 5-min bucket the rate was read from. NULL when no counter row in the freshness window. |
| `observed_bps_5m` | Publisher RX or subscriber TX at the user's tunnel. |
| `expected_bps_5m` | For subscribers: sum of all publishers' RX. NULL for publishers; NULL when any publisher in the group has no counter row. |
| `rate_status` | `reconciled \| mismatch \| unknown`. |
| `rate_status_reason` | Fine-grained: `active \| idle \| no_data` (publishers) or `reconciled \| mismatch \| monitoring_gap \| group_idle` (subscribers). |
| `control_plane_status` | The existing CP-only verdict, preserved for drill-in. |
| `health_status` | **REPLACED** with the combined CP × rate verdict per the spec matrix. |

## Behavioral knobs (baked into the view)

- **Tolerance**: `|observed - expected| <= max(5%, 1 Mbps)`.
- **Freshness window**: 15 min (3× the 5-min bucket cadence). Older counter data is treated as missing.
- **Strict idle**: any publisher in the group with no counter row collapses every subscriber's `rate_status` to `unknown` (reason: `monitoring_gap`). We refuse to reconcile against a deflated expected.
- **Group-idle**: all publishers transmitting zero → subscribers go `unknown` (reason: `group_idle`).

## Combined-status matrix

|  | rate=reconciled | rate=mismatch | rate=unknown |
|---|---|---|---|
| CP=healthy | healthy | degraded | unknown |
| CP=degraded | degraded | unhealthy | degraded |
| CP=unhealthy | unhealthy | unhealthy | unhealthy |

## Implementation notes

- **ClickHouse footgun fixed**: `LEFT JOIN` with non-Nullable right-side columns returns defaults (`0`, epoch) rather than NULL. That made "row missing" indistinguishable from "row with zero rate" via NULL checks. The view plumbs a `1 AS ur_present` sentinel column through the `user_rates` CTE so the outer joins can tell those cases apart.
- **No new data sources**: `device_interface_rollup_5m` already carries per-(device, tunnel) 5-minute rate aggregates indexed by `user_pk` and `user_tunnel_id`. No new ingestion required.

## Stacking

`bc/mcast-health-rate-view` → `bc/mcast-health-web-polish` (#630). Additive — no rebase of existing PRs needed.

Follow-ups:
- **B2**: API endpoints (`/health/users` switches to read from this view, gains rate fields).
- **B3**: web UI (per-user table gains a rate column; combined `health_status` flows through).

## Testing Verification

`TestHealthMulticastUserRate` exercises three groups (clean / monitoring-gap / idle) covering all seven rate-state combinations:

| User | Setup | rate_status | rate_status_reason | combined health |
| --- | --- | --- | --- | --- |
| `u-pub-active` | publisher, RX = 10 Mbps | reconciled | active | healthy |
| `u-sub-recon` | subscriber, TX = 10 Mbps | reconciled | reconciled | healthy |
| `u-sub-mis` | subscriber, TX = 50 Mbps | mismatch | mismatch | degraded |
| `u-pub-nodata` | publisher, no counter row | unknown | no_data | unknown |
| `u-sub-gap` | subscriber w/ no-data publisher | unknown | monitoring_gap | unknown |
| `u-pub-idle` | publisher, RX = 0 | unknown | idle | unknown |
| `u-sub-idle` | subscriber w/ idle publisher | unknown | group_idle | unknown |

Full `mroute` test package green (`go test ./indexer/pkg/dz/mroute/` 58s).

Refs malbeclabs/infra#1501.

### DIFF
--- a/api/handlers/multicast_delivery.go
+++ b/api/handlers/multicast_delivery.go
@@ -2,6 +2,11 @@
 // back future per-group and per-user health endpoints (infra#1501 Track 2).
 // They classify mroute rows and surface (active_publisher, group, device)
 // gaps without requiring this package to recompute the joins.
+//
+// The health_multicast_user_rate view (migration 20260604000001) extends
+// the per-user health classification with a 5-min rate dimension, layering
+// data-plane reconciliation on top of the control-plane verdict (Track B1
+// of infra#1501). Endpoints reading from it are added in a follow-up.
 
 package handlers
 
--- a/indexer/db/clickhouse/migrations/20260604000001_health_multicast_user_rate_view.sql
+++ b/indexer/db/clickhouse/migrations/20260604000001_health_multicast_user_rate_view.sql
@@ -1,0 +1,145 @@
+-- +goose Up
+
+-- +goose StatementBegin
+-- health_multicast_user_rate
+--
+-- Extends health_multicast_user with a data-plane rate dimension. The
+-- existing view verifies that the control plane (mroute state) reflects
+-- the onchain expectation. This view adds:
+--
+--   1. observed_bps_5m — the 5-min max rate measured at the user's tunnel.
+--      For publishers: max_in_bps (device's RX from the publisher).
+--      For subscribers: max_out_bps (device's TX toward the subscriber).
+--
+--   2. expected_bps_5m — only defined for subscribers: the sum of every
+--      publisher's RX rate in the same multicast group. A subscriber's TX
+--      should equal this sum.
+--
+--   3. rate_status — reconciled | mismatch | unknown.
+--      Tolerance: |observed - expected| <= max(5%, 1 Mbps).
+--
+--   4. rate_status_reason — a more specific cause behind rate_status.
+--      Distinguishes "no_data" (monitoring gap), "idle" (registered but
+--      zero traffic), "monitoring_gap" / "group_idle" (subscriber-side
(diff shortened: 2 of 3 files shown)


vihu:
Rate reconciliation question: health_multicast_user_rate treats P+S users as publisher-side for rate checks, and the rate join uses device plus tunnel without user_pk even though the rollup has user_pk.

If P+S users exist or tunnel IDs can be reused over the bucket/window, this could under-check subscriber-side rate or misattribute rate. Should we join on user_pk and either define P+S rate semantics explicitly or document this as publisher-side only?

ben-malbeclabs:
Folded into #633

PR #632, opened 2026-06-04
api: read multicast health from rate view (B2 of infra#1501)

ben-malbeclabs:
## Summary

B2 of infra#1501. Wires the per-user health endpoints to the new `health_multicast_user_rate` view (PR #631) and surfaces the rate dimension.

| Endpoint | Change |
| --- | --- |
| `GET /api/dz/multicast-groups/{pk}/health` | Summary counts now reflect the **combined** CP × rate verdict. |
| `GET /api/dz/multicast-groups/{pk}/health/users` | Per-user rows now expose the rate fields below. |
| `GET /api/dz/users/{pk}/health` | Same — per-(user, group) rows include rate fields. |
| `GET /api/dz/multicast-groups/{pk}/health/paths` | Untouched. Path-level full-path walk is the next follow-up. |

## Response-shape additions

```jsonc
{
  "rate_bucket_ts":   "2026-06-04T01:13:00Z",  // null when no counter row in 15-min window
  "observed_bps_5m":  10000000,                // publisher RX or subscriber TX
  "expected_bps_5m":  10000000,                // subscribers only; null otherwise
  "rate_status":       "reconciled",            // reconciled | mismatch | unknown
  "rate_status_reason": "active",               // active/idle/no_data (publisher) or
                                                // reconciled/mismatch/monitoring_gap/group_idle (subscriber)
  "control_plane_status": "healthy",            // CP-only verdict (drill-in)
  "health_status": "healthy"                    // COMBINED verdict per matrix — semantic change
}
```

The `health_status` field's semantic meaning is upgraded — it now means "healthy on both CP and rate" instead of "healthy on CP only". No JSON shape change to existing fields; the rate fields are additive.

## Stacking

`bc/mcast-health-rate-api` → `bc/mcast-health-rate-view` (#631).

## Testing Verification

- `insertMulticastHealthFixtures` now also inserts publisher RX = 10 Mbps and subscriber TX = 10 Mbps into `device_interface_rollup_5m`, exercising the round-trip of the new fields.
- `TestGetMulticastGroupHealth_Users` now asserts each row carries `control_plane_status = healthy`, `rate_status = reconciled`, combined `health_status = healthy`, and that observed/expected bps round-trip with the right reasons (`active` for publisher, `reconciled` for subscriber).
- All four existing health endpoint tests stay green; full `api/handlers` package run passes (`go test ./api/handlers` 27s).

## What's next

[...]

PR #633, opened 2026-06-04
multicast: health UI + rate reconciliation (infra#1501)

ben-malbeclabs:
## Summary

Multicast health visibility — UI + rate reconciliation. Consolidates the remaining slice of infra#1501 into a single branch (supersedes #629, #630, #631, #632).

Reviewable by commit (in order):

1. `8cb8075` **web**: multicast group Health tab + per-user health card (consumes the `/health` endpoints landed in #628).
2. `fa21d87` **web**: Health tab polish — legend, tooltips, nav arrows.
3. `c1eb701` **indexer**: `health_multicast_user_rate` view (per-user rate reconciliation: sum-of-publishers-RX vs. subscriber-TX from `device_interface_rollup_5m`, with ±5% / 1 Mbps tolerance and 15-min freshness window).
4. `1995140` **api**: read rate dimension from the new view and merge it into `/health/users` and `/users/:pk/health` responses.
5. `9230703` **web**: surface the rate dimension in the per-user health card and group Health tab.
6. `efddd9e` **web**: Radix tooltip (120 ms) replaces native `title=` (1 s) on the user health badge.
7. `9590d5f` **web**: hover-card on the per-user "Status" badge showing CP + Rate dimension verdicts side-by-side.
8. `42e50fa` **indexer** (vihu review): rate view joins on `user_pk`; P+S users reconciled subscriber-side with self's contribution subtracted (PIM-SM RPF).
9. `1c384b9` **api + web** (vihu review): `MulticastHealthStatusCounts.Unknown` explicit; per-user/path tables distinguish loading/error/empty; tooltip triggers keyboard-focusable.
10. `aef3288` **indexer**: gofmt fixup on the rate view test.
11. `76c44a4` **indexer** (codex review): rate view emits `multi_group_ambiguity` for tuples participating in more than one multicast group.
12. `3bf0a01` **web** (codex review): polling 30s → 60s; hide member filter on Health tab; drop `role="button"` from tooltip triggers; rewrite P+S help copy.
13. `48cf20f` **indexer** (vihu re-review): propagate publisher-side ambiguity to subscribers. `gpt_total_publisher_rx_bps` was contaminated by multi-group publishers' RX; now every subscriber/P+S row in such groups falls through to `multi_group_ambiguity` with `expected_bps_5m=NULL`. New test: `TestHealthMulticastUserRate_AmbiguousPublisherPropagates`.
14. `65381f1` **web** (vihu re-review): surface `missing_endpoint_reasons` on the per-path health badge via Radix tooltip when non-empty.

## Testing Verification

- Indexer view: rate tests cover healthy/degraded/freshness-expired/idle and four ambiguity scenarios — tunnel reuse via `user_pk`, P+S subscriber-side reconciliation with self subtracted, multi-group subscriber, and multi-group publisher contaminating a single-group subscriber's expected.
- API: `/health/users` and `/users/:pk/health` responses include `rate_*` fields end-to-end; `counts.unknown` surfaced.
- Web: verified per-user card and Health tab on staging — combined status badge resolves CP + Rate correctly; per-path missing-endpoint tooltip renders; hover-card renders both dimensions; tooltip latency is the expected 120 ms; keyboard nav reaches badges with focus ring; Health tab hides the inline member filter; multi-group rows display "unknown / multi_group_ambiguity" with an explanatory tooltip.

### DIFF
--- a/api/handlers/multicast_delivery.go
+++ b/api/handlers/multicast_delivery.go
@@ -2,6 +2,11 @@
 // back future per-group and per-user health endpoints (infra#1501 Track 2).
 // They classify mroute rows and surface (active_publisher, group, device)
 // gaps without requiring this package to recompute the joins.
+//
+// The health_multicast_user_rate view (migration 20260604000001) extends
+// the per-user health classification with a 5-min rate dimension, layering
+// data-plane reconciliation on top of the control-plane verdict (Track B1
+// of infra#1501). Endpoints reading from it are added in a follow-up.
 
 package handlers
 
--- a/api/handlers/multicast_health.go
+++ b/api/handlers/multicast_health.go
@@ -109,6 +109,8 @@ func addStatusCount(bucket *MulticastHealthStatusCounts, status string, n uint64
 		bucket.Degraded += n
 	case "unhealthy":
 		bucket.Unhealthy += n
+	case "unknown":
+		bucket.Unknown += n
 	}
 	bucket.Total += n
 }
--- a/api/handlers/multicast_health.go
+++ b/api/handlers/multicast_health.go
@@ -61,7 +61,7 @@ func (a *API) queryMulticastHealthCounts(ctx context.Context, groupPK string) (M
 		SELECT source, health_status, count() AS n FROM (
 			SELECT 'mroutes' AS source, health_status FROM health_mroute WHERE multicast_group_pk = ?
 			UNION ALL
-			SELECT 'users' AS source, health_status FROM health_multicast_user WHERE multicast_group_pk = ?
+			SELECT 'users' AS source, health_status FROM health_multicast_user_rate WHERE multicast_group_pk = ?
 			UNION ALL
 			SELECT 'paths' AS source, health_status FROM health_publisher_subscriber_path WHERE multicast_group_pk = ?
 		)
--- a/api/handlers/multicast_health_test.go
+++ b/api/handlers/multicast_health_test.go
@@ -46,6 +46,16 @@ func insertMulticastHealthFixtures(t *testing.T, api *handlers.API) {
 			'dev-nyc1', 'default', 'sparse', '233.0.0.1', '10.0.0.1',
 			'SMP', 0, 'Port-Channel1', '', '', 0, 0, '', 0, 0, '["Tunnel502"]', 1, now())`))
 
+	// Rate fixtures — publisher RX = 10 Mbps; subscriber TX = 10 Mbps → reconciled.
+	// Without these the rate dimension would collapse health_status to 'unknown'
+	// via 'no_data' for both endpoints.
+	require.NoError(t, api.DB.Exec(ctx, `
+		INSERT INTO device_interface_rollup_5m
+			(bucket_ts, device_pk, intf, user_tunnel_id, user_pk, max_in_bps, max_out_bps, ingested_at)
+		VALUES
+			(now() - INTERVAL 1 MINUTE, 'dev-ams1', 'Tunnel501', 501, 'user-pub', 10000000, 0, now()),
+			(now() - INTERVAL 1 MINUTE, 'dev-nyc1', 'Tunnel502', 502, 'user-sub', 0, 10000000, now())`))
+
 	require.NoError(t, api.DB.Exec(ctx, `SYSTEM REFRESH VIEW dz_device_interface_ips`))
 	require.NoError(t, api.DB.Exec(ctx, `SYSTEM WAIT VIEW dz_device_interface_ips`))
 }
@@ -124,11 +134,32 @@ func TestGetMulticastGroupHealth_Users(t *testing.T) {
 	require.Len(t, resp.Items, 2)
 	assert.EqualValues(t, 2, resp.Total)
 
-	// Both items should be reconciled and healthy.
-	for _, it := range resp.Items {
+	// Both items should be reconciled on both dimensions.
+	var pub, sub *handlers.MulticastHealthUserItem
+	for i := range resp.Items {
+		it := &resp.Items[i]
 		assert.True(t, it.Reconciled, "user %s expected reconciled", it.UserPK)
-		assert.Equal(t, "healthy", it.HealthStatus)
+		assert.Equal(t, "healthy", it.ControlPlaneStatus, "user %s expected CP healthy", it.UserPK)
+		assert.Equal(t, "reconciled", it.RateStatus, "user %s expected rate reconciled", it.UserPK)
+		assert.Equal(t, "healthy", it.HealthStatus, "user %s expected combined healthy", it.UserPK)
+		assert.NotNil(t, it.ObservedBps5m, "user %s missing observed rate", it.UserPK)
+		assert.NotNil(t, it.RateBucketTS, "user %s missing rate bucket ts", it.UserPK)
+		switch it.Mode {
+		case "P":
+			pub = it
+		case "S":
+			sub = it
+		}
 	}
+	// Publisher's rate reason is 'active' (transmitting). Subscriber's is 'reconciled'
+	// (TX matches sum of publishers).
+	require.NotNil(t, pub)
+	require.NotNil(t, sub)
+	assert.Equal(t, "active", pub.RateStatusReason)
+	assert.Equal(t, "reconciled", sub.RateStatusReason)
+	require.NotNil(t, sub.ExpectedBps5m)
+	assert.InDelta(t, 10_000_000, *sub.ExpectedBps5m, 1)
+	assert.InDelta(t, 10_000_000, *sub.ObservedBps5m, 1)
 }
 
 func TestGetMulticastGroupHealth_Paths(t *testing.T) {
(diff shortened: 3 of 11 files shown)


vihu:
A few UI/state concerns after rate health becomes visible:

1.  Unknown is now a real combined status, but summary counts only show healthy/degraded/unhealthy/total. I think unknown should be explicit.
2.  Some row tables use `data?.items ?? []`, which can show a false empty state while secondary queries are loading or failed.
3.  Nitpick really, tooltip triggers that carry health explanations should be keyboard-focusable, preferably buttons or explicit focusable triggers.

ben-malbeclabs:
@vihu thanks for the review — addressed both threads. Two new commits on top:

**`42e50fa` — rate view (your #631 thread):**
- Join now includes `user_pk` alongside `(device_pk, user_tunnel_id)` in `recent_bucket` and `user_rates`. A tunnel reassigned mid-bucket can no longer pull the wrong user's rate. Verified by `TestHealthMulticastUserRate_TunnelReuse` (two users on the same device sharing tunnel 777 each keep their own rate).
- P+S rate semantics are now defined explicitly: P+S is reconciled **subscriber-side**, with the publisher's own contribution subtracted from the expected sum (`expected = group_total − self.max_in_bps`). This matches PIM-SM RPF behaviour where a source does not receive its own multicast back via the tree. View header docstring expanded to spell this out. Verified by `TestHealthMulticastUserRate_PlusSubscriber`.

**`1c384b9` — API + UI (your #633 thread):**
1. `MulticastHealthStatusCounts` gains an `Unknown` field; `CountsRow` renders it explicitly so unknown-verdict rows are visible.
2. Per-user and per-path tables on the health tab use a new `TableStateRow` that distinguishes loading / error / empty — no more false-empty while a query is still in flight or has failed. User detail page surfaces a `loading…` / `health unavailable` chip next to each multicast group for the same reason.
3. Tooltip triggers on the rate cell, the per-user health badge (Health tab), and the user-detail health badge are now keyboard-focusable: `tabIndex={0}` + `role="button"` + `focus-visible` ring + descriptive `aria-label`.

ben-malbeclabs:
Codex review addressed across two commits.

**`76c44a4` — rate view multi-group ambiguity (#3, High):**
`device_interface_rollup_5m` keys on `(device_pk, user_tunnel_id, user_pk)` and does NOT split rate by multicast group, so when a user-tunnel participates in more than one group the observed `max_in_bps` / `max_out_bps` is a cross-group aggregate. The view now detects this via a `user_group_counts` CTE and emits `rate_status='unknown'` with `rate_status_reason='multi_group_ambiguity'` and `expected_bps_5m=NULL`. Pure idle (`max_in_bps = 0` on a publisher in multi-group) still flows through as `'idle'` since "publishing nothing globally" entails "publishing nothing per-group" — that statement remains true even with shared tunnels. UI types/help/badge maps gain the new reason; tooltip surfaces an operator-friendly explanation. Test: `TestHealthMulticastUserRate_MultiGroupAmbiguity` (one subscriber on two groups, per-tunnel max_out = 10 Mbps, neither group's row reconciles).

**`3bf0a01` — UI tweaks:**

- **#1 (High, partial)**: bumped `refetchInterval` 30s → 60s on the three Health-tab queries. The proper SQL-side pagination (with full severity ORDER BY + virtual/infinite scroll on the client) belongs in a follow-up issue — we explicitly didn't want to silently `LIMIT` reconciliation rows since the truncated tail is exactly where actionable rows sit. The 60s polling is the cheap, no-regrets win for this PR.
- **#2 (Medium)**: hide the member filter bar when `activeTab === 'health'`. The Health tab always reports across the full group; the filter would otherwise mislead operators into thinking it scopes the health queries.
- **#5 (Low)**: dropped `role="button"` from the three tooltip span triggers — the badge isn't an action target, only a focusable info trigger. `tabIndex={0}` + `aria-label` stay; Radix handles the screen-reader description via `aria-describedby`.
- **#6 (Low)**: rewrote the misleading "A user in P+S mode contributes two rows" help copy — the rate view emits one combined row per P+S user that reconciles subscriber-side against (sum of publishers − self).

**Deferred:** #4 (rendering `missing_endpoint_reasons` as an explicit column) — the data is already surfaced in the per-path tooltip; explicit column is nice-to-have. The full #1 fix (SQL-side pagination with severity sort + lazy load) will become a follow-up issue once we have a feel for production row volumes.

vihu:
> Codex review addressed across two commits.
> 
> **`76c44a4` — rate view multi-group ambiguity (#3, High):** `device_interface_rollup_5m` keys on `(device_pk, user_tunnel_id, user_pk)` and does NOT split rate by multicast group, so when a user-tunnel participates in more than one group the observed `max_in_bps` / `max_out_bps` is a cross-group aggregate. The view now detects this via a `user_group_counts` CTE and emits `rate_status='unknown'` with `rate_status_reason='multi_group_ambiguity'` and `expected_bps_5m=NULL`. Pure idle (`max_in_bps = 0` on a publisher in multi-group) still flows through as `'idle'` since "publishing nothing globally" entails "publishing nothing per-group" — that statement remains true even with shared tunnels. UI types/help/badge maps gain the new reason; tooltip surfaces an operator-friendly explanation. Test: `TestHealthMulticastUserRate_MultiGroupAmbiguity` (one subscriber on two groups, per-tunnel max_out = 10 Mbps, neither group's row reconciles).
> 
> **`3bf0a01` — UI tweaks:**
> 
>     * **[build(deps): bump github.com/aws/aws-sdk-go-v2 from 1.41.0 to 1.41.1 #1](https://github.com/malbeclabs/lake/pull/1) (High, partial)**: bumped `refetchInterval` 30s → 60s on the three Health-tab queries. The proper SQL-side pagination (with full severity ORDER BY + virtual/infinite scroll on the client) belongs in a follow-up issue — we explicitly didn't want to silently `LIMIT` reconciliation rows since the truncated tail is exactly where actionable rows sit. The 60s polling is the cheap, no-regrets win for this PR.
> 
>     * **[build(deps): bump github.com/malbeclabs/doublezero from 0.0.0-20260127003248-a7973ff42f73 to 0.8.0 #2](https://github.com/malbeclabs/lake/pull/2) (Medium)**: hide the member filter bar when `activeTab === 'health'`. The Health tab always reports across the full group; the filter would otherwise mislead operators into thinking it scopes the health queries.
> 
>     * **[build(deps): bump github.com/aws/aws-sdk-go-v2/service/s3 from 1.93.1 to 1.95.1 #5](https://github.com/malbeclabs/lake/pull/5) (Low)**: dropped `role="button"` from the three tooltip span triggers — the badge isn't an action target, only a focusable info trigger. `tabIndex={0}` + `aria-label` stay; Radix handles the screen-reader description via `aria-describedby`.
> 
>     * **[traffic: add traffic charts page and unify detail views #6](https://github.com/malbeclabs/lake/pull/6) (Low)**: rewrote the misleading "A user in P+S mode contributes two rows" help copy — the rate view emits one combined row per P+S user that reconciles subscriber-side against (sum of publishers − self).
> 
> 
> **Deferred:** #4 (rendering `missing_endpoint_reasons` as an explicit column) — the data is already surfaced in the per-path tooltip; explicit column is nice-to-have. The full #1 fix (SQL-side pagination with severity sort + lazy load) will become a follow-up issue once we have a feel for production row volumes.

- The multi-group ambiguity guard appears to protect the current row’s user/tunnel, but `group_publisher_total` still sums active multi-group publisher tunnel rates into each group’s expected total. A single-group subscriber can therefore still compare against an ambiguous publisher aggregate. Could we either propagate ambiguous publishers into `gpt_*` and mark subscriber-side rows `multi_group_ambiguity`, or add a follow-up/test for multi-group publisher + single-group subscriber?

- I don’t see `missing_endpoint_reasons` surfaced in the current path table/tooltip; path rows still render only `HealthBadge` and `verification_method`. Maybe I’m missing it in `multicast-group-health-tab.tsx`?

ben-malbeclabs:
@vihu thanks — both legit, both addressed.

**`48cf20f` — multi-group publisher contamination:** you're right that the prior guard only protected the *current row's* own `(device, tunnel, user)`. `gpt_total_publisher_rx_bps` was still computed by summing every publisher's RX in the group, including publishers whose tunnel is shared across groups. A single-group subscriber would then reconcile against a contaminated total.

Fix: `group_publisher_total` now tracks `gpt_ambiguous_publisher_count` via a join with `user_group_counts`. When `> 0`, every subscriber/P+S row in that group falls through to `rate_status_reason='multi_group_ambiguity'` with `expected_bps_5m=NULL` — same machinery as the row-level guard, just driven by the per-group flag. View header expanded to describe both propagation paths (per-row and per-group).

New test: `TestHealthMulticastUserRate_AmbiguousPublisherPropagates` — one publisher in groups X and Y over a shared tunnel, one single-group subscriber on group X. Before this fix the subscriber would compare 5 Mbps TX against a contaminated 10 Mbps expected and mismatch; after, the row is correctly `unknown / multi_group_ambiguity`.

**`65381f1` — `missing_endpoint_reasons`:** you're right, I was wrong about it being in the tooltip. The API was returning the field but the path table only rendered `HealthBadge` + `verification_method`. The status badge on the path row is now wrapped in a Radix tooltip that lists the reasons when present (focusable trigger, aria-label).

vihu (approved):

PR #644, opened 2026-06-05
multicast: under-dev banner + field:value
  search on Health tab

ben-malbeclabs:
## Summary

Two adjustments to the Multicast Group Health tab, both follow-ups to #642 driven by today's live review:

1. **Under-development banner.** A yellow info banner now renders at the top of the Health tab calling out that the verdicts are work-in-progress and state-collect only covers jump devices today, so any user / publisher / subscriber on a non-jump device renders as `unhealthy` even when functioning normally. Operators have been reading those false-positives as real outages; the banner sets expectations until the view learns to distinguish "uncollected" from "missing".

2. **Field:value search syntax.** `?search=` on `/health/users` and `/health/paths` is now whitespace-tokenized. Each token is either `field:value` (case-insensitive field, free-form value) or a bare term (broad substring fallback, same as before). Multiple tokens AND together.

    Supported fields on per-user: `device`, `status`, `tunnel`, `mode`, `user`, `account`, `owner`, `pubkey`, `ip`, `health`.
    Supported fields on per-path: `publisher`, `subscriber`, `device`, `status`, `user`, `owner`, `pubkey`, `ip`, `health`.
    Unknown prefixes fall back to substring so a typo doesn't break the whole query.

    The search input placeholders advertise the new syntax (`device:NAME  status:unhealthy  tunnel:520  …`) and a native `title` hint lists all the field prefixes.

## Testing Verification

- New unit tests cover the tokenizer (`tokenizeHealthSearch`) and the clause builder (`buildHealthSearchClause`): bare terms, field-scoped tokens, multi-token AND, lowercased field prefixes, trailing-colon fallback, unknown-field fallback, argument ordering. All green.
- Existing integration tests on the handlers still pass.
- Manual: `?search=device:nyc001 status:unhealthy` on `/health/users` narrows to rows whose device matches `nyc001` AND whose `health_status` matches `unhealthy`. Bare-token syntax (`?search=3b2Ze7VY`) still works.

### DIFF
--- a/api/handlers/multicast_health.go
+++ b/api/handlers/multicast_health.go
@@ -221,24 +221,80 @@ func parseHealthSearch(r *http.Request) string {
 	return strings.TrimSpace(r.URL.Query().Get("search"))
 }
 
-// buildHealthSearchClause builds a parameterized WHERE-LIKE OR across the
-// passed columns. Returns the SQL fragment (starting with " AND (...)" so
-// it appends cleanly) and the args to extend the query's argument slice.
-// Empty `search` returns ("", nil) so callers can unconditionally append.
-func buildHealthSearchClause(search string, cols []string) (string, []any) {
-	if search == "" || len(cols) == 0 {
+// healthSearchTokens splits a search string into whitespace-separated tokens.
+// Each token is either `field:value` (case-insensitive field, free-form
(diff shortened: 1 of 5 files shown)


ben-malbeclabs:
Folded into #643

PR #659, opened 2026-06-12
shreds: fix validator rewards page multi-client grain and claim status

SinaVafadar:
## Summary

Fixes the validator rewards detail page across four issues: per-validator earnings were understated for multi-client validators, the per-epoch claim status was unreliable, the validator's share was overstated, and the Earned 2Z amount was inflated by the wrong denominator.

- **Multi-client grain fix.** A validator publishing under more than one software client in a single epoch produces one merkle leaf per client. The leaf and leaf-distribution-status tables keyed on `(subscription_epoch, node_id)` only, so those leaves collapsed into a single row — understating leader slots/earnings and letting one client's claimable bit clobber another's. Both grains now include `client_id`; a migration rebuilds the affected tables.
- **Derived claim lifecycle state.** Each epoch row carries an explicit `state` — `claimable`, `distributed` (Paid), `pending`, or `unknown` — instead of a bare nullable `is_claimable` bit. A funded epoch with no per-validator status row (journal swept before tracking began) is reported as `unknown` rather than assumed Paid.
- **Client-proportion default fix.** The per-client reward proportion is stored as `0` when unset (real fallback in `default_proportion`), and a ClickHouse `LEFT JOIN` fills a missing row with `0`, not `NULL`. So `coalesce(proportion, default_proportion, 3500)` returned `0`, giving the validator the whole pool share instead of ~65%. A `0` now falls through to `default_proportion`, then `3500`.
- **Authoritative leader-slot denominator.** Earnings split the pool by `leader_slots / total_leader_slots`, but the denominator was `sum(leader_slots)` over the *indexed* leaves — incomplete for older epochs, so every present validator was over-credited by a per-epoch constant factor. The `ShredDistributionJournal`'s own `total_leader_slots` (byte offset 128) is now decoded, stored on the 2Z-pool row, and used as the denominator (falling back to summed leaves when not yet populated).

## Validation

Cross-referenced against actual on-chain 2Z distribution amounts. Before the denominator fix, computed earnings were a constant multiple of actual *within* each epoch (e.g. epoch 976: every validator exactly 1.073× actual; 975: 1.146×; 974: 1.242×) — the signature of a shared, too-small denominator. The factor tracked the missing-validator gap (epoch 975 coverage 87.3% ⇒ 1/0.873 ≈ 1.146). Dividing by the journal's `total_leader_slots` removes it.

## Known limitations

- **Old swept epochs**: epochs whose journal was swept before this change keep `total_leader_slots = 0` (can't be re-fetched on-chain), so they fall back to the summed-leaves denominator and remain approximate. Epochs from the deploy forward are exact.
- **Paid status** is derived from the publisher accumulation bitmap; a cleared bit means distributed, but the bitmap doesn't encode *why* it cleared, so a forfeited/expired claim can read as Paid in rare cases. The authoritative source (actual distribution transactions) isn't indexed in this repo — a tracked follow-up.

## Testing Verification

- Earnings sum across multiple client leaves in one epoch; only the claimable client's share counts as immediately-claimable (`TestGetShredsRewards_MultiClientValidator`).
- Unset/zero client proportion falls through to `default_proportion` (65% share), not a 100% share (`TestGetShredsRewards_ProportionDefaultsWhenUnset`).
- Earnings divide by the stored `total_leader_slots`, not the summed leaves, when present (`TestGetShredsRewards_UsesJournalTotalLeaderSlots`).
- `total_leader_slots` decodes from journal byte offset 128 (`TestDecodeJournalAccount_ValidLayout`).
- Claim state resolves across the lifecycle: live claimable bit → `claimable`, cleared bit → `distributed`, funded-but-no-status → `unknown`, unfunded pool → `pending` (`TestGetShredsRewardsDetail_*`).

### DIFF
--- a/api/handlers/shreds_rewards_test.go
+++ b/api/handlers/shreds_rewards_test.go
@@ -98,14 +98,14 @@ func insertShredsRewardsTestData(t *testing.T, api *handlers.API) {
 	err = api.DB.Exec(ctx, `
 		INSERT INTO dim_dz_shred_validator_leaf_distribution_status_history
 		(entity_id, snapshot_ts, ingested_at, op_id, is_deleted, attrs_hash,
-		 subscription_epoch, node_id, is_claimable, journal_mint_key)
+		 subscription_epoch, node_id, client_id, is_claimable, journal_mint_key)
 		VALUES
-		('s-100-A', now(), now(), generateUUIDv4(), 0, 1, 100, 'node-A', 0, ''),
-		('s-100-B', now(), now(), generateUUIDv4(), 0, 2, 100, 'node-B', 0, ''),
-		('s-101-A', now(), now(), generateUUIDv4(), 0, 3, 101, 'node-A', 0, ''),
-		('s-101-B', now(), now(), generateUUIDv4(), 0, 4, 101, 'node-B', 0, ''),
-		('s-102-A', now(), now(), generateUUIDv4(), 0, 5, 102, 'node-A', 1, 'journal-A'),
-		('s-102-B', now(), now(), generateUUIDv4(), 0, 6, 102, 'node-B', 1, 'journal-B')
+		('s-100-A', now(), now(), generateUUIDv4(), 0, 1, 100, 'node-A', 1, 0, ''),
+		('s-100-B', now(), now(), generateUUIDv4(), 0, 2, 100, 'node-B', 1, 0, ''),
+		('s-101-A', now(), now(), generateUUIDv4(), 0, 3, 101, 'node-A', 1, 0, ''),
+		('s-101-B', now(), now(), generateUUIDv4(), 0, 4, 101, 'node-B', 1, 0, ''),
+		('s-102-A', now(), now(), generateUUIDv4(), 0, 5, 102, 'node-A', 1, 1, 'journal-A'),
+		('s-102-B', now(), now(), generateUUIDv4(), 0, 6, 102, 'node-B', 1, 1, 'journal-B')
 	`)
 	require.NoError(t, err)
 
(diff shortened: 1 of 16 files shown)


indexer/pkg/dz/shreds/validatorrewards/token.go
```diff
@@ -0,0 +1,40 @@
...
+// TokenInfoForMint returns the (symbol, decimals) for a known reward mint and a
+// bool reporting whether the mint is recognized. Unknown mints fall back to the
+// 2Z scale so a newly-added token never silently produces NaN amounts, but the
+// ok=false return lets callers log/flag it.
+func TokenInfoForMint(mint string) (TokenInfo, bool) {
+	if info, ok := tokenInfoByMint[mint]; ok {
+		return info, true
+	}
+	return TokenInfo{Symbol: "?", Decimals: 8}, false
+}
```
martinsander00: `TokenInfo`, `tokenInfoByMint`, and `TokenInfoForMint` aren't referenced anywhere. The API does its own symbol/scale mapping in SQL (`tokenSymbolSQL`/`tokenScaleSQL`) and the indexer just uses `RewardMintKey.String()`, so this whole block is dead — it only compiles because they're exported.

Either drop it (keep the `USDCMintKey`/`WSOLMintKey` consts, those are used) or wire it in instead of the hand-rolled SQL. I'd lean toward deleting for now.

indexer/db/clickhouse/migrations/20260611000001_shred_rewards_client_grain.sql
```diff
@@ -0,0 +1,140 @@
...
+TRUNCATE TABLE IF EXISTS dim_dz_shred_validator_rewards_leaves_history;
+TRUNCATE TABLE IF EXISTS stg_dim_dz_shred_validator_rewards_leaves_snapshot;
+
+-- The leaf distribution status (claimable bit) shares the same defect: the
+-- publisher-accumulation bitmap is per leaf (i.e. per node+client), but the
+-- table keyed per (subscription_epoch, node_id), so one client's bit could
+-- overwrite another's. Extend the grain to include client_id. We drop and
+-- recreate (rather than ALTER ... ADD COLUMN) so client_id lands in the correct
+-- column position for the dim-type-2 positional batch insert. The table is
+-- window-bounded (~12 most-recent epochs) and repopulated every refresh, so the
```
martinsander00: This `~12 most-recent epochs` wording is stale as of this same PR — view.go drops the `recentJournalWindow = 12` and now projects every un-swept journal. The table is bounded by on-chain journal liveness now, not a 12-epoch window. Can you reword?

indexer/db/clickhouse/migrations/20260616000001_shred_pool_reward_token.sql
```diff
@@ -0,0 +1,96 @@
+-- +goose Up
+
+-- Clear the leaf distribution status history so the per-leaf reward-token
+-- attribution rebuilds cleanly. The pre-multi-token projection marked any
+-- cleared 2Z-journal bit in the accumulation prefix as a distributed 2Z leaf;
+-- for epochs ≥ 968 that mistagged USDC/wSOL leaves as 2Z. The new projection
+-- only marks a leaf distributed for the token it was previously attributed to,
+-- so the stale 2Z rows must be cleared or they would perpetuate. The table is
+-- window-bounded (~12 recent epochs) and repopulated every indexer refresh, so
```
martinsander00: Same stale `~12 recent epochs` wording here — we removed that window in this PR. It's now bounded by which journals are still readable on-chain.

api/handlers/shreds_rewards_test.go
```diff
@@ -347,6 +355,245 @@ func TestGetShredsRewards_DeduplicatesValidators(t *testing.T) {
...
+	require.Len(t, resp.Validators, 1)
+	// 2925 (denominator 200), NOT 5850 (summed-leaves denominator 100).
+	assert.InDelta(t, 2925.0, resp.Validators[0].TotalEarned2Z, 1e-6,
+		"must divide by the journal's total_leader_slots, not the summed indexed leaves")
+}
+
+// TestGetShredsRewards_MultiTokenUSDC covers the multi-token era: a validator
+// rewarded in USDC. Its leaf is attributed to the USDC mint via the status row's
+// journal_mint_key, joined to the USDC pool, and split over the USDC journal's
+// accumulated_slots_scaled denominator (not the epoch total), with the 10% burn
```
martinsander00: This says `with the 10% burn`, but the body right below says `NO burn` and the asserted 585 doesn't apply the ×0.9 — which is correct, USDC isn't burned. Drop the `10% burn` from this line so it stops contradicting itself.

api/handlers/shreds_rewards_test.go
```diff
@@ -191,29 +196,32 @@ func TestGetShredsRewards_GoldenPath(t *testing.T) {
...
 	a := resp.Validators[0]
 	b := resp.Validators[1]
 	assert.Equal(t, "node-A", a.NodeID)
 	assert.Equal(t, "Alpha Validator", a.ValidatorName)
 	assert.Equal(t, "vote-A", a.VotePubkey)
 	assert.Equal(t, uint64(5000000000), a.ActivatedStake)
 	assert.Equal(t, "203.0.113.10", a.DZUserIP)
-	assert.InDelta(t, 11700.0, a.TotalEarned2Z, 1e-6)
-	assert.InDelta(t, 3900.0, a.ImmediatelyClaimable2Z, 1e-6)
+	assert.InDelta(t, 10530.0, a.TotalEarned2Z, 1e-6)
```
martinsander00, line 208: While you're here — nothing asserts `resp.Total` anywhere, so the new count query is completely uncovered. Mind adding `assert.Equal(t, 2, resp.Total)` (or similar) in this test?

martinsander00 (approved):

PR #660, opened 2026-06-15
indexer: chunk telemetry-usage influxdb query to bound heap

ben-dz:
## Reason
Address indexer-error alerting:  https://doublezero.grafana.net/alerting/grafana/lake-indexer-errors/view?orgId=1

## Summary of Changes
- Chunk the `RefreshTelemetryUsage` InfluxDB query: instead of one Flux `pivot()` over the full `[queryStart, now]` window, split into contiguous sub-windows of at most `QueryChunk` (default 5m) and concatenate the rows.
- Fixes continuous `RefreshTelemetryUsage` failures in `lake-staging` (testnet + mainnet-beta) where the query hit InfluxDB Cloud's heap limit (`request too large … 'deduplicate batches': Heap exhausted`) or timed out (`context deadline exceeded`).
- Breaks the self-reinforcing failure loop: a failed query writes nothing, so the ClickHouse high-water mark freezes and the window grows to the `QueryWindow` (1h) cap, making every subsequent full-hour pivot too large to ever succeed. Bounded chunks let the activity recover on its own.

## Diff Breakdown
| Core logic   |     1 | +36 / -1    | +35  |
| Tests        |     1 | +93 / -0    | +93  |
| **Total**    |     2 | +129 / -1   | +128 |

Small, focused change: one core-logic helper plus a config default, backed by pure unit tests.

<details>

- `indexer/pkg/dz/telemetry/usage/view.go` — added `queryIntfCountersChunked` (splits `[start, end)` into ≤`QueryChunk` half-open sub-windows and concatenates rows); added `QueryChunk` to `ViewConfig` with a 5m default in `Validate`; routed `queryInfluxDB` through the helper.
- `indexer/pkg/dz/telemetry/usage/chunk_test.go` — regression tests for window splitting, remainder clamping, and error propagation.

</details>

## Testing Verification
- `_SplitsWindow`: a 1h window produces 12 sub-queries, each ≤ chunk size, contiguous, exactly covering `[start, end)` — this is the assertion that fails against the old single-shot query.
- `_RemainderWindow`: 12m window → 3 chunks with the last clamped to the 2m remainder (no overshoot past `end`).
- `_PropagatesError`: a failing chunk aborts immediately and propagates the error.
- Sub-windows exactly partition `[start, end)` (half-open, matching Flux `range`), so the concatenated result is identical to the prior single query and downstream time-sort/forward-fill is unchanged. Steady-state windows are a few minutes (one chunk), so no behavior change in the common case.
- Full `dztelemusage` package tests pass against the ClickHouse testcontainer.

### DIFF
--- a/indexer/pkg/dz/telemetry/usage/chunk_test.go
+++ b/indexer/pkg/dz/telemetry/usage/chunk_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	laketesting "github.com/malbeclabs/lake/utils/pkg/testing"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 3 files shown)


juan-malbeclabs (commented):
**`BackfillForTimeRange` bypasses the chunking and shares the exact heap-exhaustion failure mode.**

The steady-state refresh now goes through `queryIntfCountersChunked`, but the backfill path still issues a single unbounded query (`store_backfill.go:54`):

```go
rows, err := v.cfg.InfluxDB.QueryIntfCounters(ctx, startTimeUTC, endTimeUTC)
```

Backfill is the path that takes arbitrary, potentially multi-hour/multi-day ranges — i.e. the case *most* likely to trigger the "deduplicate batches: Heap exhausted" pivot this PR is meant to prevent. The refresh path is already bounded to `QueryWindow` (1h → 12 chunks), so ironically the larger, riskier query is the one left unchunked.

`BackfillForTimeRange` has no callers or tests today, so it isn't a live bug, but it's exported API with the same failure mode. Suggest routing it through `queryIntfCountersChunked(ctx, v.cfg.InfluxDB, startTimeUTC, endTimeUTC, v.cfg.QueryChunk)` too, so the heap protection holds regardless of which path reaches InfluxDB.

juan-malbeclabs (approved):
LGTM

PR #663, opened 2026-06-17
scoreboard: surface turbine-root in DZ lead-time and publishing stats

juan-malbeclabs:
## Summary

- Lead-time (p50/p95) of DZ vs Turbine/Jito now pools the full DZ delivery path — leader (`dz`), turbine-root (`edge-solana-root`) and regional retransmit — so the advantage metric matches the win-rate's `dz_edge` aggregate instead of excluding root. Applies to both leaders-only and all-slots modes.
- Adds two headline cards to the shreds scoreboard: **Turbine-Root Publishing** (count of validators publishing via the turbine-root path) and **Turbine-Root Stake Weight** (% of network stake behind them), computed the same way as the existing leader publishing/stake metrics.
- Publisher check now derives per-publisher turbine-root participation (`root_slots` / `publishing_root`) from the `edge-solana-root` feed in `publisher_shred_stats`.

## Why

The scoreboard already counts turbine-root wins inside `dz_edge`, but the lead-time stat silently excluded root, so the two DZ metrics measured different feed sets. Pooling them makes "how much DZ wins" and "by how much" consistent. The new cards give the same leader-slot stake visibility for the turbine-root path.

## Notes

- The turbine-root cards read **0 until the shredder is configured to emit per-publisher stats for the `edge-solana-root` source** (tracked in the infra repo). The feed string lives in a single named constant (`rootPublisherFeed`) for easy realignment.

## Testing Verification

- Existing `TestGetEdgeScoreboard_WithData` still passes: it only seeds `feed='dz'` lead-time rows, so widening the q2b feed set adds no rows and `root_slots` stays 0 with no leader-metric drift.
- Verified the `root_stats` CTE shares the same epoch/slot window as the leader stats and dedups by slot, so multi-host observations don't double-count.

### DIFF
--- a/api/handlers/edge_scoreboard.go
+++ b/api/handlers/edge_scoreboard.go
@@ -102,24 +102,26 @@ type EdgeScoreboardLeader struct {
 
 // EdgeScoreboardResponse is the response for the edge scoreboard endpoint.
 type EdgeScoreboardResponse struct {
-	Window             string                           `json:"window"`
-	LeadersOnly        bool                             `json:"leaders_only"`
-	GeneratedAt        time.Time                        `json:"generated_at"`
-	CurrentEpoch       uint64                           `json:"current_epoch"`
-	CurrentSlot        uint64                           `json:"current_slot"`
-	TotalSlots         uint64                           `json:"total_slots"`
-	GlobalTotalSlots   uint64                           `json:"global_total_slots"`
-	DZSlots            uint64                           `json:"dz_slots"`
-	TotalDZLeaderSlots uint64                           `json:"total_dz_leader_slots"`
-	CompletenessPct    float64                          `json:"completeness_pct"`
-	PublisherCount     uint64                           `json:"publisher_count"`
-	PublishingCount    uint64                           `json:"publishing_count"`
-	PublishingStakePct float64                          `json:"publishing_stake_pct"`
-	Nodes              []EdgeScoreboardNode             `json:"nodes"`
-	RecentSlots        []EdgeScoreboardSlotRace         `json:"recent_slots"`
-	SlotBuckets        []EdgeScoreboardSlotBucket       `json:"slot_buckets,omitempty"`
-	SlotBucketSize     uint64                           `json:"slot_bucket_size,omitempty"`
-	SlotLeaders        map[string]*EdgeScoreboardLeader `json:"slot_leaders,omitempty"`
+	Window                 string                           `json:"window"`
+	LeadersOnly            bool                             `json:"leaders_only"`
+	GeneratedAt            time.Time                        `json:"generated_at"`
+	CurrentEpoch           uint64                           `json:"current_epoch"`
+	CurrentSlot            uint64                           `json:"current_slot"`
+	TotalSlots             uint64                           `json:"total_slots"`
+	GlobalTotalSlots       uint64                           `json:"global_total_slots"`
+	DZSlots                uint64                           `json:"dz_slots"`
+	TotalDZLeaderSlots     uint64                           `json:"total_dz_leader_slots"`
+	CompletenessPct        float64                          `json:"completeness_pct"`
+	PublisherCount         uint64                           `json:"publisher_count"`
+	PublishingCount        uint64                           `json:"publishing_count"`
+	PublishingStakePct     float64                          `json:"publishing_stake_pct"`
+	RootPublishingCount    uint64                           `json:"root_publishing_count"`
+	RootPublishingStakePct float64                          `json:"root_publishing_stake_pct"`
+	Nodes                  []EdgeScoreboardNode             `json:"nodes"`
+	RecentSlots            []EdgeScoreboardSlotRace         `json:"recent_slots"`
+	SlotBuckets            []EdgeScoreboardSlotBucket       `json:"slot_buckets,omitempty"`
+	SlotBucketSize         uint64                           `json:"slot_bucket_size,omitempty"`
+	SlotLeaders            map[string]*EdgeScoreboardLeader `json:"slot_leaders,omitempty"`
 	// DataLagMs is how far behind wall clock the latest row in slot_feed_race_summary_v2 is
 	// (server now − max(event_ts)). The client adds this to its own queue depth to show a
 	// pill reflecting actual on-chain time.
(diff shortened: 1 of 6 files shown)


api/handlers/edge_scoreboard.go
```diff
@@ -327,6 +370,205 @@ func (a *API) GetEdgeScoreboard(w http.ResponseWriter, r *http.Request) {
...
+	ctx, cancel := context.WithTimeout(r.Context(), 60*time.Second)
+	defer cancel()
+
+	window := strings.TrimSpace(r.URL.Query().Get("window"))
+	if _, ok := validWindows[window]; !ok {
+		window = "1h"
+	}
+	leadersOnly := strings.TrimSpace(r.URL.Query().Get("leaders_only")) != "false"
+
+	resp, err := a.FetchEdgeScoreboardMatrix(ctx, window, leadersOnly)
```
ben-dz, line 387: The matrix endpoint has no page-cache layer (unlike the sibling `GetEdgeScoreboard`), and `window=all` is a valid `validWindows` key mapping to `""` with no `slotsPerWindow` entry — so `timeFilter` and `rangeFilter` both collapse to empty and the query runs `SETTINGS final=1` over the entire `slot_feed_race_summary_v2` table (plus, in leaders-only mode, an unbounded `SELECT DISTINCT slot FROM publisher_shred_stats`). The route is public (`OptionalAuth`, only per-IP rate limiting) and directly callable regardless of the client `?matrix=1` flag. Reject/clamp `window=all` here (or cap to `30d`), add a page-cache entry for the default view, and/or attach a ClickHouse `max_execution_time`/`max_bytes_to_read` setting.

api/handlers/edge_scoreboard.go
```diff
@@ -327,6 +370,205 @@ func (a *API) GetEdgeScoreboard(w http.ResponseWriter, r *http.Request) {
...
+func (a *API) FetchRootPublishingStats(ctx context.Context) (count uint64, stake int64, err error) {
+	const slotsPerEpoch = 432_000
+	shredderDB := fmt.Sprintf("`%s`", a.ShredderDB)
+
+	query := fmt.Sprintf(`
+		WITH
+		root_slots AS (
+			SELECT DISTINCT slot
+			FROM %[1]s.slot_feed_race_summary_v2
+			WHERE feed_type = 'shred' AND feed = %[2]s
```
ben-dz, line 548: `root_slots` filters only on `feed` + `epoch`, so it can't use the `(host, slot, …)` primary index and scans the full monthly partition. This runs in Group P on every live `FetchEdgeScoreboardData` call (including uncached non-default windows), inside the deadline budget the file's comments say was already being exhausted in prod. Add a prunable slot bound, e.g. `AND slot >= ((SELECT max(epoch) FROM %[1]s.slot_feed_race_summary_v2 WHERE feed_type='shred') - 1) * 432000`.

api/handlers/edge_scoreboard.go
```diff
@@ -327,6 +370,205 @@ func (a *API) GetEdgeScoreboard(w http.ResponseWriter, r *http.Request) {
...
+				AND epoch >= (SELECT max(epoch) FROM %[1]s.slot_feed_race_summary_v2 WHERE feed_type = 'shred') - 1
```
ben-dz, line 549: Two independent `max(epoch)` anchors — this one over `slot_feed_race_summary_v2` and the one at line 554 over `dim_solana_leader_schedule_history` — can skew the root count for the interval after an epoch boundary where the shredder has advanced but the schedule view hasn't refreshed (newest root slots go unmapped, self-heals next refresh). Compute one epoch anchor in Go and apply it to both CTEs. Separately, this subquery is missing the `slot < 1000000000000` (maxValidSlot) corruption guard used on every other max() in this file — a garbage 2^64 epoch would make `max(epoch)-1` huge and silently zero root stats permanently.

api/handlers/edge_scoreboard.go
```diff
@@ -327,6 +370,205 @@ func (a *API) GetEdgeScoreboard(w http.ResponseWriter, r *http.Request) {
...
+		sched AS (
+			SELECT toUInt64(epoch * %[3]d + arrayJoin(JSONExtract(slots, 'Array(UInt64)'))) AS slot, node_pubkey
+			FROM dim_solana_leader_schedule_history
+			WHERE epoch >= (SELECT max(epoch) FROM dim_solana_leader_schedule_history) - 1
+		),
+		root_leaders AS (
+			SELECT DISTINCT s.node_pubkey AS node_pubkey
+			FROM sched s INNER JOIN root_slots r ON s.slot = r.slot
+		)
+		SELECT count(), COALESCE(sum(v.activated_stake_lamports), 0)
```
ben-dz, line 560: `count()` counts vote-account rows, not validators: `solana_vote_accounts_current` is deduped per `vote_pubkey`, not `node_pubkey`, so a node with two `epoch_vote_account='true'` accounts counts as 2 and sums both stakes. Use `count(DISTINCT rl.node_pubkey)` and decide the stake aggregation explicitly.

api/handlers/edge_scoreboard.go
```diff
@@ -846,9 +1102,10 @@ func (a *API) FetchEdgeScoreboardData(ctx context.Context, window string, leader
 		// Group A2: pairwise lead times (q2b) for the synthetic dz_edge feed vs competitors.
 		// Runs in parallel with Group A so it is never starved by q2+q2c consuming
 		// the 45s budget, which caused intermittent empty lead-time columns.
-		// Leaders-only uses feed='dz' (leader path only). All-slots pools dz + regional
-		// retransmit feeds so retransmit wins are represented — direct quantile over the
-		// combined rows (no per-slot argMin CTE, which was too expensive over the full table).
+		// Both modes pool the full DZ path set (dz + turbine-root + regional retransmit) so the
+		// p50/p95 reflects whichever DZ path arrived first — matching the win-rate's dz_edge.
```
ben-dz, line 1106: Comment claims the p50/p95 reflects "whichever DZ path arrived first," but the query pools up to ~6 rows per slot (one per winning DZ variant) into `quantile(...)` grouped by `(host, loser_feed)` — a distribution over all winning paths, not the per-slot fastest. `count() AS slot_count` (lines 1116/1131) likewise now counts pooled rows rather than slots (field currently unused by the web UI). Correct the comment to describe the pooled approximation and use `uniqExact(slot)` for `slot_count`.

web/src/components/edge-scoreboard-page.tsx
```diff
@@ -1951,6 +1953,134 @@ function RecentSlotsChart({
...
+                          {cell}
+                        </Tooltip>
+                      )
+                    })}
+                  </tr>
+                ))}
+              </tbody>
+            </table>
+            <p className="text-[11px] text-muted-foreground mt-3">
+              <span className="text-amber-400">*</span> only one direction recorded in the source data; the reverse race isn't tracked, so the shown rate is a lower bound rather than a true head-to-head.
```
ben-dz, line 2075: For a forward-only (incomplete) cell the win rate is `wins/(wins+0)=100%`, which is an upper bound (true rate ≤ shown), but this footnote and the incomplete-cell tooltip call it a lower bound; only the reverse `0%` cell is a genuine lower bound. The Go doc comment says incomplete cells should render as n/a. Fix the wording or suppress the percentage for incomplete cells.

ben-dz (commented):
No critical or high issues. Two Mediums: (1) the new `/api/dz/edge/scoreboard/matrix` endpoint is uncached and accepts `window=all`, so an unauthenticated request can trigger a full-table pairwise scan of `slot_feed_race_summary_v2`; add a cache or reject/clamp `window=all`. (2) the `root_slots` CTE in `FetchRootPublishingStats` filters only on feed+epoch (no slot bound), so it can't use the primary index and full-scans the monthly partition on every non-cached scoreboard fetch, in a deadline budget the file already documents as tight. Plus several Lows around the dual `max(epoch)` anchors, a missing corruption guard, `count()` vs distinct validators, inverted matrix tooltip copy, and a q2b comment/`slot_count` mismatch. No SQL-injection path — `window` is allowlisted and `leaders_only` is a bool.

juan-malbeclabs:
I wanted to show how much stake the root covered, but since each validator was also receiving shreds from other validators, the number ended up being close to 99%, which didn't really add any value.

PR #666, opened 2026-06-20
shreds: flag lagging publishers on the shred publishers page

SinaVafadar:
## Summary

- Adds a **Lagging** column to the Shred Publishers page (`/dz/shreds/publishers`) that flags validators reported as not `Healthy` in `dzf_data.lagged_validators`.
- The lagging status text (the non-`Healthy` value from the table) is shown in the cell and on hover, and the column is sortable.
- Wires up the `dzf_data` ClickHouse database (config default, `CLICKHOUSE_DZF_DATA_DB` env override, and API plumbing) so the handler can read the table.

## Details

`FetchPublisherCheckData` now joins publishers against `dzf_data.lagged_validators` by `validator_vote_public_key`, selecting rows where `status != 'Healthy'`. Matching publishers are annotated with `lagging` and `lagging_status`. The lookup is non-fatal — if the table is unavailable the page still renders without lagging annotations. Because the logic lives in `FetchPublisherCheckData`, the background page cache picks up the new fields automatically.

## Testing Verification

- Existing `PublisherCheck` handler tests pass; the lagged-validators lookup degrades gracefully when the table is absent in the test DB (logged warning, publishers returned unannotated).

### DIFF
--- a/api/config/config.go
+++ b/api/config/config.go
@@ -35,6 +35,9 @@ var publisherDB = "shredder"
 // dzdpDB is the ClickHouse database name for DZDP tables e.g. location_state (default: "dzdp").
 var dzdpDB = "dzdp"
 
+// dzfDataDB is the ClickHouse database name for dzf_data tables (default: "dzf_data").
+var dzfDataDB = "dzf_data"
+
 // EnvDBs maps environment names to their ClickHouse connection pools.
 // The mainnet-beta entry always points to DB.
 var EnvDBs map[string]driver.Conn
@@ -81,6 +84,16 @@ func GetDZDPDB() string {
 // SetDZDPDB sets the DZDP database name.
 func SetDZDPDB(db string) {
 	dzdpDB = db
+}
+
+// GetDZFDataDB returns the dzf_data database name.
+func GetDZFDataDB() string {
+	return dzfDataDB
+}
+
+// SetDZFDataDB sets the dzf_data database name.
+func SetDZFDataDB(db string) {
+	dzfDataDB = db
 }
 
 // Database returns the configured database name.
@@ -154,6 +167,10 @@ func Load() error {
 		dzdpDB = db
 	}
 
+	if db := os.Getenv("CLICKHOUSE_DZF_DATA_DB"); db != "" {
+		dzfDataDB = db
+	}
+
 	// Build env -> database mapping.
 	// Devnet and testnet databases default to lake_devnet / lake_testnet
 	// unless overridden by env vars or disabled with CLICKHOUSE_NO_DEVNET / CLICKHOUSE_NO_TESTNET.
@@ -177,7 +194,7 @@ func Load() error {
 
 	secure := os.Getenv("CLICKHOUSE_SECURE") == "true"
 
-	slog.Info("connecting to ClickHouse", "addr", cfg.Addr, "database", cfg.Database, "username", cfg.Username, "secure", secure, "shredder_db", shredderDB, "publisher_db", publisherDB, "dzdp_db", dzdpDB)
+	slog.Info("connecting to ClickHouse", "addr", cfg.Addr, "database", cfg.Database, "username", cfg.Username, "secure", secure, "shredder_db", shredderDB, "publisher_db", publisherDB, "dzdp_db", dzdpDB, "dzf_data_db", dzfDataDB)
 
 	// Create connection pool
 	opts := &clickhouse.Options{
--- a/api/handlers/publisher_check.go
+++ b/api/handlers/publisher_check.go
@@ -67,6 +67,8 @@ type PublisherCheckItem struct {
 	ValidatorName           string `json:"validator_name"`
 	ValidatorVersionOk      bool   `json:"validator_version_ok"`
 	IsBackup                bool   `json:"is_backup"`
+	Lagging                 bool   `json:"lagging"`
+	LaggingStatus           string `json:"lagging_status"`
 }
 
 // PublisherCheckResponse is the response for the publisher check endpoint.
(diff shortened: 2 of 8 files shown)


web/src/components/publisher-check-page.tsx
```diff
@@ -974,6 +982,11 @@ export function PublisherCheckPage() {
                       <td className="px-4 py-3 text-center">
                         <StatusIcon ok={!pub.publishing_retransmitted} />
                       </td>
+                      <td className="px-4 py-3 text-center">
+                        <StatusIcon
+                          ok={pub.lagging_status !== "Action Needed"}
```
ben-dz, line 987: This predicate disagrees with the backend and the sort. The backend sets `lagging=true` for any `status != 'Healthy'` (publisher_check.go:386) and the sort comparator uses that boolean (line 417), but the icon is red only for the exact string `"Action Needed"` — a value that appears nowhere in the backend or the repo, from a table whose status vocabulary is undocumented. Any other non-Healthy value (or a casing change by the external writer) yields a row that sorts as lagging while showing a green check; rows with `status=''` are also flagged lagging with a green icon, and an `undefined` `lagging_status` during deploy skew renders green too. Use one predicate: `ok={!pub.lagging}` (also handles `undefined`), or filter the SQL to `status = 'Action Needed'` so the flag, sort, and icon agree.

api/handlers/publisher_check.go
```diff
@@ -358,3 +374,34 @@ func (a *API) FetchPublisherCheckData(ctx context.Context, q string, epochsParam
 		Publishers:          publishers,
 	}, nil
 }
+
+// fetchLaggedValidators returns a map of validator vote pubkey to its lagging
+// status for every validator currently reported as not Healthy in
+// dzf_data.lagged_validators.
+func (a *API) fetchLaggedValidators(ctx context.Context) (map[string]string, error) {
```
ben-dz, line 381: The success path of this function is never verified anywhere. The test harness (`api/testing/api.go`) never sets `DZFDataDB`, so in every existing test this builds ``FROM ` `.lagged_validators`` with an empty identifier and always errors — the tests cited in the PR only exercise the warn-and-degrade branch. Fix: set `DZFDataDB` in the harness constructors (e.g. to the per-test `dbName`), seed a `lagged_validators` fixture following the `createPublisherShredStatsTable` pattern with a Healthy row, a non-Healthy row matched to a fixture vote pubkey, and an unmatched pubkey, and assert `lagging`/`lagging_status` on the response — this also pins the status-string contract. Separately, the read-only prod ClickHouse user has no SELECT grant on `dzf_data.lagged_validators` (verified), so please confirm the deployed API user's grant before merge; if it's missing, the column ships permanently green with only a server-side warn every 30s.

api/handlers/publisher_check.go
```diff
@@ -358,3 +374,34 @@ func (a *API) FetchPublisherCheckData(ctx context.Context, q string, epochsParam
...
+	start := time.Now()
+	query := fmt.Sprintf(`
+		SELECT validator_vote_public_key, status
+		FROM `+"`%s`"+`.lagged_validators
+		WHERE status != 'Healthy' AND validator_vote_public_key != ''`, a.DZFDataDB)
```
ben-dz, line 386: This assumes the table holds only current state: no recency bound, no `FINAL`/`argMax` dedup, and arbitrary last-write-wins in the map when a pubkey has multiple rows. The table is written by an external process and its schema/engine exists nowhere in this repo. If it's append-mode (or a pre-merge ReplacingMergeTree), a recovered validator stays flagged; if the writer stops, flags freeze silently. Confirm the write contract (full replace per cycle?) and document it here, or dedup to latest state per pubkey with a recency bound.

web/src/components/publisher-check-page.tsx
```diff
@@ -974,6 +982,11 @@ export function PublisherCheckPage() {
                       <td className="px-4 py-3 text-center">
                         <StatusIcon ok={!pub.publishing_retransmitted} />
                       </td>
+                      <td className="px-4 py-3 text-center">
+                        <StatusIcon
```
ben-dz, line 986: `lagging_status` is never shown: the PR description says the status text appears in the cell and on hover, but `StatusIcon` accepts only `ok` and the cell has no `title`, so the field is shipped to the client and only feeds the string comparison. The legend box above the table also explains the other two icon columns but not "Fast Shreds". Add `title={pub.lagging_status || undefined}` to the cell and a legend entry (and update the PR description) — or drop the field if it has no UI consumer after unifying the predicate.

api/handlers/publisher_check.go
```diff
@@ -325,6 +327,20 @@ func (a *API) FetchPublisherCheckData(ctx context.Context, q string, epochsParam
 		publishers = []PublisherCheckItem{}
 	}
 
+	// Mark publishers whose validator is reported as lagging (status != Healthy)
+	// in dzf_data.lagged_validators. Non-fatal: if the lookup fails we still
+	// return publisher data without lagging annotations.
+	if lagging, err := a.fetchLaggedValidators(ctx); err != nil {
+		slog.Warn("publisher check: lagged validators query failed", "error", err)
```
ben-dz, line 334: The page-cache worker refreshes publisher_check every 30s, so any environment where the table is absent (local dev without remote tables) or the grant is missing logs this warning every 30s indefinitely. Consider skipping the lookup when `!isMainnet(ctx)` and/or probing table availability once at startup.

api/config/config.go
```diff
@@ -83,6 +86,16 @@ func SetDZDPDB(db string) {
 	dzdpDB = db
 }
 
+// GetDZFDataDB returns the dzf_data database name.
+func GetDZFDataDB() string {
+	return dzfDataDB
+}
+
+// SetDZFDataDB sets the dzf_data database name.
+func SetDZFDataDB(db string) {
```
ben-dz, line 95: `SetDZFDataDB` has no callers — the env-var override in `Load()` is the only write path. Drop the setter.

ben-dz (changes requested):
The lagging column ships with an internal contradiction and no verification of its success path. The backend flags any `status != 'Healthy'` (and the sort uses that boolean), but the icon is red only for the exact string `"Action Needed"` — a value that appears nowhere in the backend and isn't documented anywhere — so any other non-Healthy status sorts as lagging while displaying a green check. The test harness never sets `DZFDataDB`, so `fetchLaggedValidators` always errors in tests and every existing test exercises only the warn-and-degrade branch; the read-only prod ClickHouse user has no grant on `dzf_data.lagged_validators`, so the deployed API user's grant should be confirmed before merge or the column ships permanently green. Also worth addressing: the query assumes the externally-written table holds only current state (no recency bound or dedup), and `lagging_status` is never actually shown in the UI despite the PR description saying it appears in the cell and on hover.

ben-dz:
Removing the Preview label. noticed while debugging something else:
> lake-pr-666 has been emitting ~1M log lines per 5 minutes (~3.5k/sec) for at least 4 hours — a hot dzingest: workflow interrupted, reattaching loop on dz_env=testnet.

PR #668, opened 2026-06-26
rollup: source device-interface rollup from gNMI interface_state with per-device fallback to InfluxDB

thijsvanemmerik:
## Summary

Adds a gNMI-based source for the device-interface rollup. The rollup can now compute buckets from the `interface_state` table (gNMI/OpenConfig counters in ClickHouse) and coalesces per device with the existing InfluxDB-fed `fact_dz_device_interface_counters`: for each device, gNMI is used when present, otherwise the InfluxDB path. This lets the rollup move to the gNMI source device-by-device while remaining backward compatible.

## Details

- `ComputeDeviceInterfaceRollupFromGNMI` derives per-sample deltas from `interface_state`'s raw cumulative counters via window functions (per device + interface, ordered by time; values cast to Int64 so counter resets read negative and are dropped), then applies the same aggregation as the InfluxDB path so both sources roll up identically. Only the input sampling cadence differs.
- Enrichment: maps interfaces to links via `dim_dz_links_history` (side A/Z iface names) and parses `Tunnel<N>` for user tunnels.
- Filter: keeps interfaces mapped to a link/tunnel or carrying traffic; drops unmapped idle interfaces.
- `coalesceDeviceInterfaceBuckets` prefers gNMI per device and falls back to the fact table for devices without gNMI data.
- Shared `resolveDeviceInterfaceState` resolves device Status / ISIS / user PK for both sources.
- `RollupDeviceInterfaces` runs both computes and coalesces. The gNMI compute is non-fatal: on error it logs and falls back to fact-only, so the rollup is never worse than the InfluxDB-only path.
- `Activities.TelemetryDatabase` is derived from the network (`telemetry_<env>`); empty disables the gNMI source (fact-only).

## Tests

- gNMI delta computation, link/tunnel enrichment + filtering, per-device coalesce, device-state resolution, and per-network telemetry-DB derivation.
- Cross-source equivalence: identical input through both paths yields matching totals and average rates.
- Existing fact-table rollup tests are unchanged; the InfluxDB path behaves identically.

### DIFF
--- a/indexer/pkg/rollup/activities.go
+++ b/indexer/pkg/rollup/activities.go
@@ -18,6 +18,10 @@ type Activities struct {
 	Log          *slog.Logger
 	IngestionLog *ingestionlog.Writer
 	Network      string
+	// TelemetryDatabase is the gNMI interface_state DB (e.g. telemetry_mainnet_beta).
+	// When set, the device-interface rollup prefers it per device and falls back to the
+	// InfluxDB fact table for devices not present there. Empty = fact-only (current behavior).
+	TelemetryDatabase string
 }
 
 // tableRef returns a qualified table reference. If sourceDB is non-empty, the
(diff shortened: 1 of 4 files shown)


indexer/pkg/rollup/activities.go
```diff
@@ -593,7 +597,227 @@ func (a *Activities) ComputeDeviceInterfaceRollup(ctx context.Context, input Bac
...
+// ComputeDeviceInterfaceRollupFromGNMI computes the device-interface rollup from the
+// gNMI interface_state table (raw cumulative counters) rather than the InfluxDB-derived
+// fact table. interface_state has no deltas, so we derive per-sample deltas with window
+// functions (per device+interface ordered by timestamp), cast to Int64 so counter resets
+// read negative and are dropped, and produce the SAME Nullable(Int64)/Nullable(Float64)
+// delta columns the fact table carries. The outer aggregation is then identical to
+// ComputeDeviceInterfaceRollup, so both sources roll up the same way; only the input
+// sampling cadence differs. Link/tunnel enrichment and the de-junk filter are added in a
+// later step; this step emits empty link context.
+func (a *Activities) ComputeDeviceInterfaceRollupFromGNMI(ctx context.Context, input BackfillChunkInput) ([]DeviceInterfaceBucket, error) {
```
vihu, line 618: `ComputeDeviceInterfaceRollupFromGNMI` computes deltas from raw cumulative counters using only rows inside `[WindowStart, WindowEnd)`. The first in-window sample for each `(device_pubkey, interface_name)` gets `NULL` deltas because the pre-window sample is excluded from the window-function input:

```sql
if(rn > 1, toInt64(in_octets) - toInt64(prev_in_octets), NULL) as in_octets_delta,
...
FROM <telemetry_db>.interface_state
WHERE timestamp >= $1 AND timestamp < $2
```

This is not equivalent to the fact-table path, where rows already contain deltas. The live rollup rewrites a 30-minute window every 30 seconds; as a bucket ages out, its final rewrite occurs when that bucket is at `WindowStart`, so the persisted bucket can undercount by one sample interval. With 30s samples and 5m buckets, steady traffic can be undercounted by roughly 10% for the final bucket value. Backfills have the same class of error at chunk boundaries.

indexer/pkg/rollup/activities.go
```diff
@@ -593,7 +597,227 @@ func (a *Activities) ComputeDeviceInterfaceRollup(ctx context.Context, input Bac
...
+	a.logRollup("computed device interface rollup buckets from gnmi", len(buckets), input, time.Since(start))
+	return buckets, nil
+}
+
+// coalesceDeviceInterfaceBuckets merges gNMI-sourced and fact-table-sourced rollup buckets
+// per device: for any device present in the gNMI buckets, its gNMI buckets win and the
+// fact-table buckets for that device are dropped; devices not present in the gNMI buckets
+// fall back to their fact-table buckets. This lets monitoring shift to gNMI device-by-device
+// as contributors upgrade, with no flag day.
+func coalesceDeviceInterfaceBuckets(gnmiBuckets, factBuckets []DeviceInterfaceBucket) []DeviceInterfaceBucket {
```
vihu, line 810: `coalesceDeviceInterfaceBuckets` treats a device as gNMI-covered if **any** gNMI bucket exists for that device anywhere in the activity window, then drops **all** fact-table buckets for that device:

```go
for _, b := range gnmiBuckets {
    gnmiDevices[b.DevicePK] = struct{}{}
}
...
if _, ok := gnmiDevices[b.DevicePK]; !ok {
    out = append(out, b)
}
```

That is too coarse for `device_interface_rollup_5m`, whose replacement key is `(bucket_ts, device_pk, intf)`. Bad cases include:

- gNMI starts mid-window: recent buckets have gNMI, older buckets only have facts, but older fact buckets are dropped.
- gNMI stalls: one stale gNMI bucket in the 30-minute window suppresses newer fact buckets for that device.
- gNMI emits/keeps only one interface: fact rows for other interfaces on the same device are dropped.
- Because missing gNMI rows do not delete existing `ReplacingMergeTree` rows, partial source selection can leave stale mixed-source rows visible.

indexer/pkg/rollup/activities.go
thijsvanemmerik: Addressed. The gNMI delta query now scans a 10-minute baseline before WindowStart (separate query bound) so the first in-window sample of each (device, interface) diffs against a real prior, and the outer query filters to buckets in [WindowStart, WindowEnd) so lookback rows are not emitted — mirroring the InfluxDB QueryBaselineCounter pattern. Added a test using the live geometry (WindowStart on the bucket boundary, baseline at -60s) asserting the first in-window delta is included; gaps longer than the lookback degrade to the prior one-sample undercount, not worse.

indexer/pkg/rollup/activities.go
thijsvanemmerik: Addressed. coalesceDeviceInterfaceBuckets now merges per (device, intf, bucket_ts) — the table replacement key — instead of per device: gNMI wins per key, fact fills every key gNMI did not produce. This keeps fact coverage for mid-window gNMI start, stalls, and partial-interface coverage, and since every live key is rewritten each cycle it supersedes stale rows during the dual-source overlap. (Purging keys after a device fully retires its InfluxDB source is a follow-up — a bucket_ts TTL.) Two related changes: de-duped link_map to one row per (device, intf) to stop a JOIN fan-out that double-counted the SUM counters, and dropped the gNMI-only de-junk HAVING so the gNMI path matches the InfluxDB path exactly. Tests added for the per-key merge, fan-out, and baseline.

vihu (approved):

vihu:
Just needs a rebase, then feel free to merge @thijsvanemmerik

PR #669, opened 2026-06-29
Tiredsolid scoreboard (internal)

armcconnell:
Internal-only scoreboard for a new market-data venue (sibling to the Edge scoreboard). Gated to allowed-domain Google users.

### DIFF
--- a/api/handlers/hyperliquid_scoreboard.go
+++ b/api/handlers/hyperliquid_scoreboard.go
@@ -1,0 +1,197 @@
+package handlers
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strings"
+	"time"
+)
+
+// hyperliquidCompetitors lists the non-DoubleZero feeds shown on the scoreboard,
+// in display order, mapping each raw feed name to its label.
+var hyperliquidCompetitors = []struct{ Feed, Label string }{
+	{"hyperliquid_public_bbo", "Public API"},
+	{"hydromancer_bbo", "Hydromancer"},
+	{"hyperpc_shared_bbo", "HyperPC"},
+	{"quicknode_l2book_bbo", "QuickNode"},
+}
+
+// hyperliquidWindows maps window params to ClickHouse interval expressions.
+var hyperliquidWindows = map[string]string{
+	"1h":  "1 HOUR",
+	"24h": "24 HOUR",
+	"7d":  "7 DAY",
+}
+
+var hyperliquidSymbolRe = regexp.MustCompile(`^[A-Za-z0-9:_.-]{1,32}$`)
+
+// sanitizeHyperliquidSymbol returns the symbol if safe to inline, else "".
+func sanitizeHyperliquidSymbol(s string) string {
+	s = strings.TrimSpace(s)
+	if s == "" || strings.EqualFold(s, "all") || !hyperliquidSymbolRe.MatchString(s) {
+		return ""
+	}
+	return s
+}
+
+// HyperliquidCompetitor is DoubleZero's head-to-head record vs one competitor feed.
+type HyperliquidCompetitor struct {
+	Feed      string  `json:"feed"`
+	Label     string  `json:"label"`
+	DZWinPct  float64 `json:"dz_win_pct"`
+	LeadP50Ms float64 `json:"lead_p50_ms"`
+	LeadP95Ms float64 `json:"lead_p95_ms"`
+	Races     uint64  `json:"races"`
+}
+
+// HyperliquidNode is the per-vantage breakdown.
+type HyperliquidNode struct {
+	MeasurementNodeID string                  `json:"measurement_node_id"`
+	LocationCode      string                  `json:"location_code"`
+	DZWinSharePct     float64                 `json:"dz_win_share_pct"`
+	TotalRaces        uint64                  `json:"total_races"`
+	Competitors       []HyperliquidCompetitor `json:"competitors"`
+}
+
+// HyperliquidRace is one recent race for the live strip.
+type HyperliquidRace struct {
+	EventTs       time.Time `json:"event_ts"`
+	Symbol        string    `json:"symbol"`
+	LocationCode  string    `json:"location_code"`
+	WinnerFeed    string    `json:"winner_feed"`
+	IsDZ          bool      `json:"is_dz"`
+	RunnerUpFeed  string    `json:"runner_up_feed"`
+	RunnerUpLabel string    `json:"runner_up_label"`
+	LeadMs        float64   `json:"lead_ms"`
+}
+
+// HyperliquidScoreboardResponse is the API response.
+type HyperliquidScoreboardResponse struct {
+	Window        string                  `json:"window"`
+	Symbol        string                  `json:"symbol,omitempty"`
+	GeneratedAt   time.Time               `json:"generated_at"`
+	FeedType      string                  `json:"feed_type"`
+	DZWinSharePct float64                 `json:"dz_win_share_pct"`
+	TotalRaces    uint64                  `json:"total_races"`
+	Competitors   []HyperliquidCompetitor `json:"competitors"`
+	Nodes         []HyperliquidNode       `json:"nodes"`
+	RecentRaces   []HyperliquidRace       `json:"recent_races"`
+}
+
+// labelForFeed maps a raw competitor feed to its display label (falls back to the raw name).
+func labelForFeed(feed string) string {
+	for _, c := range hyperliquidCompetitors {
+		if c.Feed == feed {
+			return c.Label
+		}
+	}
+	return feed
+}
+
+// FetchHyperliquidScoreboardData computes the aggregated scoreboard for a window and
+// optional symbol. Empty symbol means all symbols.
+func (a *API) FetchHyperliquidScoreboardData(ctx context.Context, window, symbol string) (*HyperliquidScoreboardResponse, error) {
+	interval, ok := hyperliquidWindows[window]
+	if !ok {
+		window = "24h"
+		interval = hyperliquidWindows[window]
+	}
+	symbol = sanitizeHyperliquidSymbol(symbol)
+	symbolFilter := ""
+	if symbol != "" {
+		symbolFilter = fmt.Sprintf("AND symbol = '%s'", symbol)
+	}
+	db := fmt.Sprintf("`%s`", a.FeedsDB)
+
+	resp := &HyperliquidScoreboardResponse{
+		Window:      window,
+		Symbol:      symbol,
+		GeneratedAt: time.Now().UTC(),
+		FeedType:    "bbo",
+		Competitors: []HyperliquidCompetitor{},
+		Nodes:       []HyperliquidNode{},
+		RecentRaces: []HyperliquidRace{},
+	}
+
+	// Headline: DZ win share over all DZ-vs-competitor pairwise comparisons.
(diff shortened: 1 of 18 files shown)


api/handlers/hyperliquid_scoreboard.go
```diff
@@ -0,0 +1,610 @@
...
+	// includes measurement_node_id, so races partition cleanly by node — the ROLLUP counts are
+	// exact and the headline is derivable by summing per-node cells. One query (vs two) also
+	// refreshes atomically: on the memory-constrained proxy ClickHouse the refresh succeeds or
+	// retries as a unit, instead of needing two independent scans to both survive the
+	// OvercommitTracker when other page-cache queries have the server near its memory ceiling.
+	q := fmt.Sprintf(`
+		SELECT competitor, measurement_node_id, any(location_code) AS location_code,
+			uniqCombinedIf(rk, dz_won = 1) AS dz_wins,
+			uniqCombinedIf(rk, dz_won = 0) AS dz_losses,
+			toFloat64(quantileTDigestIf(0.5)(lead_ms, dz_won = 1)) AS lead_p50,
```
nikw9944: 🔴 **High — NaN quantile breaks the entire scoreboard response**

These percentiles are computed only over DZ-won rows (`dz_won = 1`). For any `(competitor, node)` cell — or the per-competitor rollup — where DZ won **zero** races in the window, the `If` predicate matches no rows and ClickHouse's `quantileTDigest` over an empty set returns **NaN**. That NaN is scanned into `LeadP50Ms`/`LeadP95Ms` (float64), and `writeJSON` → `json.NewEncoder().Encode()` **errors on NaN** (`json: unsupported value: NaN`), producing a broken/empty 200 body.

**Failure scenario:** at a single vantage node far from Tokyo (or in a thin 1h window) DZ loses all its head-to-head races vs one competitor → that one cell is NaN → the *entire* scoreboard fails to encode for *all* users. It also breaks the page-cache worker: `FetchHyperliquidScoreboardData` → `WritePageCache` marshaling fails the same way, so the cache never populates and every request falls through to the (also-failing) live query. No test covers the 0-DZ-wins case.

**Fix:** coalesce NaN in SQL (e.g. wrap in `ifNull`/replace `isNaN`→0) or sanitize the float64s in Go before encoding.

api/handlers/hyperliquid_scoreboard.go
```diff
@@ -0,0 +1,610 @@
...
+		FROM (
+			SELECT measurement_node_id, location_code,
+				%[1]s AS rk,
+				if(startsWith(feed,'tob_'), loser_feed, feed) AS competitor,
+				if(startsWith(feed,'tob_'), 1, 0) AS dz_won,
+				lead_time_p50_ms AS lead_ms
+			FROM %[2]s.hyperliquid_bbo_feed_race_summary
+			WHERE feed != loser_feed AND loser_feed != ''
+			  AND (startsWith(feed,'tob_') != startsWith(loser_feed,'tob_')) %[3]s
+			  AND event_ts >= now() - INTERVAL %[4]s
```
nikw9944, line 267: 🟠 **Medium — 24h/7d (and symbol-filtered) requests do full multi-day scans on the request path**

This filters on `event_ts >= now() - INTERVAL <window>` and `symbol IN (…)`, but the table is `ORDER BY (measurement_node_id, symbol, …)` with `PARTITION BY toDate(event_ts)`. With no `measurement_node_id` predicate, **neither filter uses the primary index** — the time filter only prunes at day granularity and `symbol` (2nd sort column) can't skip granules without the leading key.

`hyperliquidScoreboardCacheKey` returns `""` for any window other than `1h` and for any `symbol` request, so those bypass the page cache and run live under the 60s request timeout. A 7d window therefore scans ~7–8 full day-partitions (order of ~10⁹ matchup rows at the ~7.5M rows/hour noted in the code) computing `uniqCombined`/`tdigest` per group, on the memory-constrained proxy — likely very slow or a timeout per click.

Consider caching 24h/7d too (with a longer refresh interval), or bounding/rejecting them.

api/handlers/hyperliquid_scoreboard.go
```diff
@@ -0,0 +1,610 @@
...
+		r.WinnerLabel = hyperliquidFeedDisplay(r.WinnerFeed)
+		r.RunnerUpLabel = hyperliquidFeedDisplay(r.RunnerUpFeed)
+		out = append(out, r)
+	}
+	return out, rows.Err()
+}
+
+// hyperliquidScoreboardCacheKey returns the cache key for the default request shape, or "".
+func hyperliquidScoreboardCacheKey(r *http.Request) string {
+	if r.URL.Query().Get("since_ts") != "" || r.URL.Query().Get("symbol") != "" {
```
nikw9944: 🟢 **Low — `since_ts` param bypasses cache but is never applied**

This skips the cache when `since_ts` is present (implying support), but `GetHyperliquidScoreboard` never parses `since_ts` and `FetchHyperliquidScoreboardData` hardcodes `fetchHyperliquidRecentRaces(ctx, time.Time{}, 10)` (line 370). A `?since_ts=…` request therefore pays for an uncached live query yet still returns the default last-15-minutes races. Either wire it through or drop the `since_ts` check here. (Frontend doesn't send it today, so impact is latent.)

api/handlers/hyperliquid_scoreboard.go
```diff
@@ -0,0 +1,610 @@
...
+	defer cancel()
+
+	window := strings.TrimSpace(r.URL.Query().Get("window"))
+	if _, ok := hyperliquidWindows[window]; !ok {
+		window = "1h"
+	}
+	symbol := r.URL.Query().Get("symbol")
+
+	// Degrade gracefully if the proxy table isn't available (e.g. local dev).
+	if !a.hyperliquidFeedsTableExists(ctx) {
```
nikw9944: 🟢 **Low — redundant `EXISTS TABLE` round-trip per uncached request**

On the cache-MISS path this calls `hyperliquidFeedsTableExists`, then immediately calls `FetchHyperliquidScoreboardData`, which calls it *again* (line 208) — two `EXISTS TABLE` queries to ClickHouse per request. Cheap individually, but a pointless extra round-trip. Drop this handler-level check (Fetch already degrades gracefully) or pass the result through.

api/handlers/hyperliquid_scoreboard.go
```diff
@@ -0,0 +1,610 @@
...
+		globalRaces += races
+	}
+
+	// Headline derived from the per-node totals (no extra scan).
+	resp.TotalRaces = globalRaces
+	if globalRaces > 0 {
+		resp.DZWinSharePct = 100.0 * float64(globalWins) / float64(globalRaces)
+	}
+
+	recent, err := a.fetchHyperliquidRecentRaces(ctx, time.Time{}, 10)
```
nikw9944: 🟢 **Low — three independent scans run sequentially on the request path**

The main aggregation, `fetchHyperliquidRecentRaces` (15-min scan), and `fetchHyperliquidPrices` (1-min observations scan) run sequentially. They're independent and could run concurrently (`errgroup`). Harmless for the cached 1h default (runs on the worker), but on the uncached 24h/7d live path their latencies stack under the same 60s budget.

web/src/components/sidebar.tsx
```diff
@@ -116,7 +118,9 @@ const { resolvedTheme, setTheme } = useTheme()
     location.pathname === '/dz/shreds/rewards' ||
     location.pathname.startsWith('/dz/shreds/rewards/')
   const isShredsRoute = location.pathname.startsWith('/dz/shreds') || isShredsPublishersRoute
-  const isEdgeRoute = isShredsRoute
+  const isHyperliquidScoreboardRoute = location.pathname === '/dz/hyperliquid/scoreboard'
+  const isHyperliquidRoute = location.pathname.startsWith('/dz/hyperliquid')
+  const isEdgeRoute = isShredsRoute || isHyperliquidRoute
```
nikw9944, line 123: 🟢 **Low / cosmetic — sidebar highlight inconsistency for the Hyperliquid route**

`isEdgeRoute = isShredsRoute || isHyperliquidRoute` makes the collapsed "Edge" icon highlight on `/dz/hyperliquid` (line 286), but the expanded Edge parent link (line 559) keys off `isShredsRoute` only, so it won't reflect the hyperliquid route. Purely visual; worth confirming the intended grouping (Hyperliquid also renders as its own expanded section).

nikw9944 (commented):
Automated correctness + ClickHouse performance review. Findings are labeled by severity (🔴 high / 🟠 medium / 🟢 low). Nice feature overall — SQL is safely parameterized and it degrades gracefully when the proxy table is absent.

nikw9944 (changes requested):

armcconnell:
Thanks for the thorough review — all findings addressed in `2792d91`, with two small follow-ups after. Mapping each:

**🔴 NaN percentile → broken response** — Fixed. Both lead-time percentiles are now wrapped in `ifNotFinite(…, 0)`, so a `(competitor, node)` cell where DoubleZero won zero races serializes as `0` instead of `NaN`. Added a regression test (`TestHyperliquidScoreboard_ZeroDZWins_Encodable`) that reproduces the exact `json: unsupported value: NaN` failure and now passes.

**🟠 24h/7d full scans on the request path** — Fixed. The 24h/7d windows are now precomputed on the slow background refresher (`StartHyperliquidBackgroundRefresher`) and served from the page cache under per-window keys (`hyperliquidScoreboardWindowKey`), so the request path no longer runs a multi-day scan per click. 1h stays on the page-cache worker; a cold-cache request falls back to the same bounded live path as before.

**🟢 `since_ts` bypasses cache but never applied** — Fixed. Dropped the dead `since_ts` branch from `hyperliquidScoreboardCacheKey` (it was never parsed).

**🟢 Redundant `EXISTS TABLE` per request** — Fixed. Removed the handler-level existence check; `FetchHyperliquidScoreboardData` already degrades to an empty-but-valid response, so it's a single round-trip now (still covered by `TestGetHyperliquidScoreboard_MissingTable`).

**🟢 Three sequential scans** — Fixed. The main ROLLUP scan, recent-races scan, and price lookup now run concurrently via `errgroup` (verified race-clean).

**🟢 Sidebar highlight** — No change; I think this is a false positive. Line 559 is the **Shreds** item, not an "Edge parent" — `Edge` is a section-header `<span>`, and Shreds/`<the new venue>` are siblings that each highlight on their own route. Keying the Shreds link off the other route would incorrectly highlight Shreds while viewing that page; the collapsed group icon intentionally highlights for either. Happy to adjust if I've misread the intended grouping.

Two follow-ups also in the branch since the review: added Dwellir as a competitor, and reordered the per-vantage table (TYO/CHI/NYC) with datacenter facility names + hover detail.

Verified: `go build`, `go vet`, `gofmt`, all scoreboard unit tests pass under `-race`; `tsc` clean.

web/src/components/sidebar.tsx
nikw9944: 🟢 Follow-up (still open): the other five findings were addressed in `2131925`, but this one is unchanged — `sidebar.tsx` was not touched. `isEdgeRoute = isShredsRoute || isHyperliquidRoute` still highlights the collapsed "Edge" icon on `/dz/hyperliquid` while the expanded Edge parent link keys off `isShredsRoute` only, so the two states disagree. Lowest-priority / purely cosmetic — fine to leave as-is if the grouping is intentional, just flagging that it remains.

nikw9944 (approved):

PR #677, opened 2026-06-30
api: bound page-cache refresh concurrency

nikw9944:
## Summary

- Caps page-cache refresh concurrency at 8 (`errgroup.SetLimit`) instead of refreshing all ~28 entries fully in parallel.

Fully-parallel refresh oversubscribed staging's ClickHouse: ~55 concurrent queries pegged the 8-core node (load ~35, iowait 0), so even trivial queries hit their timeouts and the per-entry retry amplified the storm. Bounding concurrency keeps in-flight queries near what ClickHouse can actually run and breaks the retry feedback loop.

## Testing Verification

- Staging `system` tables showed ~55 concurrent queries / load ~35 on 8 cores with queries across ~15 types failing at their timeout ceilings; the cap targets that concurrency directly.
- Pairs with the staging ClickHouse CPU bump (infra #1794); capacity alone left ~3.4 queries/core.

### DIFF
--- a/api/worker/workflow.go
+++ b/api/worker/workflow.go
@@ -152,16 +152,19 @@ type refreshError struct{ msg string }
 
 func (e *refreshError) Error() string { return e.msg }
 
+// pageCacheRefreshConcurrency bounds how many cache entries refresh at once.
+// 2-wide was too slow (each entry refreshed only every few minutes), but fully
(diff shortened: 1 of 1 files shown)


thijsvanemmerik (approved):
LGTM

ben-dz (approved):

PR #678, opened 2026-06-30
api: fix shreds rewards detail header join (cannot determine join keys)

nikw9944:
## Summary

- Fixes `GetShredsRewardsDetail`'s header query, which failed on every request with ClickHouse error 403 `INVALID_JOIN_ON_EXPRESSION` ("Cannot determine join keys").

The query anchored on `FROM (SELECT 1 AS x) one` and then `LEFT JOIN solana_vote_accounts_current v ON v.node_pubkey = ? AND v.epoch_vote_account = 'true'`. The `ON` clause referenced only `v` plus a bound parameter, so there was no key relating the anchor row to `v` — ClickHouse rejects such a join. The fix makes the anchor carry the node_id (`FROM (SELECT ? AS node_id) one`) and joins on it (`ON v.node_pubkey = one.node_id`, `ON g.pubkey = one.node_id`), preserving the "always return one row, zero-valued for unknown node_ids" behavior the code intended.

## Testing Verification

- Reproduced from prod logs: `shreds rewards detail header scan failed error="code: 403 ... Cannot determine join keys in LEFT JOIN ... ON (v.node_pubkey = '<pubkey>') AND (v.epoch_vote_account = 'true')"`, recurring for every validator viewed.
- Validated the corrected query against prod ClickHouse with a real node pubkey from the logs: returns one row (vote_pubkey, validator name, stake, DZ IP) with no join-key error.
- Confirmed the sibling per-epoch query in the same handler is unaffected (it uses a `WHERE node_id = ?` filter, not a join on the param).

### DIFF
--- a/api/handlers/shreds_rewards.go
+++ b/api/handlers/shreds_rewards.go
@@ -609,26 +609,29 @@ func (a *API) GetShredsRewardsDetail(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Header: vote pubkey, validator name, activated stake, DZ user IP.
-	// Same join shape as the list endpoint. A single-row anchor SELECT lets
-	// the LEFT JOINs return zero values for unknown node_ids without erroring.
+	// The single-row anchor carries the node_id so the LEFT JOINs have a real
+	// join key (ON v.node_pubkey = one.node_id); this returns zero values for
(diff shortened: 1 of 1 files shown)


ben-dz (approved):
LGTM, probably worth digging in to why this passed tests but failed in prod? Perhaps create an issue for it?

PR #680, opened 2026-07-05
permission-audit: index serviceability permission events and add internal audit page

juan-malbeclabs:
## Summary

- Adds an onchain **audit trail** for serviceability Permission-management instructions (grant / change / suspend / resume / revoke). The program keeps no audit trail — Permission accounts store only current state and the permission processors emit no logs — so this decodes the **instruction data bytes** (variants 97–101 + Borsh args) from transaction history.
- New `permissionevents` indexer source that **watches each Permission account** (not the whole program) and writes an append-only `fact_dz_permission_events` table capturing the acting admin (tx fee-payer), grantee, flag deltas (role names + raw mask), event type, and success.
- New internal-only **Permission Audit** page (under the DoubleZero sidebar section), gated to `@malbeclabs.com` / allowed-domain users via `RequireInternalDomain`.

## Why

After the serviceability permission migration, there was no way to answer *who granted / changed / revoked which permission flag, to whom, and when* — the data only exists in raw transaction history. This surfaces it as a queryable, viewable audit trail without any onchain/program changes.

## Details

**Indexer** (`indexer/pkg/dz/serviceability/permissionevents/`)

Permission changes are sporadic among a high volume of other serviceability transactions, so the steady-state refresh **watches Permission accounts directly** rather than scanning every program transaction:

- **Discover** the current Permission PDAs via `getProgramAccounts` filtered on the `account_type` discriminator (one cheap call), then **per PDA** `getSignaturesForAddress` since that account's high-water mark and decode only those transactions. A Permission PDA is only referenced by permission instructions, so this fetches only permission txs — not the whole program's history. New grants are picked up once the PDA is discovered (its create tx references the account).
- **Per-account high-water marks** are derived from the fact table (`max(slot)` per `permission_pk`), mirroring `escrowevents`. A failed account is simply not advanced and re-fetched next refresh, so a transient RPC error never leaves a silent gap.
- **`BackfillRefresh`** is a full-history **program scan** (chunked oldest-first, resumable via a program scan cursor) — the completeness backstop for accounts created and deleted between discovery polls, and for initial history seeding. Triggerable via `admin --dz-env <env> --backfill-permission-events[ --backfill-permission-events-truncate]`.
- Decoding scans both top-level and **CPI (inner) instructions** and merges `Meta.LoadedAddresses` so v0/ALT-resolved and multisig-invoked permission ops are captured. `Signer` is the tx fee-payer (documented; may differ from the acting admin for relayer/multisig-paid txs).
- Instruction decoding is reimplemented in Go (the doublezero Go SDK only decodes account state); the decode contract is documented against the Rust source of truth and covered by fixture tests (u128 decoded without truncation; flag→name order cross-checked against `flags.rs`).
- A retrying RPC client (`rpc.NewWithRetries`) is used so a transient `getTransaction` error doesn't drop an event.
- Refresh cadence: the permission refresh runs every ~5 minutes (sporadic changes don't need per-minute polling).

**API** — `GET /api/dz/permission-audit` (internal-domain gated). Newest-first by slot (so events with a missing block time stay visible), with `instruction_index` as a tiebreak. Grantee is only carried by `CreatePermission`, so it is backfilled for the other event types via a self-join to each PDA's create row.

**Web** — Permission Audit table page (time in UTC, event, admin, grantee, flags added/removed, tx→solscan), sidebar link (end of the DoubleZero section) gated on `is_internal_user`.

## Testing Verification

- Fixture-based unit tests for the instruction decoder cover each variant (97–101), including u128 decode without truncation, fee-payer attribution, and flag→name mapping cross-checked against the Rust `flags.rs` order.
- Verified on the DZ mainnet-beta ledger that the per-account discovery + watch correctly surfaces the single existing Permission account (`foundation`) — its `Create` event is indexed on the first steady-state refresh with no manual backfill.
- `go build`, `go vet`, `golangci-lint` (0 issues), and `go test -race` pass on the changed packages.

## Note for reviewers

The indexer architecture changed materially from the first-reviewed revision. The original design scanned the **whole serviceability program** and `getTransaction`'d every tx to find the rare permission instructions, which hammered the RPC. It now **watches Permission accounts individually** (discover-then-watch), fetching only permission transactions; the full program scan is retained only for the backfill. The earlier review comments about the program-wide scan (cursor gaps, first-run timeout, CPI coverage) are addressed by this model and the inline fixes.

### DIFF
--- a/api/handlers/permission_audit.go
+++ b/api/handlers/permission_audit.go
@@ -3,23 +3,20 @@ package handlers
 import (
 	"context"
 	"net/http"
-	"strconv"
 	"time"
 )
 
-const (
-	permissionAuditDefaultLimit = 200
-	permissionAuditMaxLimit     = 1000
-)
+const permissionAuditDefaultLimit = 200
 
 // PermissionAuditRow is a single serviceability Permission-management event.
 type PermissionAuditRow struct {
 	EventTS            time.Time `json:"eventTs"`
 	TxSignature        string    `json:"txSignature"`
 	Slot               uint64    `json:"slot"`
-	Signer             string    `json:"signer"`       // acting admin (transaction fee-payer)
-	PermissionPK       string    `json:"permissionPk"` // the Permission PDA
-	TargetPubkey       string    `json:"targetPubkey"` // grantee
+	InstructionIndex   uint16    `json:"instructionIndex"` // index within the tx; distinguishes multiple events in one tx
+	Signer             string    `json:"signer"`           // acting admin (transaction fee-payer)
+	PermissionPK       string    `json:"permissionPk"`     // the Permission PDA
+	TargetPubkey       string    `json:"targetPubkey"`     // grantee
 	EventType          string    `json:"eventType"`
 	PermissionsAdded   string    `json:"permissionsAdded"`
 	PermissionsRemoved string    `json:"permissionsRemoved"`
@@ -34,15 +31,7 @@ type PermissionAuditResponse struct {
 // GetPermissionAudit serves the serviceability permission audit trail. Gated to
 // internal-domain users (see RequireInternalDomain in api/main.go).
 func (a *API) GetPermissionAudit(w http.ResponseWriter, r *http.Request) {
-	limit := permissionAuditDefaultLimit
-	if v := r.URL.Query().Get("limit"); v != "" {
-		if n, err := strconv.Atoi(v); err == nil && n > 0 {
-			limit = n
-		}
-	}
-	if limit > permissionAuditMaxLimit {
-		limit = permissionAuditMaxLimit
-	}
+	limit := ParsePagination(r, permissionAuditDefaultLimit).Limit
 
 	resp, err := a.FetchPermissionAuditData(r.Context(), limit)
 	if err != nil {
@@ -62,6 +51,7 @@ func (a *API) FetchPermissionAuditData(ctx context.Context, limit int) (Permissi
 			e.event_ts,
 			e.tx_signature,
 			e.slot,
+			e.instruction_index,
 			e.signer,
 			e.permission_pk,
 			if(e.target_pubkey != '', e.target_pubkey, tgt.target) AS target_pubkey,
@@ -69,14 +59,14 @@ func (a *API) FetchPermissionAuditData(ctx context.Context, limit int) (Permissi
 			e.permissions_added,
 			e.permissions_removed,
 			e.success
-		FROM fact_dz_permission_events FINAL AS e
+		FROM fact_dz_permission_events AS e FINAL
 		LEFT JOIN (
 			SELECT permission_pk, max(target_pubkey) AS target
 			FROM fact_dz_permission_events
 			WHERE target_pubkey != ''
 			GROUP BY permission_pk
 		) AS tgt USING (permission_pk)
-		ORDER BY e.event_ts DESC, e.slot DESC, e.tx_signature DESC
+		ORDER BY e.slot DESC, e.event_ts DESC, e.tx_signature DESC, e.instruction_index DESC
 		LIMIT ?
 	`
 
@@ -94,6 +84,7 @@ func (a *API) FetchPermissionAuditData(ctx context.Context, limit int) (Permissi
 			&row.EventTS,
 			&row.TxSignature,
 			&row.Slot,
+			&row.InstructionIndex,
 			&row.Signer,
 			&row.PermissionPK,
 			&row.TargetPubkey,
--- a/api/handlers/permission_audit.go
+++ b/api/handlers/permission_audit.go
@@ -1,0 +1,114 @@
+package handlers
+
+import (
+	"context"
+	"net/http"
+	"strconv"
+	"time"
+)
+
+const (
+	permissionAuditDefaultLimit = 200
+	permissionAuditMaxLimit     = 1000
+)
+
+// PermissionAuditRow is a single serviceability Permission-management event.
+type PermissionAuditRow struct {
+	EventTS            time.Time `json:"eventTs"`
+	TxSignature        string    `json:"txSignature"`
+	Slot               uint64    `json:"slot"`
+	Signer             string    `json:"signer"`       // acting admin (transaction fee-payer)
+	PermissionPK       string    `json:"permissionPk"` // the Permission PDA
+	TargetPubkey       string    `json:"targetPubkey"` // grantee
+	EventType          string    `json:"eventType"`
+	PermissionsAdded   string    `json:"permissionsAdded"`
+	PermissionsRemoved string    `json:"permissionsRemoved"`
+	Success            bool      `json:"success"`
+}
+
+// PermissionAuditResponse is the payload for the permission audit page.
+type PermissionAuditResponse struct {
+	Events []PermissionAuditRow `json:"events"`
+}
+
+// GetPermissionAudit serves the serviceability permission audit trail. Gated to
+// internal-domain users (see RequireInternalDomain in api/main.go).
+func (a *API) GetPermissionAudit(w http.ResponseWriter, r *http.Request) {
+	limit := permissionAuditDefaultLimit
+	if v := r.URL.Query().Get("limit"); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n > 0 {
+			limit = n
+		}
+	}
+	if limit > permissionAuditMaxLimit {
+		limit = permissionAuditMaxLimit
+	}
+
+	resp, err := a.FetchPermissionAuditData(r.Context(), limit)
+	if err != nil {
+		logError("failed to fetch permission audit", "error", err)
+		http.Error(w, "failed to fetch permission audit", http.StatusInternalServerError)
+		return
+	}
+	writeJSON(w, resp)
+}
+
+// FetchPermissionAuditData returns the most recent permission events, newest first.
+// The grantee (target_pubkey) is only carried by CreatePermission instructions, so it
+// is backfilled for the other event types by joining each permission PDA to its create row.
+func (a *API) FetchPermissionAuditData(ctx context.Context, limit int) (PermissionAuditResponse, error) {
+	const query = `
+		SELECT
+			e.event_ts,
+			e.tx_signature,
+			e.slot,
+			e.signer,
+			e.permission_pk,
+			if(e.target_pubkey != '', e.target_pubkey, tgt.target) AS target_pubkey,
+			e.event_type,
+			e.permissions_added,
+			e.permissions_removed,
+			e.success
+		FROM fact_dz_permission_events FINAL AS e
+		LEFT JOIN (
+			SELECT permission_pk, max(target_pubkey) AS target
+			FROM fact_dz_permission_events
+			WHERE target_pubkey != ''
+			GROUP BY permission_pk
+		) AS tgt USING (permission_pk)
+		ORDER BY e.event_ts DESC, e.slot DESC, e.tx_signature DESC
+		LIMIT ?
+	`
+
+	rows, err := a.safeQueryRows(ctx, query, limit)
+	if err != nil {
+		return PermissionAuditResponse{}, err
+	}
+	defer rows.Close()
+
+	resp := PermissionAuditResponse{Events: []PermissionAuditRow{}}
+	for rows.Next() {
+		var row PermissionAuditRow
+		var success uint8
+		if err := rows.Scan(
+			&row.EventTS,
+			&row.TxSignature,
+			&row.Slot,
+			&row.Signer,
+			&row.PermissionPK,
+			&row.TargetPubkey,
+			&row.EventType,
+			&row.PermissionsAdded,
+			&row.PermissionsRemoved,
+			&success,
+		); err != nil {
+			return PermissionAuditResponse{}, err
+		}
+		row.Success = success == 1
+		resp.Events = append(resp.Events, row)
+	}
+	if err := rows.Err(); err != nil {
+		return PermissionAuditResponse{}, err
+	}
+	return resp, nil
+}
--- a/indexer/pkg/dz/serviceability/permissionevents/rpc.go
+++ b/indexer/pkg/dz/serviceability/permissionevents/rpc.go
@@ -1,0 +1,18 @@
+package permissionevents
+
+import (
+	"context"
+
+	"github.com/gagliardetto/solana-go"
+	"github.com/gagliardetto/solana-go/rpc"
+)
+
+// SolanaRPC abstracts the Solana RPC methods needed for fetching serviceability
+// transaction history. Satisfied by *rpc.Client (gagliardetto solana-go).
+type SolanaRPC interface {
+	GetSignaturesForAddressWithOpts(ctx context.Context, account solana.PublicKey, opts *rpc.GetSignaturesForAddressOpts) ([]*rpc.TransactionSignature, error)
+	GetTransaction(ctx context.Context, txSig solana.Signature, opts *rpc.GetTransactionOpts) (*rpc.GetTransactionResult, error)
+}
+
+// Compile-time check that *rpc.Client satisfies SolanaRPC.
+var _ SolanaRPC = (*rpc.Client)(nil)
--- a/indexer/pkg/dz/serviceability/permissionevents/rpc.go
+++ b/indexer/pkg/dz/serviceability/permissionevents/rpc.go
@@ -1,8 +1,20 @@
 package permissionevents
 
-import "github.com/malbeclabs/lake/indexer/pkg/sol"
+import (
+	"context"
 
-// SolanaRPC abstracts the Solana RPC methods needed for fetching serviceability
-// transaction history. It aliases the shared sol.TxHistoryRPC so the escrowevents
-// and permissionevents indexers stay on one interface. Satisfied by *rpc.Client.
-type SolanaRPC = sol.TxHistoryRPC
+	"github.com/gagliardetto/solana-go"
+	"github.com/gagliardetto/solana-go/rpc"
+	"github.com/malbeclabs/lake/indexer/pkg/sol"
+)
+
+// SolanaRPC is the RPC surface the permission-events indexer needs: the shared tx-history
+// methods (getSignaturesForAddress + getTransaction, via sol.TxHistoryRPC) plus
+// getProgramAccounts for discovering Permission PDAs to watch. Satisfied by *rpc.Client.
+type SolanaRPC interface {
+	sol.TxHistoryRPC
+	GetProgramAccountsWithOpts(ctx context.Context, program solana.PublicKey, opts *rpc.GetProgramAccountsOpts) (rpc.GetProgramAccountsResult, error)
+}
+
+// Compile-time check that *rpc.Client satisfies SolanaRPC.
+var _ SolanaRPC = (*rpc.Client)(nil)
--- a/indexer/pkg/dz/serviceability/permissionevents/rpc.go
+++ b/indexer/pkg/dz/serviceability/permissionevents/rpc.go
@@ -1,18 +1,8 @@
 package permissionevents
 
-import (
-	"context"
-
-	"github.com/gagliardetto/solana-go"
-	"github.com/gagliardetto/solana-go/rpc"
-)
+import "github.com/malbeclabs/lake/indexer/pkg/sol"
 
 // SolanaRPC abstracts the Solana RPC methods needed for fetching serviceability
-// transaction history. Satisfied by *rpc.Client (gagliardetto solana-go).
-type SolanaRPC interface {
-	GetSignaturesForAddressWithOpts(ctx context.Context, account solana.PublicKey, opts *rpc.GetSignaturesForAddressOpts) ([]*rpc.TransactionSignature, error)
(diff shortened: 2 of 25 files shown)


api/handlers/permission_audit.go
```diff
@@ -0,0 +1,114 @@
...
+			e.tx_signature,
+			e.slot,
+			e.signer,
+			e.permission_pk,
+			if(e.target_pubkey != '', e.target_pubkey, tgt.target) AS target_pubkey,
+			e.event_type,
+			e.permissions_added,
+			e.permissions_removed,
+			e.success
+		FROM fact_dz_permission_events FINAL AS e
```
nikw9944: `FROM fact_dz_permission_events FINAL AS e` is a ClickHouse syntax error — `FINAL` must come after the alias. Verified directly: `EXPLAIN AST SELECT ... FROM t FINAL AS e` fails with `code: 62, Syntax error: failed at position ... (AS)`, while `FROM fact_dz_permission_events AS e FINAL` parses fine (and matches the codebase convention, e.g. `device_interface_rollup_5m r FINAL` in device_metrics.go:532). As written every request 500s and the page always shows "Failed to load permission audit". Worth a manual end-to-end check after fixing.

indexer/pkg/dz/serviceability/permissionevents/view.go
```diff
@@ -0,0 +1,376 @@
...
+	)
+	g, gctx := errgroup.WithContext(ctx)
+	g.SetLimit(maxConcurrentFetches)
+	for _, sig := range sigs {
+		g.Go(func() error {
+			events, err := v.decodeTransaction(gctx, sig)
+			if err != nil {
+				v.log.Warn("serviceability/permission-events: failed to decode transaction",
+					"signature", sig.Signature.String(), "error", err)
+				return nil // one bad tx shouldn't fail the whole refresh
```
nikw9944: Transient RPC failures permanently drop audit events: decodeTransaction errors are logged at Warn and swallowed here, yet the refresh still succeeds and advances the cursor to `sigs[0]` (line 221). One 429/timeout on getTransaction → that tx's permission events are never indexed and never revisited — a silent permanent gap in the audit trail, healed only by manual backfill. No retry exists anywhere on this path: the client is plain `solanarpc.New(...)` (indexer/cmd/indexer/main.go:344), unlike sibling clients which use `rpc.NewWithRetries(...)` (lines 336/999), and the Temporal retry policy never fires because the refresh reports success. Suggest `rpc.NewWithRetries` + failing the refresh (skipping the cursor advance) when any fetch/decode errored. That also makes the currently-unreachable `g.Wait()` error branch (lines 209-212 — every goroutine returns nil today) meaningful.

indexer/pkg/dzingest/workflow.go
```diff
@@ -59,6 +60,7 @@ func DZIngestWorkflow(ctx temporalworkflow.Context, iteration int) error {
 		geolocationFuture := temporalworkflow.ExecuteActivity(ctx, (*Activities).RefreshGeolocation)
 		shredsFuture := temporalworkflow.ExecuteActivity(ctx, (*Activities).RefreshShreds)
 		escrowEventsFuture := temporalworkflow.ExecuteActivity(ctx, (*Activities).RefreshShredEscrowEvents)
+		permissionEventsFuture := temporalworkflow.ExecuteActivity(ctx, (*Activities).RefreshPermissionEvents)
```
nikw9944: First-run bootstrap can loop forever: with an empty cursor this activity sweeps the serviceability program's entire signature history — accumulating all signatures in memory (unbounded `allSigs`), fetching every tx at concurrency 10, and only writing the cursor at the very end (view.go:221) — under DZIngestWorkflow's `StartToCloseTimeout: 5 * time.Minute` (line 41). If the sweep exceeds 5 minutes, Temporal cancels and retries from scratch; the cursor never advances, so every 60s iteration re-hammers the RPC indefinitely. Suggest checkpointing the cursor per page (chunked oldest-to-newest scan), or making "run BackfillPermissionEventsWorkflow first" (60-min timeout) an enforced bootstrap step rather than a footnote.

indexer/pkg/dz/serviceability/permissionevents/view.go
```diff
@@ -0,0 +1,376 @@
...
+			"count", len(allEvents), "scanned_signatures", len(sigs))
+	}
+	return result, nil
+}
+
+// BackfillRefresh runs a full re-scan ignoring the stored cursor. Existing events are
+// safely overwritten via ReplacingMergeTree dedup.
+func (v *View) BackfillRefresh(ctx context.Context) (ingestionlog.RefreshResult, error) {
+	orig := v.cfg.SkipHighWaterMarks
+	v.cfg.SkipHighWaterMarks = true
```
nikw9944: Data race + wrong-scope scans: `v.cfg.SkipHighWaterMarks` is written here (and restored in the defer) without holding `refreshMu`, while Refresh reads it under the lock (line 162). DZIngestWorkflow and BackfillPermissionEventsWorkflow share one Activities struct on the same worker, and the ingest workflow runs every 60s — so any backfill overlaps periodic refreshes: (a) unsynchronized bool write/read fails `-race`; (b) a periodic refresh that wins the mutex while the flag is set runs an unintended full-history scan under the 5-minute activity timeout. Pass the flag as a parameter to an internal `refresh(ctx, ignoreCursor bool)` instead of mutating shared config. The same pattern exists in escrowevents/view.go:248-251 — worth fixing both.

indexer/pkg/dz/serviceability/permissionevents/view.go
```diff
@@ -0,0 +1,376 @@
...
+	if sig.BlockTime != nil {
+		blockTime = sig.BlockTime.Time().UTC()
+	}
+	success := uint8(1)
+	if sig.Err != nil {
+		success = 0
+	}
+
+	var rows []PermissionEventRow
+	for idx, ci := range tx.Message.Instructions {
```
nikw9944: Only top-level instructions are scanned: permission instructions executed via CPI (e.g. a multisig/governance program invoking serviceability) appear only in `Meta.InnerInstructions` — exposed by the RPC result but never read — so exactly the admin operations an audit page is most needed for would be silently absent. Related gaps: v0 `Meta.LoadedAddresses` is never merged into `accountKeys`, so a looked-up Permission PDA produces a row with empty `permission_pk` (lines 352-355); and `Signer` = fee-payer (line 320) misattributes the acting admin when a relayer or multisig pays fees. If all permission ops today are top-level fee-payer-signed txs, at minimum document that assumption in the package doc; scanning inner instructions is the robust fix.

indexer/pkg/dz/serviceability/permissionevents/view.go
```diff
@@ -0,0 +1,376 @@
...
+		if !beforeSig.IsZero() {
+			opts.Before = beforeSig
+		}
+
+		page, err := v.cfg.RPC.GetSignaturesForAddressWithOpts(ctx, v.cfg.ProgramID, opts)
+		if err != nil {
+			return nil, fmt.Errorf("get signatures: %w", err)
+		}
+		allSigs = append(allSigs, page...)
+		if len(page) < maxSignaturesPerRequest {
```
nikw9944: Pagination terminates on `len(page) < maxSignaturesPerRequest` but `opts.Limit` is never set (line 275), so this relies on the server's default page size being exactly 1000. A provider/gateway with a lower cap or short mid-history pages ends pagination early, and on first run the cursor then advances past the skipped history permanently. Set `opts.Limit` explicitly and/or terminate only on an empty page. (Same pre-existing pattern in escrowevents/view.go:307.)

web/src/components/permission-audit-page.tsx
```diff
@@ -0,0 +1,164 @@
...
+                <th className="px-4 py-2.5 font-medium">Grantee</th>
+                <th className="px-4 py-2.5 font-medium">Added</th>
+                <th className="px-4 py-2.5 font-medium">Removed</th>
+                <th className="px-4 py-2.5 font-medium">Tx</th>
+              </tr>
+            </thead>
+            <tbody>
+              {events.map((e) => (
+                <tr
+                  key={`${e.txSignature}-${e.permissionPk}-${e.eventType}-${e.slot}`}
```
nikw9944: React row keys can collide: the table's dedup key is `(permission_pk, slot, tx_signature, instruction_index)` and the indexer emits one row per instruction, so two same-type instructions to the same PDA in one tx are distinct rows — but this key omits the instruction index (the API drops it: PermissionAuditRow in permission_audit.go:16-27), making the keys identical. The SQL ORDER BY (permission_audit.go:79) also lacks an `instruction_index` tiebreak, so within-tx ordering is nondeterministic. Return `instructionIndex` in the API and add it to both the ORDER BY and this key.

web/src/components/permission-audit-page.tsx
```diff
@@ -0,0 +1,164 @@
...
+        <p className="text-sm text-muted-foreground mb-8">
+          Onchain history of serviceability permission grants, changes, suspensions and revocations —
+          who did what, to whom, and when.
+        </p>
+
+        <div className="border border-border rounded-lg overflow-x-auto">
+          <table className="w-full">
+            <thead>
+              <tr className="text-left text-sm text-muted-foreground border-b border-border bg-muted/30">
+                <th className="px-4 py-2.5 font-medium">Time (UTC local)</th>
```
nikw9944: "Time (UTC local)" is self-contradictory, and `formatTs` (lines 29-40) calls `toLocaleString` with no `timeZone`, so timestamps render in the browser's local zone — viewers in different zones see different times under a header implying UTC, which matters on an audit page. Either add `timeZone: 'UTC'` and label "Time (UTC)", or label it "Time (local)".

indexer/pkg/dz/serviceability/permissionevents/view.go
```diff
@@ -0,0 +1,376 @@
...
+	}
+	if tx == nil || len(tx.Message.AccountKeys) == 0 {
+		return nil, nil
+	}
+
+	accountKeys := tx.Message.AccountKeys
+	feePayer := accountKeys[0].String()
+
+	var blockTime time.Time
+	if sig.BlockTime != nil {
```
nikw9944, line 557: A nil `sig.BlockTime` (a valid RPC state) leaves EventTS as Go's zero time — outside DateTime64(3)'s supported range, and it sorts last under `ORDER BY event_ts DESC`, so those rows fall past the LIMIT on the newest-first audit page: the anomalous events an auditor most needs to see become invisible. Fall back to a slot-derived timestamp, or make `slot` the primary sort key.

api/handlers/permission_audit.go
```diff
@@ -0,0 +1,114 @@
...
+// GetPermissionAudit serves the serviceability permission audit trail. Gated to
+// internal-domain users (see RequireInternalDomain in api/main.go).
+func (a *API) GetPermissionAudit(w http.ResponseWriter, r *http.Request) {
+	limit := permissionAuditDefaultLimit
+	if v := r.URL.Query().Get("limit"); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n > 0 {
+			limit = n
+		}
+	}
+	if limit > permissionAuditMaxLimit {
```
nikw9944: This hand-rolled default/clamp block duplicates the existing `ParsePagination` helper (api/handlers/pagination.go:27, whose MaxLimit is already 1000): `limit := ParsePagination(r, permissionAuditDefaultLimit).Limit`.

indexer/pkg/dz/serviceability/permissionevents/rpc.go
```diff
@@ -0,0 +1,18 @@
...
+	"context"
+
+	"github.com/gagliardetto/solana-go"
+	"github.com/gagliardetto/solana-go/rpc"
+)
+
+// SolanaRPC abstracts the Solana RPC methods needed for fetching serviceability
+// transaction history. Satisfied by *rpc.Client (gagliardetto solana-go).
+type SolanaRPC interface {
+	GetSignaturesForAddressWithOpts(ctx context.Context, account solana.PublicKey, opts *rpc.GetSignaturesForAddressOpts) ([]*rpc.TransactionSignature, error)
```
nikw9944: Verbatim copy of escrowevents/rpc.go's `SolanaRPC` — now the third such interface in the tree. Consider extracting one shared tx-history interface (e.g. alongside indexer/pkg/sol) and aliasing it from both event packages.

indexer/pkg/dz/serviceability/permissionevents/view.go
```diff
@@ -0,0 +1,376 @@
...
+			PermissionsAdded:       FlagNames(decoded.Added),
+			PermissionsRemoved:     FlagNames(decoded.Removed),
+			PermissionsAddedMask:   decoded.Added.Hex(),
+			PermissionsRemovedMask: decoded.Removed.Hex(),
+			Success:                success,
+		})
+	}
+
+	// Deterministic order within a tx (stable across concurrent refreshes).
+	sort.SliceStable(rows, func(i, j int) bool { return rows[i].InstructionIndex < rows[j].InstructionIndex })
```
nikw9944: This `sort.SliceStable` is a no-op — `rows` is appended in ascending instruction index within a single goroutine, and the comment about concurrent refreshes is misleading (concurrency is per-transaction). Related minor cleanups: store.go:104 re-creates the dataset on every InsertEvents (build once in NewStore), and decode.go:110-113 recomputes the constant `known` mask on every FlagNames call (hoist to a package-level var).

web/src/components/sidebar.tsx
```diff
@@ -600,6 +602,12 @@ const { resolvedTheme, setTheme } = useTheme()
                 )}
               </>
             )}
+            {showPermissionAudit && (
+              <Link to="/dz/permission-audit" className={navItemClass(isPermissionAuditRoute)}>
```
nikw9944: The Permission Audit link renders under the "Edge" section heading; it's a serviceability feature and fits better under the DoubleZero section (or a dedicated internal/ops group).

nikw9944 (changes requested):
Solid decoder with good fixture tests, but the feature is non-functional as written: the audit query uses `FROM fact_dz_permission_events FINAL AS e`, which is a ClickHouse syntax error (verified via EXPLAIN AST) — every request 500s. Beyond that, the indexer's cursor handling undermines the audit-trail guarantee: transient getTransaction failures are swallowed while the cursor still advances (permanent silent gaps), the first-run full-history scan cannot fit the 5-minute Temporal activity timeout (potential infinite retry loop on bootstrap), `BackfillRefresh` mutates shared config outside the mutex (data race with the always-running periodic refresh), and CPI-invoked permission instructions are never scanned. Verification: `go build ./...`, `golangci-lint run` (0 issues), and `go test -race ./indexer/pkg/dz/serviceability/permissionevents/` all pass. Details and suggested fixes in the inline comments.

api/handlers/permission_audit.go
juan-malbeclabs: Fixed in 1896f53 — changed to `FROM fact_dz_permission_events AS e FINAL` (alias before `FINAL`), matching the codebase convention. Added `instruction_index` as an `ORDER BY` tiebreak too. Couldn't run it against ClickHouse in my env, so the manual end-to-end check is still worth doing.

api/handlers/permission_audit.go
juan-malbeclabs: Fixed in 1896f53 — now `limit := ParsePagination(r, permissionAuditDefaultLimit).Limit`. Removed the hand-rolled clamp and the now-unused `permissionAuditMaxLimit` const.

indexer/pkg/dz/serviceability/permissionevents/view.go
juan-malbeclabs: Addressed in 1896f53. Falls back to `txResult.BlockTime` when `sig.BlockTime` is nil, and the API now sorts by `slot DESC` first (event_ts secondary), so a row with a missing block time stays at the top of the newest-first view instead of sinking past the LIMIT.

indexer/pkg/dz/serviceability/permissionevents/rpc.go
juan-malbeclabs: Done in 1896f53. Extracted `sol.TxHistoryRPC` in `indexer/pkg/sol`; both `permissionevents` and `escrowevents` `SolanaRPC` are now type aliases of it.

indexer/pkg/dzingest/workflow.go
juan-malbeclabs: Addressed in 1896f53. The scan is now processed oldest-first in chunks (`scanChunkSize = 200`) and the cursor is checkpointed after each chunk is durably written, so a cancelled/timed-out refresh resumes from the last completed chunk instead of restarting — no more no-progress loop under the 5-minute activity timeout. Caveat: the signature list is still gathered up-front, so a large cold start is still best run via `BackfillPermissionEventsWorkflow` (60-minute timeout, now also resumable).

indexer/pkg/dz/serviceability/permissionevents/view.go
juan-malbeclabs: Fixed in 1896f53. Replaced the `SkipHighWaterMarks` mutation with an internal `refresh(ctx, ignoreCursor bool)`: `Refresh` passes `cfg.SkipHighWaterMarks`, `BackfillRefresh` passes `true`. No shared-config write remains, so no race with periodic refreshes. Applied the same fix to `escrowevents`. Passes `go test -race`.

indexer/pkg/dz/serviceability/permissionevents/view.go
juan-malbeclabs: Fixed in 1896f53. The permission-events client is now `rpc.NewWithRetries(...)` (main.go, both call sites). Decoding moved into `decodeChunk`, which propagates the first fetch/decode error via a plain `errgroup` (no early cancel, so the rest of the chunk's good events still decode); the refresh then inserts what decoded and returns the error **without advancing the cursor**, so the affected signatures are re-scanned instead of silently dropped. The `g.Wait()` error path is now meaningful.

indexer/pkg/dz/serviceability/permissionevents/view.go
juan-malbeclabs: Fixed in 1896f53. Removed the no-op `sort` — rows are emitted in order via a per-instruction `seq` counter. Also hoisted the `known` flag mask to a package-level var, and the dataset is now built once in `NewStore` instead of per `InsertEvents`.

web/src/components/sidebar.tsx
juan-malbeclabs: Moved in 1896f53 — the Permission Audit link is now at the end of the DoubleZero section.

web/src/components/permission-audit-page.tsx
juan-malbeclabs: Fixed in 1896f53. Header is now "Time (UTC)" and `formatTs` renders with `timeZone: 'UTC'`.

indexer/pkg/dz/serviceability/permissionevents/view.go
juan-malbeclabs: Fixed in 1896f53. `opts.Limit` is now set explicitly and pagination terminates only on an empty page (robust to gateways that cap pages below the requested limit). Same fix applied to `escrowevents`.

indexer/pkg/dz/serviceability/permissionevents/view.go
juan-malbeclabs: Fixed in 1896f53. `decodeTransaction` now scans `Meta.InnerInstructions` in addition to top-level instructions, and merges `Meta.LoadedAddresses` (writable, then read-only) into `accountKeys` so v0/ALT-resolved Permission PDAs decode instead of producing an empty `permission_pk`. Documented in the func doc that `Signer` is the fee-payer and may differ from the acting admin for relayer/multisig-paid txs.

web/src/components/permission-audit-page.tsx
juan-malbeclabs: Fixed in 1896f53. `instruction_index` is now returned by the API (`PermissionAuditRow`), added to the `ORDER BY`, and included in the row key: `${txSignature}-${permissionPk}-${slot}-${instructionIndex}`. The indexer emits a stable per-instruction index across top-level and inner instructions.

juan-malbeclabs:
Pushed follow-ups since the review (commits `1896f53`, `c68c26a`, `c0c4196`):

**1. Fixes for all inline comments** (`1896f53`) — replied inline on each.

**2. Architecture change — per-account watching** (`c0c4196`). The steady-state refresh no longer scans the whole serviceability program and `getTransaction`s every tx to find the rare permission instructions (that was hammering the RPC). It now:
- discovers Permission PDAs via `getProgramAccounts` (one cheap call), then
- per PDA, `getSignaturesForAddress` since that account's high-water mark and decodes only those txs.

Per-account high-water marks are derived from the fact table (`max(slot)` per `permission_pk`), like `escrowevents`. `BackfillRefresh` keeps the full-history **program** scan (chunked, resumable) as the completeness backstop for created-then-deleted accounts. Permission refresh cadence dropped to ~5 min.

This resolves the program-wide-scan concerns from the review (first-run timeout, cursor gaps, RPC volume) at the design level; the inline fixes cover the rest.

**3. `admin --backfill-permission-events`** (`c68c26a`) — the backfill workflow had no CLI trigger; added one mirroring `--backfill-escrow-events`.

Verified on mainnet-beta that discovery surfaces the one existing `foundation` permission and its `Create` is indexed on the first steady-state refresh. PR description updated to match the new model.

nikw9944 (approved):

PR #681, opened 2026-07-06
api: log transient CH/RPC errors at WARN so they don't page on-call

nikw9944:
## Summary

Prod `lake-api` was paging `#alerts` for self-healing, non-actionable errors. This makes transient failures log at WARN (which doesn't alert) and reserves ERROR for genuinely actionable ones. Three observed false alarms, all now quieted:

- **Transient ClickHouse connection drops** (`multicast device endpoint health query error`, `validators rows iteration failed`) — a single `failed to read packet from …:9440` (connection reset / i/o timeout) logged ERROR immediately. These request-path handlers now log at WARN when the error is transient (`dberror.IsTransient`), ERROR otherwise.
- **External RPC rate-limiting** (`dz ledger` cache, `rate limited (status 429)`) — a brief Solana-RPC 429 burst reached the page-cache's 3-consecutive-failure ERROR threshold. `dberror` now classifies rate-limits (429) as transient, and the page-cache escalates transient causes at a higher threshold (`transientErrorAfterFailures = 10`, ~5 min) so a short blip stays WARN. A *sustained* failure still escalates to ERROR, so real outages (e.g. ClickHouse actually down) still page.

The principle: transient/self-healing conditions never page; sustained or genuinely-broken conditions still do.

## Testing Verification

- New `dberror` table test asserts the real prod error strings classify correctly: the two CH `read packet …:9440` messages (reset / i/o timeout) and the RPC `rate limited (status 429)` are transient; syntax/auth/context-cancel errors remain non-transient (still ERROR-worthy).
- Confirmed the page-cache escalation now branches on `IsTransient`: transient causes use the 10-failure threshold, everything else the existing 3.

### DIFF
--- a/api/handlers/dberror/dberror.go
+++ b/api/handlers/dberror/dberror.go
@@ -23,6 +23,8 @@ const (
 	ErrorTypeAuth
 	// ErrorTypeQuery indicates a query/syntax error.
 	ErrorTypeQuery
+	// ErrorTypeRateLimit indicates the upstream throttled the request (e.g. HTTP 429).
+	ErrorTypeRateLimit
 )
 
 // IsTransient returns true if the error is likely transient and worth retrying.
@@ -38,7 +40,7 @@ func IsTransient(err error) bool {
 
 	errType := Classify(err)
 	switch errType {
-	case ErrorTypeConnectivity, ErrorTypeTimeout:
+	case ErrorTypeConnectivity, ErrorTypeTimeout, ErrorTypeRateLimit:
 		return true
 	default:
 		return false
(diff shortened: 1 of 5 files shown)


thijsvanemmerik (approved):
Nice cleanup. Confirmed the mechanism end to end: WARN renders as `WRN`, the errors alert only matches `ERR`, so downgrading these lines genuinely stops the page. Sustained failures still page, since the cache worker runs the same queries in-process and escalates to ERROR at the higher threshold. Real outages stay covered.

A few points, none blocking:

1. **Completeness.** The transient-to-WARN handling is added to two handlers (`validators`, multicast delivery). About a dozen other read handlers that query ClickHouse directly (`users`, `links`, `devices`, `metros`, and so on) still log transient errors at ERROR, so the same blip can still page from them. Consider centralizing the `IsTransient` check (a small helper or middleware) so every read handler gets the same treatment, rather than per-handler edits.
2. **Classifier ordering** (`dberror.go:109-121`). The rate-limit patterns run before the auth and query patterns, so a composite like `authentication failed: too many requests` would classify as transient. Low real-world likelihood, but ordering auth/query first, or tightening the 429 match, removes the hazard.
3. **Test coverage** (`dberror_test.go`). The `"status 429"` pattern isn't actually exercised, since the sample string matches `"rate limited"` first. A case with only `status 429` would guard it. A unit test around the worker threshold selection (transient vs non-transient) and the reset-on-success behavior would also be good, since that's the paging logic this PR ships.
4. **Optional consistency.** `validators.go` returns `err.Error()` directly to the client, while the multicast path returns `dberror.UserMessage(err)`. Using `UserMessage` in both keeps responses consistent and avoids returning raw driver detail to callers.

Small note: the worker comment says the transient threshold is about 5 minutes; with the counter it's nearer 4.5, and it varies by refresh interval per cache. Not important.

Happy to approve.

PR #685, opened 2026-07-08
indexer: ingest onchain BGP status into dz_users_current

ben-malbeclabs:
## Summary

Ingest the onchain BGP session status (`bgp_status`) into `dz_users_current` so multicast health can tell "client not connected (BGP down)" apart from a real forwarding fault.

The multicast Health tab flags any activated publisher/subscriber whose BGP session is down as **unhealthy** ("publisher TunnelN not seen as RPF interface"). Investigation of the `edge-solana-shreds` group found **100% of its 92 unhealthy publishers were onchain BGP-down** — the health views only knew onchain membership, not connectivity. `bgp_status` is onchain on the serviceability User account but lake's indexer dropped it. This PR extracts it; the follow-up (#686) consumes it as a distinct `disconnected` state.

- Add `bgp_status` to `dim_dz_users_history` / `stg_dim_dz_users_snapshot` and expose it on `dz_users_current` (migration mirrors the `tenant_pk` pattern).
- Map it from the Go serviceability SDK in `convertUsers`, normalizing to `up` | `down` | `unknown` (out-of-range enum values → `unknown`).
- `DEFAULT 'unknown'` on the new column; consumers treat anything `!= 'down'` as fail-open (pre-migration/unreported rows keep old behavior).

### Deliberately not ingested
The onchain `last_bgp_up_at` / `last_bgp_reported_at` slots are **not** ingested. `last_bgp_reported_at` is rewritten by a ~6-hourly onchain keepalive; since `attrs_hash` covers every payload column, ingesting it would write a new `dim_dz_users_history` row per user ~4×/day forever (contentless timeline events, growing `dz_users_current` scan). `bgp_status` changes only on real transitions, so there is no ongoing churn. If a freshness signal is needed later, it can be added as a non-hashed column.

### Local dev
The `lake.*` remote-proxy tables snapshot the remote schema at creation, so a local dev instance won't see `bgp_status` until the indexer restarts with `--setup-remote-tables` after this migration lands. Self-healing.

## Testing Verification

- `convertUsers` maps `bgp_status` and normalizes an out-of-range enum value to `unknown`.
- Change-detection guard: re-ingesting an unchanged `bgp_status` writes no new history row; a real transition writes exactly one (pins that there is no keepalive churn).
- Full write→`dz_users_current` round-trip through the migration (ClickHouse testcontainer).

### DIFF
--- a/indexer/db/clickhouse/migrations/20260708000001_users_bgp_status.sql
+++ b/indexer/db/clickhouse/migrations/20260708000001_users_bgp_status.sql
@@ -1,25 +1,35 @@
 -- +goose Up
-
--- Add onchain BGP session state (User account: bgp_status, last_bgp_up_at,
--- last_bgp_reported_at) to the users dimension so consumers can distinguish a
--- user that is not connected (BGP down) from a multicast forwarding fault.
+--
+-- Add the onchain BGP session status (User account) to the users dimension so
+-- consumers can distinguish a user that is not connected (BGP down) from a
+-- multicast forwarding fault.
+--
+-- Only bgp_status is ingested. The onchain last_bgp_up_at / last_bgp_reported_at
+-- slots are deliberately NOT ingested: last_bgp_reported_at is rewritten by a
+-- ~6-hourly onchain keepalive, and since attrs_hash covers every payload column
+-- it would churn a new dim_dz_users_history row per user ~4x/day forever. If a
+-- freshness signal is needed later, add it as a non-hashed column.
+--
+-- Column values: 'up' | 'down' | 'unknown' (agent never reported). DEFAULT
+-- 'unknown' means "not yet reported"; consumers treat anything != 'down' as
+-- fail-open (old behavior), so an unreported user is never masked.
+--
+-- NOTE: the AFTER placement is load-bearing. loadSnapshotIntoStaging appends a
+-- positional batch with no column list, so the physical column order MUST match
+-- UserSchema.PayloadColumns() order (bgp_status last, after subscribers).
 
 -- +goose StatementBegin
 ALTER TABLE dim_dz_users_history
-    ADD COLUMN IF NOT EXISTS bgp_status String DEFAULT '' AFTER subscribers,
-    ADD COLUMN IF NOT EXISTS last_bgp_up_at UInt64 DEFAULT 0 AFTER bgp_status,
-    ADD COLUMN IF NOT EXISTS last_bgp_reported_at UInt64 DEFAULT 0 AFTER last_bgp_up_at;
+    ADD COLUMN IF NOT EXISTS bgp_status String DEFAULT 'unknown' AFTER subscribers;
 -- +goose StatementEnd
 
 -- +goose StatementBegin
 ALTER TABLE stg_dim_dz_users_snapshot
-    ADD COLUMN IF NOT EXISTS bgp_status String DEFAULT '' AFTER subscribers,
-    ADD COLUMN IF NOT EXISTS last_bgp_up_at UInt64 DEFAULT 0 AFTER bgp_status,
-    ADD COLUMN IF NOT EXISTS last_bgp_reported_at UInt64 DEFAULT 0 AFTER last_bgp_up_at;
+    ADD COLUMN IF NOT EXISTS bgp_status String DEFAULT 'unknown' AFTER subscribers;
 -- +goose StatementEnd
 
 -- +goose StatementBegin
--- Recreate dz_users_current to expose the BGP columns.
+-- Recreate dz_users_current to expose bgp_status.
 CREATE OR REPLACE VIEW dz_users_current
 AS
 WITH ranked AS (
@@ -45,9 +55,7 @@ SELECT
     tunnel_id,
     publishers,
     subscribers,
-    bgp_status,
-    last_bgp_up_at,
-    last_bgp_reported_at
+    bgp_status
 FROM ranked
 WHERE rn = 1 AND is_deleted = 0;
 -- +goose StatementEnd
@@ -55,7 +63,7 @@ WHERE rn = 1 AND is_deleted = 0;
 -- +goose Down
 
 -- +goose StatementBegin
--- Restore dz_users_current without the BGP columns.
+-- Restore dz_users_current without bgp_status.
 CREATE OR REPLACE VIEW dz_users_current
 AS
 WITH ranked AS (
@@ -87,14 +95,10 @@ WHERE rn = 1 AND is_deleted = 0;
 
 -- +goose StatementBegin
 ALTER TABLE dim_dz_users_history
-    DROP COLUMN IF EXISTS bgp_status,
-    DROP COLUMN IF EXISTS last_bgp_up_at,
-    DROP COLUMN IF EXISTS last_bgp_reported_at;
+    DROP COLUMN IF EXISTS bgp_status;
 -- +goose StatementEnd
 
 -- +goose StatementBegin
 ALTER TABLE stg_dim_dz_users_snapshot
-    DROP COLUMN IF EXISTS bgp_status,
-    DROP COLUMN IF EXISTS last_bgp_up_at,
-    DROP COLUMN IF EXISTS last_bgp_reported_at;
+    DROP COLUMN IF EXISTS bgp_status;
 -- +goose StatementEnd
--- a/indexer/db/clickhouse/migrations/20260708000001_users_bgp_status.sql
+++ b/indexer/db/clickhouse/migrations/20260708000001_users_bgp_status.sql
@@ -1,0 +1,100 @@
+-- +goose Up
+
+-- Add onchain BGP session state (User account: bgp_status, last_bgp_up_at,
+-- last_bgp_reported_at) to the users dimension so consumers can distinguish a
+-- user that is not connected (BGP down) from a multicast forwarding fault.
+
+-- +goose StatementBegin
+ALTER TABLE dim_dz_users_history
+    ADD COLUMN IF NOT EXISTS bgp_status String DEFAULT '' AFTER subscribers,
+    ADD COLUMN IF NOT EXISTS last_bgp_up_at UInt64 DEFAULT 0 AFTER bgp_status,
+    ADD COLUMN IF NOT EXISTS last_bgp_reported_at UInt64 DEFAULT 0 AFTER last_bgp_up_at;
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+ALTER TABLE stg_dim_dz_users_snapshot
+    ADD COLUMN IF NOT EXISTS bgp_status String DEFAULT '' AFTER subscribers,
+    ADD COLUMN IF NOT EXISTS last_bgp_up_at UInt64 DEFAULT 0 AFTER bgp_status,
+    ADD COLUMN IF NOT EXISTS last_bgp_reported_at UInt64 DEFAULT 0 AFTER last_bgp_up_at;
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+-- Recreate dz_users_current to expose the BGP columns.
+CREATE OR REPLACE VIEW dz_users_current
+AS
+WITH ranked AS (
+    SELECT
+        *,
+        row_number() OVER (PARTITION BY entity_id ORDER BY snapshot_ts DESC, ingested_at DESC, op_id DESC) AS rn
+    FROM dim_dz_users_history
+)
+SELECT
+    entity_id,
+    snapshot_ts,
+    ingested_at,
+    op_id,
+    attrs_hash,
+    pk,
+    owner_pubkey,
+    status,
+    kind,
+    client_ip,
+    dz_ip,
+    device_pk,
+    tenant_pk,
+    tunnel_id,
+    publishers,
+    subscribers,
+    bgp_status,
+    last_bgp_up_at,
+    last_bgp_reported_at
+FROM ranked
+WHERE rn = 1 AND is_deleted = 0;
+-- +goose StatementEnd
+
+-- +goose Down
+
+-- +goose StatementBegin
+-- Restore dz_users_current without the BGP columns.
+CREATE OR REPLACE VIEW dz_users_current
+AS
+WITH ranked AS (
+    SELECT
+        *,
+        row_number() OVER (PARTITION BY entity_id ORDER BY snapshot_ts DESC, ingested_at DESC, op_id DESC) AS rn
+    FROM dim_dz_users_history
+)
+SELECT
+    entity_id,
+    snapshot_ts,
+    ingested_at,
+    op_id,
+    attrs_hash,
+    pk,
+    owner_pubkey,
+    status,
+    kind,
+    client_ip,
+    dz_ip,
+    device_pk,
+    tenant_pk,
+    tunnel_id,
+    publishers,
+    subscribers
+FROM ranked
+WHERE rn = 1 AND is_deleted = 0;
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+ALTER TABLE dim_dz_users_history
+    DROP COLUMN IF EXISTS bgp_status,
+    DROP COLUMN IF EXISTS last_bgp_up_at,
+    DROP COLUMN IF EXISTS last_bgp_reported_at;
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+ALTER TABLE stg_dim_dz_users_snapshot
+    DROP COLUMN IF EXISTS bgp_status,
+    DROP COLUMN IF EXISTS last_bgp_up_at,
+    DROP COLUMN IF EXISTS last_bgp_reported_at;
+-- +goose StatementEnd
--- a/indexer/pkg/dz/serviceability/schema.go
+++ b/indexer/pkg/dz/serviceability/schema.go
@@ -164,8 +164,6 @@ func (s *UserSchema) PayloadColumns() []string {
 		"publishers:VARCHAR",
 		"subscribers:VARCHAR",
 		"bgp_status:VARCHAR",
-		"last_bgp_up_at:BIGINT",
-		"last_bgp_reported_at:BIGINT",
 	}
 }
 
@@ -185,8 +183,6 @@ func (s *UserSchema) ToRow(u User) []any {
 		string(publishersJSON),
 		string(subscribersJSON),
 		u.BgpStatus,
-		u.LastBgpUpAt,
-		u.LastBgpReportedAt,
 	}
 }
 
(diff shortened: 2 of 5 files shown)


indexer/pkg/dz/serviceability/schema.go
```diff
@@ -163,6 +163,9 @@ func (s *UserSchema) PayloadColumns() []string {
 		"tunnel_id:INTEGER",
 		"publishers:VARCHAR",
 		"subscribers:VARCHAR",
+		"bgp_status:VARCHAR",
+		"last_bgp_up_at:BIGINT",
+		"last_bgp_reported_at:BIGINT",
```
juan-malbeclabs: Nit: these columns are declared `BIGINT` here but `UInt64` in the migration. Harmless in practice — the type half of `name:type` is discarded by `extractColumnName` (only the column name is used for INSERT), and the write uses the Go `uint64` value against the `UInt64` column, which the round-trip test confirms. But `BIGINT` conventionally implies signed `Int64`, so the label is slightly misleading vs. the actual unsigned column. Purely cosmetic.

indexer/pkg/dz/serviceability/view.go
```diff
@@ -137,6 +137,10 @@ type User struct {
 	TunnelID    uint16
 	Publishers  []string // multicast group PKs this user publishes to
 	Subscribers []string // multicast group PKs this user subscribes to
+	// BGP session state as last reported by the device agent (onchain).
+	BgpStatus         string // "up" | "down" | "unknown"
+	LastBgpUpAt       uint64 // ledger slot of the most recent BGP up event
+	LastBgpReportedAt uint64 // ledger slot of the most recent BGP status report
```
juan-malbeclabs: Nit: worth double-checking these are ledger *slots* and not unix timestamps. The follow-up health view will likely compute BGP staleness from `last_bgp_reported_at`, so this comment is effectively the contract it will rely on — getting the unit wrong there would be a real bug.

juan-malbeclabs (approved):
LGTM. Purely-additive BGP status ingest; traced the full path (SDK → convertUsers → schema → migration → view) and found no correctness issues. Migration mirrors the tenant_pk pattern, additive columns are safe for the by-name consumers, and both unit + testcontainer round-trip tests cover it. Only left two cosmetic nits inline.

indexer/pkg/dz/serviceability/schema.go
ben-dz: `attrs_hash` is computed over ALL payload columns (dim.go:85-103) and drives the SCD delta (dim_write.go:226). The onchain program sets `last_bgp_reported_at = slot` on every SetUserBGPStatus, and the device agent submits on status change plus a ~6h periodic refresh per user — so every user gets a new `dim_dz_users_history` row at least ~4×/day forever (table has no TTL), and each row emits an `entity_changes_v` 'updated' event with empty `changed_fields` (BGP fields aren't in its lag list), rendered by the timeline as "User X updated (0 fields)" inside a LIMIT 500 window. Plus a one-time per-user flood at deploy (hash tuple changed). Options: (a) support non-hashed payload columns so the slot fields are carried but excluded from `attrs_hash` (keep `bgp_status` hashed); (b) put the slot fields in a separate narrow table; (c) if churn is accepted, fix `entity_changes_v` in this PR to suppress BGP-only user rows. A follow-up store test with a second ReplaceUsers call changing only LastBgpReportedAt would pin the chosen semantics.

indexer/db/clickhouse/migrations/20260708000001_users_bgp_status.sql
```diff
@@ -0,0 +1,100 @@
+-- +goose Up
+
+-- Add onchain BGP session state (User account: bgp_status, last_bgp_up_at,
+-- last_bgp_reported_at) to the users dimension so consumers can distinguish a
+-- user that is not connected (BGP down) from a multicast forwarding fault.
+
+-- +goose StatementBegin
+ALTER TABLE dim_dz_users_history
+    ADD COLUMN IF NOT EXISTS bgp_status String DEFAULT '' AFTER subscribers,
+    ADD COLUMN IF NOT EXISTS last_bgp_up_at UInt64 DEFAULT 0 AFTER bgp_status,
```
ben-dz: `last_bgp_up_at`/`last_bgp_reported_at` hold ledger slots (onchain: `user.last_bgp_reported_at = slot`), but the `_at` suffix on a UInt64 reads as a unix timestamp. The SQL agent introspects only table/name/type from system.columns (agent/pkg/workflow/schema.go:109-119) — the Go doc comments are invisible to it — so it will plausibly generate `toDateTime(last_bgp_up_at)` and answer wrongly. The migration hasn't shipped: rename to `last_bgp_up_slot`/`last_bgp_reported_slot` now, or add ClickHouse COMMENT clauses and surface them to the agent, plus SQL_CONTEXT.md notes when the follow-up consumes these fields.

indexer/db/clickhouse/migrations/20260708000001_users_bgp_status.sql
```diff
@@ -0,0 +1,100 @@
...
+-- +goose StatementBegin
+ALTER TABLE stg_dim_dz_users_snapshot
+    ADD COLUMN IF NOT EXISTS bgp_status String DEFAULT '' AFTER subscribers,
+    ADD COLUMN IF NOT EXISTS last_bgp_up_at UInt64 DEFAULT 0 AFTER bgp_status,
+    ADD COLUMN IF NOT EXISTS last_bgp_reported_at UInt64 DEFAULT 0 AFTER last_bgp_up_at;
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+-- Recreate dz_users_current to expose the BGP columns.
+CREATE OR REPLACE VIEW dz_users_current
```
ben-dz, line 33: Companion to the churn finding: `dz_users_current` is a row_number() window over the entire history table and is read by hot API paths and the page cache, so with keepalive-driven rows the scan cost grows linearly with time even when nothing changes, and a history row no longer means "config changed". Resolved by the same fix as the attrs_hash finding; if churn is knowingly accepted, say so in the migration comment and consider a TTL on dim_dz_users_history.

indexer/pkg/dz/serviceability/view.go
```diff
@@ -542,17 +546,20 @@ func convertUsers(onchain []serviceability.User) []User {
...
+			Status:            statusString(user.Status),
+			Kind:              user.UserType.String(),
+			ClientIP:          net.IP(user.ClientIp[:]),
+			DZIP:              net.IP(user.DzIp[:]),
+			DevicePK:          solana.PublicKeyFromBytes(user.DevicePubKey[:]).String(),
+			TenantPK:          solana.PublicKeyFromBytes(user.TenantPubKey[:]).String(),
+			TunnelID:          user.TunnelId,
+			Publishers:        publishers,
+			Subscribers:       subscribers,
+			BgpStatus:         serviceability.BGPStatus(user.BgpStatus).String(),
```
ben-dz: `BGPStatus.String()` returns `"BGPStatus(N)"` for values outside 0-2 (SDK default arm), so a future onchain enum variant lands as a non-vocabulary string, violating the documented `"up" | "down" | "unknown"` contract at view.go:141. Normalize out-of-range values to `"unknown"` (mirroring the `statusString` precedent) and add a unit case for e.g. `BgpStatus: 99` — the tests currently cover only up/down. Relatedly, `DEFAULT ''` in the migration creates a fourth state on pre-migration history rows; `DEFAULT 'unknown'` is semantically accurate for "never reported" and free at read time, or document that empty = predates ingestion.

indexer/db/clickhouse/migrations/20260708000001_users_bgp_status.sql
```diff
@@ -0,0 +1,100 @@
+-- +goose Up
+
+-- Add onchain BGP session state (User account: bgp_status, last_bgp_up_at,
+-- last_bgp_reported_at) to the users dimension so consumers can distinguish a
+-- user that is not connected (BGP down) from a multicast forwarding fault.
+
+-- +goose StatementBegin
+ALTER TABLE dim_dz_users_history
+    ADD COLUMN IF NOT EXISTS bgp_status String DEFAULT '' AFTER subscribers,
```
ben-dz: The `AFTER subscribers` placement is load-bearing and undocumented: the staging insert is `INSERT INTO stg_…` with no column list and positional batch.Append (dim_write.go:525-575), so physical column order must exactly match PayloadColumns() order. This PR gets it right (verified against the 20250117/20250126/20250304 DDL history), and forward deploys are safe since migrations run in-binary — but a binary rollback without the Down migration breaks user ingestion with an arity error. Add a one-line comment stating the AFTER placement must match UserSchema.PayloadColumns() order; longer term, give loadSnapshotIntoStaging an explicit column list.

indexer/pkg/dz/serviceability/store_test.go
```diff
@@ -329,6 +329,64 @@ func TestLake_Serviceability_Store_ReplaceUsers(t *testing.T) {
 		require.Equal(t, "10.0.0.2", current["dz_ip"])
 		require.Equal(t, devicePK, current["device_pk"])
 	})
+
+	t.Run("persists onchain BGP status through to dz_users_current", func(t *testing.T) {
```
ben-dz, line 333: Good round-trip test (exercises the real migration via the shared container), but it doesn't pin the change-detection behavior the follow-up depends on: add a second ReplaceUsers call with only `LastBgpReportedAt` changed and assert the resulting history row count, pinning whichever attrs_hash semantics is chosen.

indexer/pkg/dz/serviceability/schema.go
```diff
@@ -163,6 +163,9 @@ func (s *UserSchema) PayloadColumns() []string {
 		"tunnel_id:INTEGER",
 		"publishers:VARCHAR",
 		"subscribers:VARCHAR",
+		"bgp_status:VARCHAR",
+		"last_bgp_up_at:BIGINT",
```
ben-dz: Nit: the `:BIGINT` type label is never parsed (dim.go extracts names only), but BIGINT (signed) vs the migration's UInt64 is misleading to the next reader; other schemas already use ClickHouse spellings (TopologySchema uses UInt8).

ben-dz (changes requested):
The plumbing is clean and correctly mirrors the tenant_pk pattern (positional staging order verified, name-based history insert, solid round-trip test, no security findings), but one High issue should be resolved before the migration ships: `attrs_hash` covers all payload columns, and `last_bgp_reported_at` is rewritten onchain as a ~6-hourly keepalive per user, so every user gains a new `dim_dz_users_history` row ~4×/day forever and each one surfaces in `entity_changes_v`/the timeline as a contentless "User X updated (0 fields)" event competing for the LIMIT 500 window — the "purely additive, no consumer behavior changes" claim doesn't hold. Also worth fixing pre-ship (cheap now, expensive later): the `*_at` columns hold ledger slots, not timestamps, and the SQL agent only sees name/type, so it will plausibly emit `toDateTime(last_bgp_up_at)`; and unknown enum values ingest as `"BGPStatus(N)"` while pre-migration rows read as `''`, both outside the documented up/down/unknown contract.

- Low: local dev `lake.*` remote proxy tables snapshot the remote schema at creation (admin/remotetables/setup.go:145) — they won't expose the new columns until the indexer restarts with --setup-remote-tables after the prod migration lands. Self-healing; worth a line in the PR description for local dev users.
- Note: the AI agent's SQL_CONTEXT.md and the multicast health views don't reference the new columns yet — fine for this ingestion-first PR, but the follow-up should update SQL_CONTEXT.md (with evals, per repo policy) and document the ''-means-pre-migration semantics if the default isn't changed.
- Security review: no findings — values are parameter-bound (never interpolated into SQL), the enum conversion cannot panic, onchain writes are gated to the device's registered metrics publisher, the u64→UInt64 path is lossless, and BGP state is public onchain data.
- Verification: go vet, CGO_ENABLED=0 go build ./indexer/..., golangci-lint (0 issues), and gofmt all clean. The package's tests require a ClickHouse testcontainer and could not be executed in this environment (no Docker daemon); they were reviewed by reading. Migration Up/Down ordering, IF (NOT) EXISTS idempotency, tombstone/empty-snapshot paths, and dz_users_current consumers (no positional SELECT * reads) all checked clean.

ben-malbeclabs:
Thanks both — pushed a revision addressing the review.

**Reworked the design around the churn finding (Ben, HIGH):** dropped `last_bgp_up_at` / `last_bgp_reported_at` entirely and ingest only `bgp_status`. `last_bgp_reported_at` was the ~6-hourly keepalive driving the per-user ~4×/day history churn; `bgp_status` changes only on real transitions, so there's no ongoing churn and the "additive" claim now holds. Added a change-detection guard test (re-ingesting unchanged `bgp_status` writes no new history row; a transition writes exactly one).

Point by point:
- **Enum contract** — `convertUsers` now normalizes out-of-range values to `unknown` (new test with `BgpStatus: 99`); column is `DEFAULT 'unknown'` (not `''`), and consumers treat `!= 'down'` as fail-open.
- **`*_at` naming / SQL-agent `toDateTime` risk** — moot: those slot columns are gone.
- **`AFTER` placement** — documented in the migration that physical order must match `UserSchema.PayloadColumns()`.
- **Local-dev remote proxy** — noted in the PR description (self-healing on indexer restart with `--setup-remote-tables`).
- **SQL_CONTEXT.md + evals** — deferred to the health follow-up (#686) as agreed, since nothing consumes the column in this PR.

**Juan** — both nits are now moot: no more `BIGINT` label (only `bgp_status:VARCHAR` remains), and the slot/timestamp ambiguity is gone with the slots.

indexer/pkg/dz/serviceability/schema.go
ben-malbeclabs: Moot now — the slot columns were dropped, so only `bgp_status:VARCHAR` remains (no more `BIGINT` label).

indexer/pkg/dz/serviceability/view.go
ben-malbeclabs: Moot — `last_bgp_up_at`/`last_bgp_reported_at` are no longer ingested; only `bgp_status` is carried now.

indexer/pkg/dz/serviceability/schema.go
ben-malbeclabs: Fixed by dropping both keepalive slot columns and ingesting only `bgp_status`, which changes only on real up/down transitions — so no per-user keepalive churn and no contentless timeline events. Added a change-detection guard test (unchanged `bgp_status` → no new history row; a transition → exactly one).

indexer/db/clickhouse/migrations/20260708000001_users_bgp_status.sql
ben-malbeclabs: Moot — the `*_at` slot columns are dropped, so there's nothing for the SQL agent to misread as a timestamp.

indexer/db/clickhouse/migrations/20260708000001_users_bgp_status.sql
ben-malbeclabs: Resolved by the same drop: with no keepalive-driven rows, `dz_users_current`'s window scan no longer grows over time and a history row again means "config changed".

indexer/pkg/dz/serviceability/view.go
ben-malbeclabs: Fixed — `convertUsers` now uses a `bgpStatusString` helper that normalizes any out-of-range enum value to `"unknown"` (added a `BgpStatus: 99` test), and the column is `DEFAULT 'unknown'` so pre-migration rows read as a valid state.

indexer/db/clickhouse/migrations/20260708000001_users_bgp_status.sql
ben-malbeclabs: Documented — the migration now states the `AFTER subscribers` placement must match `UserSchema.PayloadColumns()` order.

indexer/pkg/dz/serviceability/store_test.go
ben-malbeclabs: Added — a second `ReplaceUsers` with unchanged `bgp_status` asserts no new `dim_dz_users_history` row, and a real transition asserts exactly one. Pins the no-churn semantics.

indexer/pkg/dz/serviceability/schema.go
ben-malbeclabs: Moot — slot columns dropped; no `BIGINT` labels remain in `UserSchema`.

ben-dz (approved):
Re-review of commit 5f3eff3: all findings from the initial review are addressed. The High churn/timeline issue is resolved the clean way — the volatile `last_bgp_up_at`/`last_bgp_reported_at` keepalive slots are no longer ingested at all, and a new store test pins that an unchanged `bgp_status` re-ingest writes no history row while a real transition writes exactly one. Enum values outside up/down now normalize to `unknown` (unit-tested with 99), the column default is `'unknown'` instead of `''`, the load-bearing `AFTER` placement is documented in the migration, and the PR description covers the deliberately-not-ingested rationale and the local-dev proxy refresh. vet/build/lint clean. Two non-blocking notes for follow-up #686: add `bgp_status` to `entity_changes_v`'s tracked-field list so BGP flaps don't render as "updated (0 fields)", and expect a one-time timeline spike at deploy as every user re-versions under the new hash tuple.

- Follow-up (#686): a real bgp_status transition still emits an entity_changes_v 'updated' event with empty changed_fields (rendered as "User X updated (0 fields)") since bgp_status isn't in the view's lag list. Transition-driven and low-frequency now, so non-blocking — but add it to the tracked fields (or suppress empty-changed_fields user rows) so flaps render meaningfully.
- Deploy expectation: the first post-migration snapshot re-versions every user (attrs_hash tuple gained a column) — a one-off spike of timeline events, same as the tenant_pk precedent. No action needed.

juan-malbeclabs (approved):
Verifiqué cada hallazgo contra el código real y todos quedaron resueltos:

- **Churn de `attrs_hash` / crecimiento del window scan de `dz_users_current`**: resuelto de raíz — se eliminaron ambos slots de keepalive y solo se ingiere `bgp_status`, que cambia únicamente en transiciones up/down reales. Una fila de historia vuelve a significar "cambió la config".
- **Nombres `*_at` / `BIGINT` vs `UInt64`**: moot al eliminar las columnas slot.
- **`BGPStatus.String()` fuera de rango**: normalizado a `"unknown"` vía `bgpStatusString`, con test `BgpStatus: 99`. `DEFAULT 'unknown'` en la migración.
- **Placement `AFTER subscribers`**: documentado (debe coincidir con `UserSchema.PayloadColumns()`).
- **Test de detección de cambios**: añadido — re-ingesta sin cambio no crea fila, transición sí.

¡Gracias Ben por atender todo con cuidado! 🙏

PR #686, opened 2026-07-08
multicast health: reclassify BGP-down users as disconnected

ben-malbeclabs:
## Summary

Adds a distinct **`disconnected`** multicast health state for users whose onchain BGP session is down, so the Health tab stops reporting client-side outages as multicast forwarding faults.

A user with BGP down has no route to/from its source, so it has no `(S,G)` mroute and no OIF entry — previously reported as "not seen as RPF interface" / "not in any OIF list". ~100% of the `edge-solana-shreds` group's 92 "unhealthy" publishers were simply BGP-down (same on device pages, both publisher and subscriber sides), drowning out real faults.

Depends on #685 (ingests `dz_users_current.bgp_status`); based on that branch until it merges.

- **Views:** `health_multicast_user` short-circuits to `disconnected` when `bgp_status='down'`; `health_multicast_user_rate` propagates it **and** excludes disconnected publishers from `group_publisher_total` (so a counter-less BGP-down publisher doesn't force co-group subscribers to `monitoring_gap`/`unknown`); `health_publisher_subscriber_path` is `disconnected` if either endpoint is down; `health_mroute` excludes BGP-down publishers from the active-publisher set (cleared 32 false `(*,G)` unhealthy rows for the shreds group).
- **API:** `disconnected` count bucket + accepted on the `health=` filter (both the delivery-entity allowlist and the search-token path); sort orders it below `unknown` via a shared `healthStatusSeverityOrderSQL`; unrecognized statuses fold into `Unknown` so buckets always sum to total.
- **Web:** renders `disconnected` (sky/blue, distinct from red faults) across the group Health tab (badges, counts row, legend, filter help), device/link delivery panels, and user detail page.

### Deferred (documented in migration headers)
- **`health_missing_sg`** keeps the old active-publisher logic — left as-is since no API handler reads it; a follow-up can apply the same exclusion or drop the view.
- **BGP freshness:** onchain `down` takes precedence over observed dataplane, so a stale `down` can mask a recovered path until re-ingested. A freshness gate needs `last_bgp_reported_at`, which #685 deliberately doesn't ingest; noted as a possible follow-up. `''`/`unknown` fail open to old behavior.

## Testing Verification

- `health_multicast_user`: BGP-down → `disconnected`; BGP-up with the same missing RPF entry stays `unhealthy`.
- `health_publisher_subscriber_path`: BGP-down endpoint → `disconnected`; BGP-up with same missing endpoint → `unhealthy`.
- `health_mroute`: BGP-down publisher excluded from the active set; a BGP-up publisher with no `(S,G)` still counts.
- `health_multicast_user_rate`: `disconnected` verdict propagates; a disconnected publisher does not create a monitoring gap for a co-group subscriber (stays `healthy`).
- API: `health=disconnected` → 200 on the delivery entity endpoint.

### DIFF
--- a/api/handlers/multicast_health.go
+++ b/api/handlers/multicast_health.go
@@ -197,6 +197,16 @@ func getCountsBucket(c *MulticastHealthCounts, source string) *MulticastHealthSt
 	return nil
 }
 
+// healthStatusSeverityOrderSQL ranks multicast health rows so the most
+// actionable land first: unhealthy → degraded → unknown → disconnected →
+// healthy. 'disconnected' is a client-side BGP outage, not a network fault, so
+// it sorts below unknown. Shared by every health query builder that paginates
+// so the ordering (and the disconnected placement) stays consistent.
+const healthStatusSeverityOrderSQL = `multiIf(health_status = 'unhealthy', 0, ` +
+	`health_status = 'degraded', 1, ` +
+	`health_status = 'unknown', 2, ` +
+	`health_status = 'disconnected', 3, 4)`
+
 func addStatusCount(bucket *MulticastHealthStatusCounts, status string, n uint64) {
 	if bucket == nil {
 		return
@@ -211,6 +221,10 @@ func addStatusCount(bucket *MulticastHealthStatusCounts, status string, n uint64
 	case "disconnected":
 		bucket.Disconnected += n
 	case "unknown":
+		bucket.Unknown += n
+	default:
+		// Fold any unrecognized status (e.g. a future value seen by an old
+		// binary) into Unknown so the buckets always sum to Total.
 		bucket.Unknown += n
 	}
 	bucket.Total += n
--- a/api/handlers/multicast_health.go
+++ b/api/handlers/multicast_health.go
@@ -208,6 +208,8 @@ func addStatusCount(bucket *MulticastHealthStatusCounts, status string, n uint64
 		bucket.Degraded += n
 	case "unhealthy":
 		bucket.Unhealthy += n
+	case "disconnected":
+		bucket.Disconnected += n
 	case "unknown":
 		bucket.Unknown += n
 	}
--- a/api/handlers/multicast_health_users.go
+++ b/api/handlers/multicast_health_users.go
@@ -235,16 +235,8 @@ func (a *API) queryMulticastHealthUsers(ctx context.Context, whereClause string,
 			health_status
 		FROM health_multicast_user_rate
 		WHERE ` + whereClause + `
-		-- Sort actionable rows first (unhealthy → degraded → unknown →
-		-- disconnected → healthy) so paginated consumers land on the rows
-		-- operators most need to see. 'disconnected' is a client-side BGP
-		-- outage, not a network fault, so it sorts below unknown.
-		ORDER BY
-			multiIf(health_status = 'unhealthy', 0,
-			        health_status = 'degraded',  1,
-			        health_status = 'unknown',   2,
-			        health_status = 'disconnected', 3,
-			                                     4),
+		-- Sort actionable rows first (see healthStatusSeverityOrderSQL).
+		ORDER BY ` + healthStatusSeverityOrderSQL + `,
 			multicast_group_code, user_pk` + limitClause + `
 		SETTINGS max_execution_time = 30, timeout_before_checking_execution_speed = 0
 	`
--- a/api/handlers/multicast_health_users.go
+++ b/api/handlers/multicast_health_users.go
@@ -235,13 +235,16 @@ func (a *API) queryMulticastHealthUsers(ctx context.Context, whereClause string,
 			health_status
 		FROM health_multicast_user_rate
 		WHERE ` + whereClause + `
-		-- Sort actionable rows first (unhealthy → degraded → unknown → healthy)
-		-- so paginated consumers land on the rows operators most need to see.
+		-- Sort actionable rows first (unhealthy → degraded → unknown →
+		-- disconnected → healthy) so paginated consumers land on the rows
+		-- operators most need to see. 'disconnected' is a client-side BGP
+		-- outage, not a network fault, so it sorts below unknown.
 		ORDER BY
 			multiIf(health_status = 'unhealthy', 0,
 			        health_status = 'degraded',  1,
 			        health_status = 'unknown',   2,
-			                                     3),
+			        health_status = 'disconnected', 3,
+			                                     4),
 			multicast_group_code, user_pk` + limitClause + `
 		SETTINGS max_execution_time = 30, timeout_before_checking_execution_speed = 0
 	`
--- a/indexer/db/clickhouse/migrations/20260708000003_health_user_rate_disconnected.sql
+++ b/indexer/db/clickhouse/migrations/20260708000003_health_user_rate_disconnected.sql
@@ -1,8 +1,13 @@
 -- +goose Up
 --
 -- Propagate the 'disconnected' control-plane verdict (BGP down, from
--- health_multicast_user) through the combined rate verdict. Requires
--- 20260708000002.
+-- health_multicast_user) through the combined rate verdict, and exclude
+-- disconnected publishers from group_publisher_total so they do not force
+-- co-group subscribers to monitoring_gap. Requires 20260708000002.
+--
+-- Rollback: must be rolled back together with 000002. Rolling back this
+-- migration alone, while 000002 still emits 'disconnected', routes that verdict
+-- into the restored rate matrix's 'unknown' catch-all instead of 'unhealthy'.
 
 -- +goose StatementBegin
 CREATE OR REPLACE VIEW health_multicast_user_rate
@@ -46,7 +51,10 @@ group_publisher_total AS (
         ON ur.ur_device_pk = h.user_device_pk
        AND ur.ur_user_tunnel_id = h.user_tunnel_id
        AND ur.ur_user_pk = h.user_pk
-    WHERE h.mode IN ('P', 'P+S')
+    -- Exclude BGP-down (disconnected) publishers: no counter row, so counting them
+    -- forces subscribers to monitoring_gap/unknown (the masking this feature removes).
+    -- Mirrors health_mroute (000005).
+    WHERE h.mode IN ('P', 'P+S') AND h.health_status != 'disconnected'
     GROUP BY gpt_multicast_group_pk
 ),
 -- Inner layer: compute rate_status_reason / observed_bps_5m / expected_bps_5m
--- a/indexer/db/clickhouse/migrations/20260708000003_health_user_rate_disconnected.sql
+++ b/indexer/db/clickhouse/migrations/20260708000003_health_user_rate_disconnected.sql
@@ -1,0 +1,251 @@
+-- +goose Up
+--
+-- Propagate the 'disconnected' control-plane verdict (BGP down, from
+-- health_multicast_user) through the combined rate verdict. Requires
+-- 20260708000002.
+
+-- +goose StatementBegin
+CREATE OR REPLACE VIEW health_multicast_user_rate
+AS
+WITH
+-- Aggregating + argMax collapses two sources of duplication into one CTE:
+--   1. Multiple buckets per (device, tunnel, user) → keep max(bucket_ts).
+--   2. ReplacingMergeTree part-row duplicates within that bucket → argMax
+--      picks one (values are identical by RMT contract, so tie-break safe).
+-- ur_present = 1 lets the outer LEFT JOIN distinguish "no row" (default
+-- 0) from "row exists with rate 0" (sentinel 1).
+user_rates AS (
+    SELECT
+        device_pk AS ur_device_pk,
+        user_tunnel_id AS ur_user_tunnel_id,
+        user_pk AS ur_user_pk,
+        max(bucket_ts) AS ur_bucket_ts,
+        argMax(max_in_bps, bucket_ts) AS ur_max_in_bps,
+        argMax(max_out_bps, bucket_ts) AS ur_max_out_bps,
+        1 AS ur_present
+    FROM device_interface_rollup_5m
+    WHERE bucket_ts >= now() - INTERVAL 15 MINUTE
+      AND user_tunnel_id IS NOT NULL
+      AND user_pk != ''
+    GROUP BY ur_device_pk, ur_user_tunnel_id, ur_user_pk
+),
+-- Per-group publisher RX total + missing-publisher count so subscribers
+-- (and P+S users on the subscriber side) can reconcile their TX against
+-- expected sum-of-publishers. gpt_missing_publisher_count > 0 means at
+-- least one publisher in the group has no counter row, which forces
+-- subscribers to 'monitoring_gap' (we refuse to reconcile against a
+-- deflated expected).
+group_publisher_total AS (
+    SELECT
+        h.multicast_group_pk AS gpt_multicast_group_pk,
+        sumIf(ur.ur_max_in_bps, ur.ur_present = 1) AS gpt_total_publisher_rx_bps,
+        countIf(ur.ur_present = 0) AS gpt_missing_publisher_count,
+        count() AS gpt_publisher_count
+    FROM health_multicast_user h
+    LEFT JOIN user_rates ur
+        ON ur.ur_device_pk = h.user_device_pk
+       AND ur.ur_user_tunnel_id = h.user_tunnel_id
+       AND ur.ur_user_pk = h.user_pk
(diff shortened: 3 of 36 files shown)


indexer/db/clickhouse/migrations/20260708000003_health_user_rate_disconnected.sql
```diff
@@ -0,0 +1,251 @@
...
+        h.multicast_group_pk AS gpt_multicast_group_pk,
+        sumIf(ur.ur_max_in_bps, ur.ur_present = 1) AS gpt_total_publisher_rx_bps,
+        countIf(ur.ur_present = 0) AS gpt_missing_publisher_count,
+        count() AS gpt_publisher_count
+    FROM health_multicast_user h
+    LEFT JOIN user_rates ur
+        ON ur.ur_device_pk = h.user_device_pk
+       AND ur.ur_user_tunnel_id = h.user_tunnel_id
+       AND ur.ur_user_pk = h.user_pk
+    WHERE h.mode IN ('P', 'P+S')
```
ben-dz: `group_publisher_total` keeps `WHERE h.mode IN ('P','P+S')` with no BGP/disconnected exclusion. A BGP-down publisher has no counter row in the 15-minute rollup window → `ur_present = 0` → `gpt_missing_publisher_count > 0` → every subscriber in the group gets `rate_status_reason='monitoring_gap'` → `rate_status='unknown'`, so CP-healthy subscribers roll up to `unknown` indefinitely. Migration 000005 excludes BGP-down publishers from `health_mroute`'s expected set for exactly this "legitimately absent" reasoning; mirror it here (e.g. `AND h.health_status != 'disconnected'`). Test: one BGP-down counter-less publisher + one healthy pub/sub pair; assert the subscriber stays `healthy`, not `unknown`.

indexer/db/clickhouse/migrations/20260708000004_health_pubsub_path_disconnected.sql
```diff
@@ -0,0 +1,264 @@
...
+    s.s_device_code AS subscriber_device_code,
+
+    has(coalesce(pi.pi_iif_tunnel_ids, []), p.p_tunnel_id) AS publisher_endpoint_observed,
+    has(coalesce(so.so_oif_tunnel_ids, []), s.s_tunnel_id) AS subscriber_endpoint_observed,
+
+    has(coalesce(pi.pi_iif_tunnel_ids, []), p.p_tunnel_id)
+        AND has(coalesce(so.so_oif_tunnel_ids, []), s.s_tunnel_id) AS endpoints_reconciled,
+
+    multiIf(
+        p.p_bgp_status = 'down' OR s.s_bgp_status = 'down', 'disconnected',
```
ben-dz, line 105: The `disconnected` branch precedes the endpoints-observed branch, and none of the views consult `last_bgp_reported_at`/`last_bgp_up_at` (ingested by the dependency migration). A stale onchain `down` (agent dies after reporting down, session recovers) renders a live — possibly genuinely faulty — path as `disconnected` until the next report is ingested, inverting the PR's goal. `''`/`'unknown'` failing open to old behavior is the safe direction. At minimum document the "onchain status wins over observed dataplane" precedence in the view headers/legend; consider gating the short-circuit on report freshness in a follow-up.

indexer/db/clickhouse/migrations/20260708000003_health_user_rate_disconnected.sql
```diff
@@ -0,0 +1,251 @@
...
+    if(rate_status_reason IN ('active', 'reconciled'), 'reconciled',
+        if(rate_status_reason = 'mismatch', 'mismatch', 'unknown')
+    ) AS rate_status,
+    -- Combined verdict per the matrix.
+    --                | rate=reconciled | rate=mismatch | rate=unknown
+    --  CP=healthy    | healthy         | degraded      | unknown
+    --  CP=degraded   | degraded        | unhealthy     | degraded
+    --  CP=unhealthy  | unhealthy       | unhealthy     | unhealthy
+    multiIf(
+        control_plane_status = 'disconnected', 'disconnected',
```
ben-dz, line 126: No test covers this propagation: `health_multicast_user_rate_test.go` is untouched, so nothing exercises `control_plane_status='disconnected' → 'disconnected'` nor its interaction with `no_data`/`monitoring_gap` (where the `group_publisher_total` issue hides). Add a subtest alongside the existing rate reconciliation tests.

api/handlers/multicast_health_users.go
```diff
@@ -235,13 +235,16 @@ func (a *API) queryMulticastHealthUsers(ctx context.Context, whereClause string,
...
+		-- Sort actionable rows first (unhealthy → degraded → unknown →
+		-- disconnected → healthy) so paginated consumers land on the rows
+		-- operators most need to see. 'disconnected' is a client-side BGP
+		-- outage, not a network fault, so it sorts below unknown.
 		ORDER BY
 			multiIf(health_status = 'unhealthy', 0,
 			        health_status = 'degraded',  1,
 			        health_status = 'unknown',   2,
-			                                     3),
+			        health_status = 'disconnected', 3,
```
ben-dz: This severity-sort `multiIf` is now copy-pasted in four query builders (here, `multicast_health_paths.go:190-194`, `multicast_delivery_entity_health_queries.go:44,132`), plus the parallel `worstHealth` ordering in `multicast-delivery-panel.tsx:97-104`. This PR had to touch all four in lockstep and still missed the `health=` allowlist and the Health tab counts row. Extract a shared `healthStatusSeverityOrderSQL` constant carrying the "disconnected sorts below unknown" comment.

api/handlers/multicast_health.go
```diff
@@ -208,6 +208,8 @@ func addStatusCount(bucket *MulticastHealthStatusCounts, status string, n uint64
 		bucket.Degraded += n
 	case "unhealthy":
 		bucket.Unhealthy += n
+	case "disconnected":
```
ben-dz, line 221: `addStatusCount` silently drops unrecognized statuses (they bump `Total` but no bucket — the counts-don't-sum failure, at the API layer, for any future status against an old API), while `addEntityStatusCount` defaults them into `Unknown`. Give this the same `default: bucket.Unknown += n`, or merge the two field-identical structs/helpers (`MulticastHealthStatusCounts` / `MulticastEntityHealthStatusCounts`).

ben-dz (changes requested):
The core design is sound — the reclassification lives in the ClickHouse views with faithful Down migrations, a sensible severity ordering (disconnected below unknown, rationale documented), and good view-level test contrast between BGP-down and genuinely faulty users. But three high findings each undercut the feature: the delivery entity endpoints reject `health=disconnected` with 400 (allowlist at `api/handlers/multicast_delivery_entity.go:332` not updated); the group Health tab's `CountsRow` (`multicast-group-health-tab.tsx:353-357`) omits the disconnected bucket so the visible counts no longer sum to total and the new state is invisible on its primary surface; and the rate view's `group_publisher_total` still counts BGP-down publishers, so one counter-less disconnected publisher forces every subscriber in the group to `monitoring_gap`/`unknown` — the same masking this PR set out to remove, which `health_mroute` (migration 000005) already fixes. Mediums: `health_missing_sg` still treats BGP-down publishers as active (now disagrees with `health_mroute`); a stale onchain `down` overrides observed dataplane state with no `last_bgp_reported_at` freshness check; no test covers the rate-view propagation; the severity-sort `multiIf` is copy-pasted across four query builders. Lows and open questions in inline/general comments.

- HIGH — `api/handlers/multicast_delivery_entity.go:332`: the `health=` param allowlist (`healthy|degraded|unhealthy|unknown`) was not updated, so `?health=disconnected` returns 400 "invalid health" on the device/link delivery endpoints even though the same PR adds a `disconnected` count bucket to their responses and the web `MulticastHealthStatus` type now permits sending it. The PR description's "generic status: filter accepts it with no allowlist change" holds only for the search-token path (`multicast_health_users.go:31`). Add `"disconnected"` to the allowlist (ideally derive all three status enumerations from one shared set) and extend the existing `health=broken → 400` test with a `health=disconnected → 200` case.
- HIGH — `web/src/components/multicast-group-health-tab.tsx:353-357`: `CountsRow` renders healthy/degraded/unhealthy/unknown then `/ {counts.total}`, omitting `counts.disconnected` while the API adds disconnected rows to `Total` unconditionally. On the shreds group (92 BGP-down publishers) the visible buckets will be ~92 short of the total, and the new state is invisible on exactly the surface meant to expose it. Add a sky-colored `{counts.disconnected} disconnected` span, ordered as in the delivery panel's `HealthCountsStrip` (which was updated).
- MEDIUM — `indexer/db/clickhouse/migrations/20260603000002_multicast_health_views.sql:163-175`: `health_missing_sg`'s `active_pubs` CTE is a copy of the active-publisher logic migration 000005 just changed, minus the new `bgp_status != 'down'` filter — a BGP-down publisher still produces `fhr_missing`/`downstream_missing` rows on every device, so the two views now disagree about the same fact. No API handler reads it, but operators and the SQL agent can. Follow-up migration applying the same exclusion (or drop the view if dead).
- LOW — Down migrations drop the explanatory header comments, and 000002/000003 should be rolled back together: rolling back 000003 alone leaves 000002's `disconnected` verdict falling into the restored matrix's `unknown` catch-all rather than the pre-PR `unhealthy`. Note the pairing in the migration headers. (Down bodies otherwise restore the prior definitions exactly — verified.)
- LOW — Mixed-version deploy window: until the API redeploys after the views migrate, the old API drops `disconnected` rows from every count bucket while still counting them in `Total`, and the old sort ranks them with `healthy`. Transient; sequence indexer → API closely.
- LOW — Web status color maps are now five near-identical copies (`multicast-delivery-panel.tsx:30-36`, `multicast-group-health-tab.tsx:163-176,253-262`, `user-detail-page.tsx:23-44`); consider one shared module next to `MulticastHealthStatus`.
- LOW — `health_multicast_user` consumes `bgp_status` but doesn't expose it (or `last_bgp_up_at`) as an output column; the only trace is the `mismatch_reason` string. Consider surfacing `user_bgp_status` in a follow-up. Security posture is otherwise clean: the value is a bounded enum through the SDK, inserts are parameterized, and the PR introduces no injection or XSS surface.
- QUESTION — `bgp_status IN ('', 'unknown')` counts as an active publisher everywhere (`= 'down'` / `!= 'down'` checks): a user whose agent never reported still marks a `(*,G)` unhealthy for its missing `(S,G)`. Fail-open to old behavior looks intended — worth stating in the migration header.
- QUESTION — All-publishers-down asymmetry: with every publisher of a group BGP-down, the `(*,G)` becomes `healthy` (000005: empty active set), pub→sub pairs become `disconnected` (000004), but a BGP-up subscriber still shows `unhealthy` ("no (S,G) mroutes", 000002 mode='S'). Three verdicts for one scenario across the three views — intended?
- QUESTION — `multicast-delivery-panel.tsx:315`: the "affected" rollup counts `unhealthy + degraded` only; excluding `disconnected` looks deliberate (client-side outage) — confirm. Also a copy nit: the Health tab disclaimer (`multicast-group-health-tab.tsx:513`) still says non-jump-device users "will appear as unhealthy"; some now appear as `disconnected`.

ben-malbeclabs:
Thanks — pushed a revision. All three HIGH findings fixed, plus the mediums. Rebased onto the updated #685.

**HIGH**
- **`health=disconnected` → 400** — added `disconnected` to the delivery-entity allowlist; extended the `health=broken → 400` test with a `health=disconnected → 200` case.
- **Group tab counts omitted disconnected** — `CountsRow` now renders a sky `{disconnected}` span, so buckets sum to total and the state is visible on its primary surface.
- **Rate view counted BGP-down publishers** — `group_publisher_total` now excludes `health_status = 'disconnected'`, so a counter-less disconnected publisher no longer forces co-group subscribers to `monitoring_gap`/`unknown`. New test: one disconnected counter-less publisher + a healthy pub/sub pair → subscriber stays `healthy`.

**Medium**
- **No rate-propagation test** — added (covers `control_plane_status='disconnected' → 'disconnected'` alongside the exclusion above).
- **Severity-sort copy-pasted ×4** — extracted `healthStatusSeverityOrderSQL`, now shared by all four query builders.
- **`addStatusCount` dropped unknown statuses** — added `default: bucket.Unknown += n` to match `addEntityStatusCount`, so buckets always sum to total.
- **`health_missing_sg` disagrees** — left as-is (no API handler reads it); flagged as a follow-up to apply the same exclusion or drop the view. Noted in the migration header.
- **Stale-`down` masking** — documented the precedence (onchain `down` wins over observed dataplane) and the staleness caveat in the migration headers. A real freshness gate needs `last_bgp_reported_at`, which #685 deliberately doesn't ingest to avoid churn — noted as a possible follow-up.

**Low / questions**
- Rollback pairing (000002↔000003) documented in both migration headers.
- Health tab disclaimer copy updated to mention `disconnected`.
- `''`/`unknown` fail-open is intended (an unreported user is never masked) — documented in the header.
- All-publishers-down asymmetry (`(*,G)` healthy / pairs disconnected / BGP-up subscriber unhealthy) is intended: each view answers a different question, and only the subscriber genuinely has no `(S,G)` to receive. Happy to revisit if you'd prefer a unified verdict.
- "Affected" rollup excluding `disconnected` is intentional (client-side outage, not a network fault).
- Five near-identical color maps and surfacing `user_bgp_status` as an output column — agreed; small follow-up, out of scope here.

indexer/db/clickhouse/migrations/20260708000003_health_user_rate_disconnected.sql
ben-malbeclabs: Fixed — `group_publisher_total` now filters `AND h.health_status != 'disconnected'`, so a counter-less BGP-down publisher no longer forces co-group subscribers to `monitoring_gap`/`unknown`. New test asserts the subscriber stays `healthy`.

indexer/db/clickhouse/migrations/20260708000004_health_pubsub_path_disconnected.sql
ben-malbeclabs: Documented the precedence (onchain `down` wins over observed dataplane), the stale-`down` caveat, and the `''`/`unknown` fail-open in the migration header. A real freshness gate needs `last_bgp_reported_at`, which #685 deliberately doesn't ingest (churn) — flagged as a possible follow-up.

indexer/db/clickhouse/migrations/20260708000003_health_user_rate_disconnected.sql
ben-malbeclabs: Added `TestHealthMulticastUserRate_DisconnectedPublisherNotCountedAsMissing`, which exercises `control_plane_status='disconnected' → 'disconnected'` propagation alongside the `group_publisher_total` exclusion.

api/handlers/multicast_health_users.go
ben-malbeclabs: Extracted `healthStatusSeverityOrderSQL`; all four query builders now reference the shared constant (carrying the "disconnected sorts below unknown" rationale).

api/handlers/multicast_health.go
ben-malbeclabs: Fixed — `addStatusCount` now has `default: bucket.Unknown += n`, matching `addEntityStatusCount`, so buckets always sum to `Total`.

ben-dz:
health_missing_sg—I think the follow-up note is missing: grep finds no mention of health_missing_sg in any 20260708 migration or anywhere outside the original 20260603 file and its tests. Perhaps file the follow-up issue so it doesn't get lost — the view still silently disagrees with health_mroute.

ben-malbeclabs:
Folded in the non-blocking `entity_changes_v` follow-up from the #685 approval (5f0090c): a real BGP up/down transition now surfaces as `changed_fields=['bgp_status']` in the timeline instead of a contentless "updated (0 fields)" event — `bgp_status` (and its `lag`) added to the users section of the view, with a test covering a two-snapshot transition. The one-time deploy re-versioning spike (attrs_hash tuple gained a column) needs no action, same as the `tenant_pk` precedent.

ben-dz (approved):

ben-malbeclabs:
Good catch — and you're right that my earlier "noted in the migration header" was inaccurate; there was no such note.

Rather than track a one-liner, I applied the same exclusion to `health_missing_sg` in 32109cc: its `active_pubs` CTE now filters `AND u.bgp_status != 'down'`, matching `health_mroute` (20260708000005), so the two views no longer disagree. Extended the `health_mroute` exclusion test to also assert `health_missing_sg` omits the BGP-down publisher and still reports the BGP-up one. No API consumer today — this just keeps it consistent if it's ever surfaced.

PR #688, opened 2026-07-09
api: make page-cache refresh interval + concurrency env-configurable

nikw9944:
## Summary

Staging's page-cache errors (`geo concentration` cache hitting its deadline) come from the refresh burst — ~28 entries, up to 8 concurrent, every 30s — oversubscribing staging's smaller self-hosted ClickHouse. Rather than grow that instance, this lets us **spread the load over time** per-environment.

- `refreshInterval` (30s) and `pageCacheRefreshConcurrency` (8) become env-overridable via **`PAGE_CACHE_REFRESH_INTERVAL`** and **`PAGE_CACHE_REFRESH_CONCURRENCY`**, read once in `worker.Start` before the worker/workflow start.
- **Prod defaults are unchanged** (30s / 8) — CH Cloud handles that fine. Staging will set gentler values (lower concurrency, longer interval) via k8s env to flatten the peak.
- Both are read once at startup and then constant for the worker's life, so the workflow's use of `refreshInterval` stays deterministic across replay/continue-as-new within a deployment (no per-decision env reads).

## Testing Verification

- Unit tests for the env parsing (`envDuration`/`envInt`): unset/invalid → default, valid → parsed.

## Rollout (two-phase, to avoid a one-time replay wedge)

`PageCacheWorkflow` is now 3-arg (`PageCacheParams`). During the migration deploy's rolling overlap, an old pre-PR 2-arg worker can pick up a new run and ignore the params payload, recording 30s-constant-shaped history for it. That's harmless **only if the env vars are unset** in that deploy (old and new code produce command-identical histories; the zero-params path normalizes via `withDefaults`). If a non-default `PAGE_CACHE_REFRESH_INTERVAL` shipped in the *same* deploy, the old worker's history would diverge from the new worker's replay once — until the next deploy.

So ship it in two phases:
1. **Merge/deploy this PR with the env vars unset** (prod stays at 30s/8 defaults, no behavior change).
2. **Then set staging's values in a follow-up deploy** (infra#1905) once every worker is on the 3-arg code.

(Reverse direction is safe — zero-params histories normalize via `withDefaults`.)

### DIFF
--- a/api/worker/pagecache.go
+++ b/api/worker/pagecache.go
@@ -29,17 +29,16 @@ func Start(ctx context.Context, cfg Config) error {
 		log = slog.Default()
 	}
 
-	// Load-shaping knobs (see workflow.go), overridable per-environment. Staging
-	// runs gentler values (lower concurrency, longer interval) to spread page-cache
-	// load across its smaller self-hosted ClickHouse rather than scaling it up.
-	// Read once here, before the worker/workflow start, so both are in effect.
-	if d := envDuration("PAGE_CACHE_REFRESH_INTERVAL", 0); d > 0 {
-		refreshInterval = d
-	}
-	if n := envInt("PAGE_CACHE_REFRESH_CONCURRENCY", 0); n > 0 {
-		pageCacheRefreshConcurrency = n
-	}
-	log.Info("page-cache: refresh config", "interval", refreshInterval, "concurrency", pageCacheRefreshConcurrency)
+	// Load-shaping knobs (see PageCacheParams / loadRefreshConfig), overridable
+	// per-environment. Staging runs gentler values (lower concurrency, longer
+	// interval) to spread page-cache load across its smaller self-hosted
+	// ClickHouse rather than scaling it up. Read once here; the interval/timeout
+	// go into the workflow as arguments (replay-deterministic) and concurrency
+	// onto the activity struct.
+	params, concurrency := loadRefreshConfig(log)
+	log.Info("page-cache: refresh config",
+		"interval", params.RefreshInterval, "concurrency", concurrency,
+		"activity_timeout", params.ActivityTimeout, "continue_as_new_after", params.ContinueAsNewThreshold)
 
 	// Connect to Temporal
 	temporalHost := envOrDefault("TEMPORAL_HOST_PORT", "localhost:7233")
@@ -57,8 +56,9 @@ func Start(ctx context.Context, cfg Config) error {
 
 	// Register workflows and activities
 	activities := &Activities{
-		Log: log.With("component", "page-cache"),
-		API: cfg.API,
+		Log:                log.With("component", "page-cache"),
+		API:                cfg.API,
+		RefreshConcurrency: concurrency,
 	}
 
 	w := worker.New(tc, TaskQueue, worker.Options{})
@@ -70,7 +70,7 @@ func Start(ctx context.Context, cfg Config) error {
 	run, err := tc.ExecuteWorkflow(ctx, temporalclient.StartWorkflowOptions{
 		ID:        WorkflowID,
 		TaskQueue: TaskQueue,
-	}, PageCacheWorkflow, 0)
+	}, PageCacheWorkflow, 0, params)
 	if err != nil {
 		return fmt.Errorf("page-cache: failed to start workflow: %w", err)
 	}
@@ -161,23 +161,71 @@ func envOrDefault(key, defaultValue string) string {
 	return defaultValue
 }
 
-// envDuration returns the env var parsed as a Go duration (e.g. "90s"), or def
-// if unset/invalid.
-func envDuration(key string, def time.Duration) time.Duration {
-	if v := os.Getenv(key); v != "" {
-		if d, err := time.ParseDuration(v); err == nil {
-			return d
-		}
-	}
-	return def
+// loadRefreshConfig reads the page-cache load-shaping knobs from the environment,
+// clamping to sane bounds and warning on anything set-but-invalid/out-of-range so
+// a typo (e.g. PAGE_CACHE_REFRESH_INTERVAL=90 with no unit) is visible rather than
+// silently falling back to prod defaults. Returns the workflow params (interval,
+// activity timeout, derived continue-as-new threshold) and the activity-side
+// concurrency. Pure and side-effect-free (aside from logging) so it's unit-testable
+// without a Temporal connection.
+func loadRefreshConfig(log *slog.Logger) (PageCacheParams, int) {
+	interval := durationEnv(log, "PAGE_CACHE_REFRESH_INTERVAL", defaultRefreshInterval, minRefreshInterval, maxRefreshInterval)
+	concurrency := intEnv(log, "PAGE_CACHE_REFRESH_CONCURRENCY", defaultRefreshConcurrency, minRefreshConcurrency, maxRefreshConcurrency)
+	timeout := durationEnv(log, "PAGE_CACHE_REFRESH_TIMEOUT", defaultActivityTimeout, minActivityTimeout, maxActivityTimeout)
+
+	// Continue-as-new after roughly continueAsNewTargetWindow of wall-clock so a
+	// longer interval can't inflate per-run history toward Temporal's limits.
+	threshold := max(int(continueAsNewTargetWindow/interval), 1)
+	return PageCacheParams{
+		RefreshInterval:        interval,
+		ActivityTimeout:        timeout,
+		ContinueAsNewThreshold: threshold,
+	}, concurrency
 }
 
-// envInt returns the env var parsed as an int, or def if unset/invalid.
-func envInt(key string, def int) int {
-	if v := os.Getenv(key); v != "" {
-		if n, err := strconv.Atoi(v); err == nil {
-			return n
-		}
+// durationEnv parses key as a Go duration (e.g. "90s"), clamped to [min,max].
+// Unset → def; set-but-unparseable → def with a warn; out-of-range → clamped with
+// a warn (so an operator sees their value was adjusted).
+func durationEnv(log *slog.Logger, key string, def, min, max time.Duration) time.Duration {
+	raw := os.Getenv(key)
+	if raw == "" {
+		return def
 	}
-	return def
+	d, err := time.ParseDuration(raw)
+	if err != nil {
+		log.Warn("ignoring invalid duration env var; using default", "var", key, "value", raw, "default", def)
+		return def
+	}
+	if d < min {
+		log.Warn("duration env var below minimum; clamping", "var", key, "value", d.String(), "min", min.String())
+		return min
+	}
+	if d > max {
+		log.Warn("duration env var above maximum; clamping", "var", key, "value", d.String(), "max", max.String())
+		return max
+	}
+	return d
+}
+
+// intEnv parses key as an int, clamped to [min,max]. Unset → def;
+// set-but-unparseable → def with a warn; out-of-range → clamped with a warn.
+func intEnv(log *slog.Logger, key string, def, min, max int) int {
+	raw := os.Getenv(key)
+	if raw == "" {
+		return def
+	}
+	n, err := strconv.Atoi(raw)
+	if err != nil {
+		log.Warn("ignoring invalid int env var; using default", "var", key, "value", raw, "default", def)
+		return def
+	}
+	if n < min {
+		log.Warn("int env var below minimum; clamping", "var", key, "value", n, "min", min)
+		return min
+	}
+	if n > max {
+		log.Warn("int env var above maximum; clamping", "var", key, "value", n, "max", max)
+		return max
+	}
+	return n
 }
--- a/api/worker/pagecache.go
+++ b/api/worker/pagecache.go
@@ -36,6 +36,7 @@ func Start(ctx context.Context, cfg Config) error {
 	// go into the workflow as arguments (replay-deterministic) and concurrency
 	// onto the activity struct.
 	params, concurrency := loadRefreshConfig(log)
+	params = params.withDefaults() // derive ContinueAsNewThreshold (single source) for the log + workflow arg
 	log.Info("page-cache: refresh config",
 		"interval", params.RefreshInterval, "concurrency", concurrency,
 		"activity_timeout", params.ActivityTimeout, "continue_as_new_after", params.ContinueAsNewThreshold)
@@ -165,21 +166,17 @@ func envOrDefault(key, defaultValue string) string {
 // clamping to sane bounds and warning on anything set-but-invalid/out-of-range so
 // a typo (e.g. PAGE_CACHE_REFRESH_INTERVAL=90 with no unit) is visible rather than
 // silently falling back to prod defaults. Returns the workflow params (interval,
-// activity timeout, derived continue-as-new threshold) and the activity-side
-// concurrency. Pure and side-effect-free (aside from logging) so it's unit-testable
-// without a Temporal connection.
+// activity timeout) and the activity-side concurrency. ContinueAsNewThreshold is
+// left zero and derived by PageCacheParams.withDefaults (the single source, so
+// the derivation isn't duplicated). Pure and side-effect-free (aside from logging)
+// so it's unit-testable without a Temporal connection.
 func loadRefreshConfig(log *slog.Logger) (PageCacheParams, int) {
 	interval := durationEnv(log, "PAGE_CACHE_REFRESH_INTERVAL", defaultRefreshInterval, minRefreshInterval, maxRefreshInterval)
 	concurrency := intEnv(log, "PAGE_CACHE_REFRESH_CONCURRENCY", defaultRefreshConcurrency, minRefreshConcurrency, maxRefreshConcurrency)
 	timeout := durationEnv(log, "PAGE_CACHE_REFRESH_TIMEOUT", defaultActivityTimeout, minActivityTimeout, maxActivityTimeout)
-
-	// Continue-as-new after roughly continueAsNewTargetWindow of wall-clock so a
-	// longer interval can't inflate per-run history toward Temporal's limits.
-	threshold := max(int(continueAsNewTargetWindow/interval), 1)
 	return PageCacheParams{
-		RefreshInterval:        interval,
-		ActivityTimeout:        timeout,
-		ContinueAsNewThreshold: threshold,
+		RefreshInterval: interval,
+		ActivityTimeout: timeout,
 	}, concurrency
 }
 
--- a/api/worker/pagecache.go
+++ b/api/worker/pagecache.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"strconv"
 	"strings"
+	"time"
 
 	temporalclient "go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/worker"
@@ -26,6 +28,18 @@ func Start(ctx context.Context, cfg Config) error {
 	if log == nil {
 		log = slog.Default()
 	}
+
+	// Load-shaping knobs (see workflow.go), overridable per-environment. Staging
+	// runs gentler values (lower concurrency, longer interval) to spread page-cache
+	// load across its smaller self-hosted ClickHouse rather than scaling it up.
+	// Read once here, before the worker/workflow start, so both are in effect.
+	if d := envDuration("PAGE_CACHE_REFRESH_INTERVAL", 0); d > 0 {
+		refreshInterval = d
+	}
+	if n := envInt("PAGE_CACHE_REFRESH_CONCURRENCY", 0); n > 0 {
+		pageCacheRefreshConcurrency = n
+	}
+	log.Info("page-cache: refresh config", "interval", refreshInterval, "concurrency", pageCacheRefreshConcurrency)
 
 	// Connect to Temporal
 	temporalHost := envOrDefault("TEMPORAL_HOST_PORT", "localhost:7233")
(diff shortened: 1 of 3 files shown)


api/worker/workflow.go
```diff
@@ -164,13 +163,26 @@ type refreshError struct{ msg string }
...
+// minutes); fully unbounded (~28 entries) oversubscribed ClickHouse (~55
+// concurrent queries pegged the node, timeouts + per-entry retries amplified the
+// storm). A bounded limit keeps in-flight queries near what ClickHouse can run
+// while still refreshing the batch within the cycle.
+//
+// These are read once at startup (Start) and then treated as constant for the
+// life of the worker, so the workflow's use of refreshInterval stays
+// deterministic across replay/continue-as-new within a deployment.
+var (
+	refreshInterval             = 30 * time.Second
```
ben-dz: refreshInterval is now a mutable package var read inside workflow code (workflow.go:363). During a rolling deploy that changes PAGE_CACHE_REFRESH_INTERVAL, the old pod's worker still polls the same task queue (see the comment at pagecache.go:80-81) and can make decisions for the new run using the old value; the fast-refresh loop's command count depends on the interval, so when the new worker later full-replays with the new value the command sequence diverges from history → nondeterminism error → workflow task retries fail forever → page cache silently frozen until the next deploy. Fix: pass the interval as a workflow argument (ExecuteWorkflow(..., PageCacheWorkflow, 0, interval) and thread through NewContinueAsNewError) so it's recorded in history and replay-deterministic by construction. Pin with worker.WorkflowReplayer: record history with interval A, replay with interval B. (pageCacheRefreshConcurrency can stay a var — it's activity-side only.)

api/worker/workflow.go
```diff
@@ -164,13 +163,26 @@ type refreshError struct{ msg string }
...
+	pageCacheRefreshConcurrency = 8
```
ben-dz: Lowering this knob interacts badly with the fixed 3-minute StartToCloseTimeout (workflow.go:336, MaximumAttempts:1). Batch wall time ≈ ceil(28/concurrency) waves × per-entry cost (up to 120s with the internal retry), so at concurrency 2–4 on staging's slow ClickHouse the budget can expire mid-batch every cycle. errgroup launches entries in fixed list order, so the same tail entries — geo_concentration (entry 21 of 28) among them — never refresh, and the deadline cancellation is logged as "cache refresh interrupted (shutdown)" (workflow.go:250, :263) without incrementing the failure counter: silent permanent staleness of exactly the entries this PR targets. Fix: derive/configure the activity timeout alongside these knobs, and distinguish deadline-exceeded from shutdown in the interrupted-refresh log branch so starvation is visible and counted.

api/worker/pagecache.go
```diff
@@ -146,3 +160,24 @@ func envOrDefault(key, defaultValue string) string {
 	}
 	return defaultValue
 }
+
+// envDuration returns the env var parsed as a Go duration (e.g. "90s"), or def
+// if unset/invalid.
+func envDuration(key string, def time.Duration) time.Duration {
```
ben-dz: Set-but-invalid values fall back silently: PAGE_CACHE_REFRESH_INTERVAL=90 (missing unit) fails ParseDuration and staging silently runs prod defaults — reintroducing the exact overload the deploy was meant to fix. Same for a typo'd int or a set-but-non-positive value dropped by the > 0 guards. The startup log shows effective values but nothing says "your var was ignored". Fix: log.Warn with the raw value when the var is set but unparseable/non-positive. Test: setenv "90" → assert warn + default retained.

api/worker/pagecache.go
```diff
@@ -27,6 +29,18 @@ func Start(ctx context.Context, cfg Config) error {
 		log = slog.Default()
 	}
 
+	// Load-shaping knobs (see workflow.go), overridable per-environment. Staging
+	// runs gentler values (lower concurrency, longer interval) to spread page-cache
+	// load across its smaller self-hosted ClickHouse rather than scaling it up.
+	// Read once here, before the worker/workflow start, so both are in effect.
+	if d := envDuration("PAGE_CACHE_REFRESH_INTERVAL", 0); d > 0 {
```
ben-dz: No sanity bounds: PAGE_CACHE_REFRESH_INTERVAL=1ms produces a hot back-to-back RefreshCaches loop (the deadline window never admits a fast tick or sleep), and a huge PAGE_CACHE_REFRESH_CONCURRENCY is effectively unbounded — the comment history at workflow.go:173-177 documents that unbounded pegged ClickHouse. Operator-controlled so low severity, but consider clamping to a sane range with a warn.

api/worker/pagecache_test.go
```diff
@@ -0,0 +1,28 @@
+package worker
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestEnvDuration(t *testing.T) {
```
ben-dz: These tests cover the parsers only; the behavior that matters — the > 0 guards, override precedence, "unset keeps prod defaults" — lives inline in Start and is untestable without a Temporal connection. Consider extracting a loadRefreshConfig(log) (time.Duration, int) helper and testing it directly; that's also the natural home for the warn-on-invalid-value log.

ben-dz (changes requested):
0 critical, 2 high, 2 medium, 2 low. The env plumbing itself is sound (no data race — writes in Start happen-before all reading goroutines; prod defaults unchanged when unset; both entry points covered; -race tests and lint pass; no security findings). Two high findings: (1) refreshInterval is read from a mutable package var inside Temporal workflow code, so a deploy that changes PAGE_CACHE_REFRESH_INTERVAL can hit replay nondeterminism during the rolling-deploy worker overlap, freezing the page-cache workflow until the next deploy — pass the interval as a workflow argument instead; (2) lowering concurrency interacts with the fixed 3-minute activity StartToCloseTimeout and fixed entry launch order to deterministically starve tail entries (including geo_concentration, the motivating case), and the deadline cancellation is mislogged as shutdown so it never escalates. Also: invalid env values fall back silently (warn wanted), and continueAsNewThreshold=60 doesn't scale with the interval (larger intervals inflate per-run history toward Temporal limits).

- medium: continueAsNewThreshold (workflow.go:22) is fixed at 60 iterations, calibrated for the 30s interval ("~30 min at 30s intervals" — comment now stale). Longer intervals multiply fast-refresh ticks per iteration (~8 history events per tick): at 120s → ~19k events per run, past Temporal's ~10k suggested continue-as-new threshold; intervals beyond ~4–5 min approach the 51.2k hard history limit where the server terminates the workflow. Compute the threshold from the interval (target fixed wall time or tick budget), ideally passed alongside the interval workflow argument; at minimum update the comment.
- Security review: no findings. Data race between the Start writes and Temporal goroutine reads refuted (happens-before via goroutine creation; Start has exactly two callers, neither concurrent nor repeated; vars unexported and read nowhere outside api/worker). Extreme env values are operator-controlled self-DoS only; zero/negative rejected by the > 0 guards; the silent-fallback parsers gate no auth/authz decision.
- Checked and clean: prod defaults byte-identical when env unset; both entry points (api/main.go:379, api/worker/cmd/main.go:56) route through Start so overrides apply uniformly; concurrency knob is activity-side only (no replay concern); t.Setenv used correctly. go build, go vet, go test -race ./api/worker/, and golangci-lint all pass.

nikw9944:
Thanks for the thorough review — addressed all six findings in `c9214e1` (also rebased onto `main`).

**HIGH — replay nondeterminism (interval as mutable package var):** Fixed. Interval, activity timeout, and continue-as-new threshold are now carried in a `PageCacheParams` struct passed as a **workflow argument** — threaded through `ExecuteWorkflow(..., 0, params)` and `NewContinueAsNewError(ctx, PageCacheWorkflow, 0, p)` — so they're recorded in history and replay-deterministic by construction. A config change now starts a fresh run (Start already terminates + restarts) rather than mutating state an in-flight replay would read. Concurrency stays a plain `Activities` field (activity-side, no replay concern), as you noted. (I did not add a `WorkflowReplayer` test — it needs recorded history and I don't have Temporal infra in this environment; happy to add one against a captured history if you'd like. Determinism is now structural rather than test-enforced.)

**HIGH — concurrency vs fixed timeout starves tail entries + mislogged as shutdown:** Fixed both halves. (1) The activity `StartToCloseTimeout` is now configurable (`PAGE_CACHE_REFRESH_TIMEOUT`, passed via params) so operators can size it alongside the concurrency knob. (2) `refresh()` now distinguishes a genuine worker shutdown (via `activity.GetWorkerStopChannel`) from an activity-deadline cancellation: the deadline case is **counted** as a failure (`errBatchDeadline`, deliberately phrased to miss `dberror`'s transient keywords so it escalates at `errorAfterFailures`) instead of being swallowed as `"interrupted (shutdown)"`. So starvation of tail entries (geo_concentration included) now surfaces at WARN and escalates to ERROR if persistent.

**MEDIUM — continueAsNewThreshold fixed at 60:** Fixed. Now derived from the interval to target ~`continueAsNewTargetWindow` (30 min) of wall-clock (e.g. 90s → 20 iterations), keeping per-run history bounded regardless of interval. Stale comment updated.

**MEDIUM — invalid env silently falls back:** Fixed. `durationEnv`/`intEnv` now `log.Warn` on set-but-unparseable (and clamp-with-warn on out-of-range). Test covers `PAGE_CACHE_REFRESH_INTERVAL=90` (no unit) → warn + default retained.

**LOW — no sanity bounds:** Fixed. Clamped with warn: interval 5s–10m, concurrency 1–32, timeout 30s–10m.

**LOW — extract + test the config logic:** Done. `loadRefreshConfig(log)` extracted (pure, no Temporal dep); unit tests cover unset/valid/invalid/clamp/precedence and the threshold derivation.

Verified: `go build ./api/...`, `go vet`, `go test -race ./api/worker/`, `gofmt` all clean.

api/worker/workflow.go
nikw9944: Fixed (c9214e1): interval + activity timeout + continue-as-new threshold now travel as a `PageCacheParams` workflow argument (threaded through `ExecuteWorkflow` and `NewContinueAsNewError`), so they're recorded in history and replay-deterministic. Concurrency stays activity-side.

api/worker/workflow.go
nikw9944: Fixed: activity `StartToCloseTimeout` is now configurable (`PAGE_CACHE_REFRESH_TIMEOUT`), and `refresh()` distinguishes worker shutdown (`activity.GetWorkerStopChannel`) from a deadline cancellation — the latter is counted as `errBatchDeadline` (non-transient → escalates at `errorAfterFailures`), so tail-entry starvation is now visible/counted instead of logged as shutdown.

api/worker/pagecache.go
nikw9944: Fixed: `durationEnv`/`intEnv` now `log.Warn` on set-but-unparseable (and clamp-with-warn on out-of-range). Test covers `PAGE_CACHE_REFRESH_INTERVAL=90` (missing unit) → warn + default retained.

api/worker/pagecache.go
nikw9944: Fixed: clamped to sane bounds with a warn — interval 5s–10m, concurrency 1–32, timeout 30s–10m.

api/worker/pagecache_test.go
nikw9944: Done: extracted `loadRefreshConfig(log)` (pure, no Temporal dep); unit tests now cover the >0/clamp guards, override precedence, unset-keeps-defaults, and the continue-as-new threshold derivation.

api/worker/pagecache.go
```diff
@@ -56,7 +70,7 @@ func Start(ctx context.Context, cfg Config) error {
 	run, err := tc.ExecuteWorkflow(ctx, temporalclient.StartWorkflowOptions{
 		ID:        WorkflowID,
 		TaskQueue: TaskQueue,
-	}, PageCacheWorkflow, 0)
+	}, PageCacheWorkflow, 0, params)
```
ben-dz, line 74: Rollout constraint (no code change needed): verified in SDK v1.41.1 that converter.FromPayloads silently drops extra payloads, so during the migration deploy's overlap window the old pre-PR worker (2-arg PageCacheWorkflow) can pick up this new 3-arg run, ignore the params payload, and execute it with its compiled 30s constants. Harmless if the env vars are unset at that deploy (old and new code produce command-identical histories). But if PAGE_CACHE_REFRESH_INTERVAL ships with a non-default value in the same deploy, the old worker records 30s-shaped history for a run whose input says otherwise, and full replay on the new worker diverges — the original replay wedge, one time, until the next deploy. Please state the two-phase rollout in the PR description: merge/deploy with env unset, then set staging's values in a follow-up deploy. (The reverse direction is safe: zero-params histories normalize via withDefaults.)

api/worker/workflow.go
```diff
@@ -217,25 +301,27 @@ func (a *Activities) latestEntries() []cacheEntry {
 
 // RefreshLatestCaches refreshes just the fast-cadence entries (latest slots slice).
 func (a *Activities) RefreshLatestCaches(ctx context.Context) error {
+	shuttingDown := workerStopping(ctx)
```
ben-dz, line 313: New regression: RefreshLatestCaches entries cut by the fast activity's 60s StartToCloseTimeout now flow into interrupted() → errBatchDeadline, which is non-transient by design and escalates to ERROR at errorAfterFailures=3. But the comment at workflow.go:439-443 says these blowouts occur under ordinary ClickHouse contention and self-heal — exactly the class routed to transientErrorAfterFailures=10 to avoid paging on blips. ~3 minutes of contention on staging's slow ClickHouse now produces ERROR logs on this path (which wasn't counted at all before). The strict threshold is right for the slow batch (deterministic tail starvation — the point of the fix) but wrong for the fast loop. Fix: use a distinct sentinel or flag for the fast path so its parent-deadline expiry counts at the transient threshold. Test: drive interrupted() for a fast-entry key 3× and assert WARN, not ERROR, below 10.

api/worker/pagecache_test.go
```diff
@@ -0,0 +1,126 @@
...
+		require.Equal(t, 1, warnCount(*recs))
+	})
+}
+
+func TestLoadRefreshConfig(t *testing.T) {
+	log, _ := capLogger()
+
+	t.Run("unset → prod defaults, threshold derived for 30s", func(t *testing.T) {
+		p, conc := loadRefreshConfig(log)
+		require.Equal(t, defaultRefreshInterval, p.RefreshInterval)
```
ben-dz, line 120: TestLoadRefreshConfig is not hermetic — verified failing: PAGE_CACHE_REFRESH_INTERVAL=45s go test -run TestLoadRefreshConfig ./api/worker/ fails this "unset → prod defaults" subtest, and PAGE_CACHE_REFRESH_TIMEOUT is also read from the ambient env. Fix: t.Setenv(key, "") for all three vars in each subtest (the code treats empty as unset).

api/worker/pagecache.go
```diff
@@ -146,3 +160,72 @@ func envOrDefault(key, defaultValue string) string {
...
+// concurrency. Pure and side-effect-free (aside from logging) so it's unit-testable
+// without a Temporal connection.
+func loadRefreshConfig(log *slog.Logger) (PageCacheParams, int) {
+	interval := durationEnv(log, "PAGE_CACHE_REFRESH_INTERVAL", defaultRefreshInterval, minRefreshInterval, maxRefreshInterval)
+	concurrency := intEnv(log, "PAGE_CACHE_REFRESH_CONCURRENCY", defaultRefreshConcurrency, minRefreshConcurrency, maxRefreshConcurrency)
+	timeout := durationEnv(log, "PAGE_CACHE_REFRESH_TIMEOUT", defaultActivityTimeout, minActivityTimeout, maxActivityTimeout)
+
+	// Continue-as-new after roughly continueAsNewTargetWindow of wall-clock so a
+	// longer interval can't inflate per-run history toward Temporal's limits.
+	threshold := max(int(continueAsNewTargetWindow/interval), 1)
```
ben-dz: This max(continueAsNewTargetWindow/interval, 1) derivation is duplicated at workflow.go:86 (withDefaults) — drift risk. Consider having loadRefreshConfig leave ContinueAsNewThreshold zero and letting withDefaults be the single source, plus a unit test for PageCacheParams{}.withDefaults() — the zero-params compat path that protects pre-change histories is currently untested.

api/worker/workflow.go
```diff
@@ -316,12 +389,47 @@ func (a *Activities) incFailures(key string) int {
...
+// otherwise the RefreshCaches activity ran out of its StartToCloseTimeout budget
+// and this entry is being starved — count it (as errBatchDeadline) so persistent
+// starvation of tail entries surfaces and escalates rather than hiding as
+// "shutdown". cause is the underlying fn error if any (nil if the entry never ran).
+func (a *Activities) interrupted(name, key string, cause error, shuttingDown func() bool) {
+	if shuttingDown == nil || shuttingDown() {
+		a.Log.Warn("cache refresh interrupted (shutdown)", "cache", name, "error", cause)
+		return
+	}
+	a.recordFailure(name, key, errBatchDeadline)
```
ben-dz: The counted branch drops `cause`: recordFailure gets bare errBatchDeadline, so the escalating log line loses which query was cut and why (cause is only logged in the shutdown branch). Attach cause as a log attribute — log-only, don't wrap it into the error, or a cause wrapping context.DeadlineExceeded would subtly interact with the transient classification.

ben-dz (approved):
All six original findings are verified fixed in c9214e1 — the PageCacheParams workflow-argument refactor is structurally replay-deterministic (verified: no mutable-state reads remain in workflow code; zero-value params from pre-change histories normalize deterministically), the shutdown-vs-deadline classification is correct on every cancellation path (verified against Temporal SDK v1.41.1 source), errBatchDeadline dodges dberror's transient patterns so starvation escalates at 3, the continue-as-new threshold now targets ~30min wall-clock so per-run history stays constant at any interval, and env parsing warns/clamps with extracted, tested config logic. Build, vet, -race tests, and lint all pass. Remaining: 2 medium, 3 low, none blocking as code — (1) a rollout constraint: ship this code with the env vars unset and set staging's values in a follow-up deploy, else the old 2-arg worker executing the new 3-arg run during the migration overlap (extra payloads are silently dropped) can recreate the replay wedge one time; document this on the PR; (2) a small regression: fast-refresh entries cut by their 60s activity budget now count as errBatchDeadline at the strict threshold (ERROR after 3), though the code's own comment says these blowouts are ordinary self-healing contention that belongs at the transient threshold (10). Plus three optional cleanups (non-hermetic TestLoadRefreshConfig, duplicated threshold derivation, dropped cause in the counted interrupted branch).

- Original findings, all verified fixed at c9214e1: High #1 replay nondeterminism → PageCacheParams passed as workflow argument through ExecuteWorkflow (pagecache.go:70-73) and NewContinueAsNewError (workflow.go:477), determinism now structural (skipping the WorkflowReplayer test is acceptable); High #2 tail starvation → PAGE_CACHE_REFRESH_TIMEOUT knob (workflow.go:432) plus workerStopping/errBatchDeadline counting, with errBatchDeadline verified to match no dberror transient pattern so it escalates at 3; Medium #3 → warn on invalid env + clamp-with-warn, tested including "90" without a unit; Medium #4 → threshold = 30min/interval keeps per-run history ~constant at any interval; Low #5 → bounds 5s–10m / 1–32 / 30s–10m; Low #6 → loadRefreshConfig extracted and unit-tested.
- Checked and clean: workflow code reads only params and compile-time consts; withDefaults is a pure function of recorded input; worker stop channel verified (SDK v1.41.1) to close before ctx cancellation so shutdown classification can't race false; workflow termination doesn't cancel a running non-heartbeating activity, so the deadline/shutdown branches cover all parent-cancel sources; concurrency correctly stays activity-side with a 0-fallback; go build, go vet, go test -race ./api/worker/, and golangci-lint all pass at c9214e1.

api/worker/workflow.go
nikw9944: Fixed (95af49e): the fast loop now records a distinct `errFastRefreshDeadline` sentinel (phrased as a timeout → classified transient → escalates at `transientErrorAfterFailures`), while the slow batch keeps `errBatchDeadline` (strict). Test `TestInterruptedEscalation` drives both: fast key ×4 stays WARN / 0 ERROR; slow key escalates to ERROR at the strict threshold.

api/worker/pagecache_test.go
nikw9944: Fixed: added `clearRefreshEnv(t)` (`t.Setenv(..., "")` for all three vars) at the top of each subtest, and loadRefreshConfig no longer derives the threshold (so no ambient-env dependency there). Verified: `PAGE_CACHE_REFRESH_INTERVAL=45s go test -run 'TestLoadRefreshConfig|TestPageCacheParamsWithDefaults' ./api/worker/` now passes.

api/worker/pagecache.go
nikw9944: Fixed: `loadRefreshConfig` now leaves `ContinueAsNewThreshold` zero; `withDefaults` is the single source of the `max(target/interval, 1)` derivation (Start calls `withDefaults` for the log/arg, the workflow calls it again idempotently). Added `TestPageCacheParamsWithDefaults` covering the zero-params compat path, the interval→threshold scaling, and idempotency.

api/worker/workflow.go
nikw9944: Fixed: `recordFailure` now takes variadic extra attrs; `interrupted` passes `"cause", cause` so the escalating line keeps which query was cut, while the classification error stays the bare sentinel (no wrapping of `context.DeadlineExceeded` into it, so transient classification is unaffected).

api/worker/pagecache.go
nikw9944: Documented: added a two-phase rollout section to the PR description — deploy this PR with the env vars unset first (command-identical histories, zero-params path normalizes via `withDefaults`), then set staging's values (infra#1905) once every worker is on the 3-arg code.

PR #689, opened 2026-07-09
multicast-health: drop rate reconciliation, add path root-cause rollup

ben-malbeclabs:
## Summary

- **Drop rate reconciliation from the multicast health verdict.** A user's `device_interface_rollup_5m` counter is per-tunnel and aggregates every group that user publishes/subscribes to, so a per-group "expected" rate (sum of the group's publishers' RX) can't be attributed once a user is in more than one group — which is the common case. In practice 100% of the "degraded" verdicts this check produced were multi-group aggregation artifacts, not delivery problems. The rate dimension is now **presence-only** (`active` / `idle` / `no_data`) and never gates `health_status`. Health is the control-plane verdict, which already includes `disconnected` for BGP-down users.
- **Keep observed bps + link to the traffic charts.** Each rate cell still shows the observed 5-min rate and now links to the tunnel's Traffic History on the user page, so operators can eyeball the actual rates without the (impossible) per-group reconciliation.
- **Add a "Root causes" rollup for the per-path fan-out.** One broken publisher drags down every one of its subscriber pairs (and vice-versa), so a handful of faulty endpoints can produce hundreds of unhealthy `(publisher, subscriber)` rows. A new `GET /api/dz/multicast-groups/{pk}/health/path-root-causes` endpoint collapses the unhealthy fan-out to the faulting endpoint(s) with an affected-pair count, rendered as a "Root causes" panel above the per-path table.

## Why

The group Health tab flagged large numbers of publishers/paths as unhealthy or degraded for reasons that weren't real delivery faults — degraded from an un-attributable rate mismatch, and a raw per-path table where 3 bad endpoints read as 450 bad rows. This removes the noise source and replaces the fan-out with an actionable root-cause view.

## Changes

- New migration `20260709000003_health_rate_presence_only.sql` recreates `health_multicast_user_rate`: drops `group_publisher_total` and the sum-reconciliation, verdict = control-plane `health_status`, rate reduced to a presence signal. `expected_bps_5m` retained as `NULL` for output-schema stability.
- New `path-root-causes` endpoint + query attributing each non-healthy path to the endpoint(s) not observed (a BGP-down endpoint has no session → not observed, so one filter catches both `unhealthy` and `disconnected` attributions).
- Web: rate badges → `active`/`idle`/`unknown`; removed expected/mismatch rendering; "traffic" deep-link to the user Traffic History section; `RootCausePanel` in the group Health tab.

## Testing Verification

- `TestHealthMulticastUserRate*` (mroute) rewritten to presence-only semantics — multi-group subscriber that previously flagged `degraded` now stays `active`/`healthy`; observed-rate dedup preserved; disconnected propagation still asserted.
- New `TestGetMulticastGroupHealth_PathRootCauses`: two unobserved subscribers of the same group surface as distinct root causes — one `unhealthy` (BGP up), one `disconnected` (BGP down) — each `affects 1 pair`; the observed publisher and the healthy pair contribute no root cause.
- `TestGetMulticastGroupHealth_Users` updated for presence-only (both endpoints transmitting → `active`, `expected_bps_5m` nil).
- Full `-run Multicast` API handler suite green.

## Related

Follows #685 (ingest onchain `bgp_status`) and #686 (reclassify BGP-down as `disconnected`).

### DIFF
--- a/api/handlers/multicast_health_paths.go
+++ b/api/handlers/multicast_health_paths.go
@@ -99,6 +99,11 @@ func (a *API) GetMulticastGroupHealthPaths(w http.ResponseWriter, r *http.Reques
 	})
 }
 
+// multicastHealthRootCauseLimit bounds the returned root causes; after a
+// whole-group event every endpoint can be a fault, so the panel shows the
+// top-N by blast radius while Total still reports the true distinct count.
+const multicastHealthRootCauseLimit = 100
+
 // GetMulticastGroupHealthPathRootCauses returns the unhealthy per-path fan-out
 // collapsed to the faulting endpoint (publisher or subscriber), with a count of
 // how many (publisher, subscriber) pairs each one drags down. This is the
@@ -123,7 +128,7 @@ func (a *API) GetMulticastGroupHealthPathRootCauses(w http.ResponseWriter, r *ht
 		return
 	}
 
-	items, err := a.queryMulticastHealthPathRootCauses(ctx, group.PK)
+	items, total, err := a.queryMulticastHealthPathRootCauses(ctx, group.PK)
 	if err != nil {
 		logError("multicast group health/path-root-causes query error", "error", err, "pk", group.PK)
 		http.Error(w, dberror.UserMessage(err), http.StatusInternalServerError)
@@ -134,81 +139,106 @@ func (a *API) GetMulticastGroupHealthPathRootCauses(w http.ResponseWriter, r *ht
 		Group:       group,
 		GeneratedAt: formatMulticastTime(time.Now().UTC()),
 		Items:       items,
-		Total:       len(items),
+		Total:       total,
 	})
 }
 
 // queryMulticastHealthPathRootCauses attributes every non-healthy path to a
-// single primary faulting endpoint. A BGP-down endpoint has no session, so no
-// (S,G)/OIF, so its *_endpoint_observed is false — meaning the not-observed
-// filter catches both 'unhealthy' (endpoint present but not forwarding) and
-// 'disconnected' (BGP down). endpoint_status reports the endpoint's own
-// condition (disconnected if its BGP is down, else unhealthy).
+// single primary faulting endpoint and rolls the fan-out up per endpoint.
 //
-// The subscriber branch additionally requires publisher_endpoint_observed = 1:
-// when a publisher is itself unobserved (e.g. BGP down) there is no (S,G) for
-// that source anywhere, so every one of its subscribers is trivially
-// unobserved too. Attributing those subscribers would re-create the exact
-// fan-out this rollup exists to collapse (1 down publisher → N false
-// subscriber root causes). The fault is upstream, so the pair is attributed
-// to the publisher alone.
-func (a *API) queryMulticastHealthPathRootCauses(ctx context.Context, groupPK string) ([]MulticastHealthPathRootCause, error) {
+// A single scan of the (expensive, non-materialized) path view feeds an
+// ARRAY JOIN over two candidate endpoint tuples per pair, each gated by the
+// observed flags — replacing a UNION ALL that scanned the view twice.
+// Candidates:
+//   - publisher, when publisher_endpoint_observed = 0
+//   - subscriber, when subscriber_endpoint_observed = 0 AND publisher is
+//     observed
+//
+// The publisher gate on the subscriber candidate makes each pair attribute to
+// a single primary endpoint: when a publisher is itself unobserved (e.g. BGP
+// down) there is no (S,G) for that source, so every one of its subscribers is
+// trivially unobserved too — the fault is upstream and the pair belongs to the
+// publisher alone. Without this a down publisher with N subscribers would
+// re-create the exact fan-out this rollup exists to collapse.
+//
+// A BGP-down endpoint has no session, so no (S,G)/OIF, so its *_endpoint_observed
+// is false — the not-observed gate therefore catches both 'unhealthy' (present
+// but not forwarding) and 'disconnected' (BGP down). endpoint_status reports the
+// endpoint's own condition via dz_users_current (the path view only exposes the
+// pair-level verdict, not which side is down). Grouping is by endpoint alone so
+// a P+S user broken on both sides is one row with a combined role label.
+//
+// Returns the top-N rows by blast radius plus the true distinct-endpoint total.
+func (a *API) queryMulticastHealthPathRootCauses(ctx context.Context, groupPK string) ([]MulticastHealthPathRootCause, int, error) {
 	query := `
 		WITH faults AS (
 			SELECT
-				'publisher' AS faulting_role,
-				publisher_user_pk AS endpoint_user_pk,
-				publisher_owner_pubkey AS endpoint_owner_pubkey,
-				publisher_dz_ip AS endpoint_dz_ip,
-				publisher_tunnel_id AS endpoint_tunnel_id,
-				publisher_device_pk AS endpoint_device_pk,
-				publisher_device_code AS endpoint_device_code
+				c.1 AS faulting_role,
+				c.2 AS endpoint_user_pk,
+				c.3 AS endpoint_owner_pubkey,
+				c.4 AS endpoint_dz_ip,
+				c.5 AS endpoint_tunnel_id,
+				c.6 AS endpoint_device_pk,
+				c.7 AS endpoint_device_code
 			FROM health_publisher_subscriber_path
+			ARRAY JOIN [
+				if(publisher_endpoint_observed = 0,
+					('publisher', publisher_user_pk, publisher_owner_pubkey, publisher_dz_ip,
+					 CAST(publisher_tunnel_id AS Int32), publisher_device_pk, publisher_device_code),
+					('', '', '', '', CAST(0 AS Int32), '', '')),
+				if(subscriber_endpoint_observed = 0 AND publisher_endpoint_observed = 1,
+					('subscriber', subscriber_user_pk, subscriber_owner_pubkey, subscriber_dz_ip,
+					 CAST(subscriber_tunnel_id AS Int32), subscriber_device_pk, subscriber_device_code),
+					('', '', '', '', CAST(0 AS Int32), '', ''))
+			] AS c
 			WHERE multicast_group_pk = ?
 			  AND health_status != 'healthy'
-			  AND publisher_endpoint_observed = 0
-			UNION ALL
+		),
+		grouped AS (
 			SELECT
-				'subscriber' AS faulting_role,
-				subscriber_user_pk AS endpoint_user_pk,
-				subscriber_owner_pubkey AS endpoint_owner_pubkey,
-				subscriber_dz_ip AS endpoint_dz_ip,
-				subscriber_tunnel_id AS endpoint_tunnel_id,
-				subscriber_device_pk AS endpoint_device_pk,
-				subscriber_device_code AS endpoint_device_code
-			FROM health_publisher_subscriber_path
-			WHERE multicast_group_pk = ?
-			  AND health_status != 'healthy'
-			  AND subscriber_endpoint_observed = 0
-			  AND publisher_endpoint_observed = 1
+				arrayStringConcat(arraySort(groupUniqArray(f.faulting_role)), '+') AS faulting_role,
+				f.endpoint_user_pk AS endpoint_user_pk,
+				any(f.endpoint_owner_pubkey) AS endpoint_owner_pubkey,
+				any(f.endpoint_dz_ip) AS endpoint_dz_ip,
+				any(f.endpoint_tunnel_id) AS endpoint_tunnel_id,
+				any(f.endpoint_device_pk) AS endpoint_device_pk,
+				any(f.endpoint_device_code) AS endpoint_device_code,
+				if(any(u.bgp_status) = 'down', 'disconnected', 'unhealthy') AS endpoint_status,
+				toInt32(count()) AS affected_pairs
+			FROM faults f
+			LEFT ANY JOIN dz_users_current u ON f.endpoint_user_pk = u.pk
+			WHERE f.faulting_role != ''
+			GROUP BY f.endpoint_user_pk
 		)
 		SELECT
-			f.faulting_role,
-			f.endpoint_user_pk,
-			any(f.endpoint_owner_pubkey),
-			any(f.endpoint_dz_ip),
-			any(f.endpoint_tunnel_id),
-			any(f.endpoint_device_pk),
-			any(f.endpoint_device_code),
-			if(any(u.bgp_status) = 'down', 'disconnected', 'unhealthy') AS endpoint_status,
-			toInt32(count()) AS affected_pairs
-		FROM faults f
-		LEFT ANY JOIN dz_users_current u ON f.endpoint_user_pk = u.pk
-		GROUP BY f.faulting_role, f.endpoint_user_pk
-		ORDER BY affected_pairs DESC, endpoint_status, f.endpoint_user_pk
+			faulting_role,
+			endpoint_user_pk,
+			endpoint_owner_pubkey,
+			endpoint_dz_ip,
+			endpoint_tunnel_id,
+			endpoint_device_pk,
+			endpoint_device_code,
+			endpoint_status,
+			affected_pairs,
+			toInt32(count() OVER ()) AS total_endpoints
+		FROM grouped
+		ORDER BY affected_pairs DESC, endpoint_status, endpoint_user_pk
+		LIMIT ?
 		SETTINGS max_execution_time = 30, timeout_before_checking_execution_speed = 0
 	`
 	start := time.Now()
-	rows, err := a.envDB(ctx).Query(ctx, query, groupPK, groupPK)
+	rows, err := a.envDB(ctx).Query(ctx, query, groupPK, multicastHealthRootCauseLimit)
 	metrics.RecordClickHouseQuery("multicast_health_path_root_causes", time.Since(start), err)
 	if err != nil {
-		return nil, err
+		return nil, 0, err
 	}
 	defer rows.Close()
 
 	items := []MulticastHealthPathRootCause{}
+	total := 0
 	for rows.Next() {
 		var it MulticastHealthPathRootCause
+		var rowTotal int32
 		if err := rows.Scan(
 			&it.FaultingRole,
 			&it.UserPK,
@@ -219,15 +249,17 @@ func (a *API) queryMulticastHealthPathRootCauses(ctx context.Context, groupPK st
 			&it.DeviceCode,
 			&it.EndpointStatus,
 			&it.AffectedPairs,
+			&rowTotal,
 		); err != nil {
-			return nil, err
+			return nil, 0, err
 		}
+		total = int(rowTotal)
 		items = append(items, it)
 	}
 	if err := rows.Err(); err != nil {
-		return nil, err
+		return nil, 0, err
 	}
-	return items, nil
+	return items, total, nil
 }
 
 // readMulticastHealthPathsCache returns the cached first-page paths response
--- a/api/handlers/multicast_health_paths.go
+++ b/api/handlers/multicast_health_paths.go
@@ -99,6 +99,128 @@ func (a *API) GetMulticastGroupHealthPaths(w http.ResponseWriter, r *http.Reques
 	})
 }
 
+// GetMulticastGroupHealthPathRootCauses returns the unhealthy per-path fan-out
+// collapsed to the faulting endpoint (publisher or subscriber), with a count of
+// how many (publisher, subscriber) pairs each one drags down. This is the
+// actionable summary behind the raw per-path table.
+func (a *API) GetMulticastGroupHealthPathRootCauses(w http.ResponseWriter, r *http.Request) {
+	ctx, cancel := context.WithTimeout(r.Context(), 30*time.Second)
+	defer cancel()
+
+	pkOrCode := chi.URLParam(r, "pk")
+	if pkOrCode == "" {
+		http.Error(w, "missing multicast group pk", http.StatusBadRequest)
+		return
+	}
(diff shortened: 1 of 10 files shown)


api/handlers/multicast_health_paths.go
```diff
@@ -99,6 +99,128 @@ func (a *API) GetMulticastGroupHealthPaths(w http.ResponseWriter, r *http.Reques
...
+				subscriber_user_pk AS endpoint_user_pk,
+				subscriber_owner_pubkey AS endpoint_owner_pubkey,
+				subscriber_dz_ip AS endpoint_dz_ip,
+				subscriber_tunnel_id AS endpoint_tunnel_id,
+				subscriber_device_pk AS endpoint_device_pk,
+				subscriber_device_code AS endpoint_device_code
+			FROM health_publisher_subscriber_path
+			WHERE multicast_group_pk = ?
+			  AND health_status != 'healthy'
+			  AND subscriber_endpoint_observed = 0
```
juan-malbeclabs: **A BGP-down publisher will surface a false `unhealthy` root cause for every one of its (healthy) subscribers.**

When a publisher's BGP is down there is no `(S,G)` for that source anywhere, so *both* endpoints of every pair are unobserved:
- `publisher_endpoint_observed = 0` → attributed as `disconnected` ✅
- `subscriber_endpoint_observed = 0` (no `(S,G)` for that source → no OIF) → attributed as `unhealthy` ❌

`health_publisher_subscriber_path` marks the whole path `disconnected` when `p_bgp_status = 'down'` and does not exclude BGP-down publishers from the pairing, so a single down publisher with N subscribers produces:
- 1 `disconnected` publisher root cause (`affects N pairs`) — correct, ranked first
- **N separate `unhealthy` subscriber root causes** (`affects 1 pair` each) — healthy subscribers mislabeled as at-fault

This re-introduces the exact fan-out this feature exists to collapse (1 down publisher → N+1 rows). Consider suppressing a subscriber attribution when the pair's publisher is itself unobserved (attribute each pair to a single primary faulting endpoint, preferring the publisher / the disconnected side).

The new `TestGetMulticastGroupHealth_PathRootCauses` only covers an observed, healthy publisher, so this path is untested — worth a fixture with a BGP-down publisher fanning out to multiple subscribers either way.

web/src/components/user-detail-page.tsx
```diff
@@ -214,6 +210,14 @@ function UserTrafficChart({ userPk }: { userPk: string }) {
   const [agg, setAgg] = useState<AggMethod>('max')
   const [hoveredIdx, setHoveredIdx] = useState<number | null>(null)
 
+  // Scroll into view when linked to with #traffic (e.g. from a multicast
+  // health tab's "traffic" link). Client-side nav doesn't honor the hash.
+  useEffect(() => {
+    if (window.location.hash === '#traffic') {
```
juan-malbeclabs, line 216: This effect reads `window.location.hash` with a `[]` dependency, so it only scrolls on initial mount. An in-app client-side nav to `/dz/users/<pk>#traffic` while this page is already mounted won't scroll. Acceptable given the link almost always triggers a fresh navigation — just noting the limitation.

juan-malbeclabs (commented):
Solid change — dropping the un-attributable rate reconciliation removes a real noise source, and the migration is careful (column order preserved, `expected_bps_5m` kept as NULL for schema stability). The presence-only test rewrites are thorough.

One substantive concern on the new root-cause query around BGP-down publishers (inline), plus two minor cleanups:

- **Dead badge keys** in `multicast-group-health-tab.tsx` `DIM_BADGE_CLASS` (~L257–270): `reconciled` / `mismatch` / `monitoring_gap` / `group_idle` are no longer reachable now that rate is presence-only — the parallel map in `user-detail-page.tsx` had these removed but this one didn't. Also `idle` maps to muted-gray here vs. amber in `RATE_STATUS_BADGE`, so the rate `idle` renders gray in this hover-card and amber in the rate cell — worth aligning.

api/handlers/multicast_health_paths.go
```diff
@@ -99,6 +99,137 @@ func (a *API) GetMulticastGroupHealthPaths(w http.ResponseWriter, r *http.Reques
...
+// subscriber root causes). The fault is upstream, so the pair is attributed
+// to the publisher alone.
+func (a *API) queryMulticastHealthPathRootCauses(ctx context.Context, groupPK string) ([]MulticastHealthPathRootCause, error) {
+	query := `
+		WITH faults AS (
+			SELECT
+				'publisher' AS faulting_role,
+				publisher_user_pk AS endpoint_user_pk,
+				publisher_owner_pubkey AS endpoint_owner_pubkey,
+				publisher_dz_ip AS endpoint_dz_ip,
```
ben-dz: **Query cost: the UNION ALL evaluates the expensive non-materialized `health_publisher_subscriber_path` view twice per request** (P×S fan-out + three mroute joins; the worker calls it multi-second on the shred group, which is why /health/paths has a page-cache entry). This endpoint is uncached and the UI refetches every 60s per open tab, enabled precisely on the hot faulty group. Collapse to a single scan (filter once on `multicast_group_pk`/`health_status != 'healthy'`, then arrayJoin over the two candidate endpoint tuples gated by the observed flags) and/or add a worker page-cache entry mirroring the paths cache. The `LEFT ANY JOIN dz_users_current` also re-derives BGP state the path view already computed.

web/src/components/multicast-group-health-tab.tsx
```diff
@@ -412,6 +413,76 @@ function TableStateRow({
...
+          {faultCount.toLocaleString()} unhealthy pairs, but no single endpoint could be attributed.
+        </div>
+      ) : (
+        <>
+          <div className="text-xs text-muted-foreground mb-2">
+            {faultCount.toLocaleString()} unhealthy publisher → subscriber pairs trace back to{' '}
+            <span className="font-medium text-foreground">{items.length}</span>{' '}
+            {items.length === 1 ? 'endpoint' : 'endpoints'}:
+          </div>
+          <ul className="space-y-1">
```
ben-dz, line 454: The copy asserts a 1:1 trace-back that doesn't hold: a pair with both endpoints unobserved is counted in both UNION branches (sum(affected_pairs) > faultCount), a disconnected pair with a stale-observed OIF matches the gate but attributes nothing (worth a test for the resulting empty/"no attribution" render), and a P+S user with a single broken tunnel lists as two "endpoints" (GROUP BY faulting_role, endpoint_user_pk). Also "unhealthy pairs" includes disconnected pairs. Juan's single-primary-faulting-endpoint fix removes most of the overlap; for the rest, group by endpoint alone with an aggregated role label and soften to "faulting pairs".

web/src/components/multicast-group-health-tab.tsx
```diff
@@ -412,6 +413,76 @@ function TableStateRow({
...
+      </div>
+      {isLoading && items.length === 0 ? (
+        <div className="text-xs text-muted-foreground">Loading…</div>
+      ) : items.length === 0 ? (
+        <div className="text-xs text-muted-foreground">
+          {faultCount.toLocaleString()} unhealthy pairs, but no single endpoint could be attributed.
+        </div>
+      ) : (
+        <>
+          <div className="text-xs text-muted-foreground mb-2">
```
ben-dz, line 449: No error state: on fetch failure `items` is [] and isLoading is false, so this renders "N unhealthy pairs, but no single endpoint could be attributed." — a confidently wrong diagnostic. Pass `rootCausesQuery.isError` into the panel and render a distinct error message (siblings in this file use TableStateRow for this).

indexer/db/clickhouse/migrations/20260709000003_health_rate_presence_only.sql
```diff
@@ -0,0 +1,199 @@
...
+    -- Verdict = control-plane + disconnected. Rate never downgrades it.
+    h.health_status AS health_status
+FROM health_multicast_user h
+LEFT JOIN user_rates ur
+    ON ur.ur_device_pk = h.user_device_pk
+   AND ur.ur_user_tunnel_id = h.user_tunnel_id
+   AND ur.ur_user_pk = h.user_pk;
+-- +goose StatementEnd
+
+-- +goose Down
```
ben-dz, line 77: The Down body restores the reconciled/mismatch/monitoring_gap/group_idle vocabulary, but this PR's web bundle deletes those keys from the badge/reason maps and narrows the TS unions — a DB-only rollback leaves unstyled badges and blank reasons. Migration 20260708000003 documented exactly this coupling in its header ("must be rolled back together with…"); add the equivalent note here.

api/handlers/multicast_health_paths.go
```diff
@@ -99,6 +99,137 @@ func (a *API) GetMulticastGroupHealthPaths(w http.ResponseWriter, r *http.Reques
...
+	}
+
+	items, err := a.queryMulticastHealthPathRootCauses(ctx, group.PK)
+	if err != nil {
+		logError("multicast group health/path-root-causes query error", "error", err, "pk", group.PK)
+		http.Error(w, dberror.UserMessage(err), http.StatusInternalServerError)
+		return
+	}
+
+	writeJSON(w, MulticastHealthPathRootCausesResponse{
```
ben-dz, line 138: No server-side LIMIT and `Total` is always len(items); after a whole-group event the panel renders an unbounded list. A LIMIT (e.g. top 25 by affected_pairs) with a true total, or a render cap with a "show all" affordance, would bound it.

web/src/components/multicast-delivery-panel.tsx
```diff
@@ -36,19 +36,15 @@ const STATUS_BADGE: Record<MulticastHealthStatus, string> = {
 }
 
 const RATE_STATUS_BADGE: Record<MulticastRateStatus, string> = {
-  reconciled: 'bg-emerald-500/15 text-emerald-500',
-  mismatch: 'bg-red-500/15 text-red-500',
+  active: 'bg-emerald-500/15 text-emerald-500',
```
ben-dz, line 39: `RATE_STATUS_BADGE[item.rate_status]` (line 278) has no `?? RATE_STATUS_BADGE.unknown` fallback, so stale 'reconciled'/'mismatch' values from a cached response or pre-migration view during the deploy window render `className="... undefined"`. The group-health tab has the fallback (line 210). Complementary to the dead-keys cleanup already requested: once legacy keys are pruned, a fallback is the only guard against an unstyled badge.

indexer/db/clickhouse/migrations/20260709000003_health_rate_presence_only.sql
```diff
@@ -0,0 +1,199 @@
...
+        NULL
+    ) AS observed_bps_5m,
+    -- Retained as NULL: reconciliation removed (see header).
+    CAST(NULL AS Nullable(Float64)) AS expected_bps_5m,
+    -- Presence-only signal; does NOT gate health_status.
+    multiIf(
+        ur.ur_present = 0, 'no_data',
+        (h.mode = 'P' AND ur.ur_max_in_bps > 0)
+            OR (h.mode IN ('S', 'P+S') AND ur.ur_max_out_bps > 0), 'active',
+        'idle'
```
ben-dz, line 63: `rate_status_reason` and `rate_status` are byte-identical multiIf blocks differing only in 'no_data' vs 'unknown' — easy future drift if one is edited without the other. Derive one: `if(rate_status_reason = 'no_data', 'unknown', rate_status_reason) AS rate_status` (alias reuse is allowed in the same SELECT).

web/src/lib/api.ts
```diff
@@ -2325,6 +2322,27 @@ export interface MulticastHealthGroupPathsResponse {
   offset: number
 }
 
+// One faulting endpoint behind the unhealthy per-path fan-out. affected_pairs
+// is how many (publisher, subscriber) pairs this endpoint drags down.
+export interface MulticastHealthPathRootCause {
+  faulting_role: 'publisher' | 'subscriber' | string
```
ben-dz: `'publisher' | 'subscriber' | string` widens to plain `string`, so the literals add no type safety. Similarly `endpoint_status` below is typed as the full MulticastHealthStatus while the Go contract is `disconnected | unhealthy`. Use the strict unions.

ben-dz (changes requested):
Deduplicated against @juan-malbeclabs review (his threads cover the BGP-down-publisher root-cause cascade at multicast_health_paths.go:174 — which were independently confirmed against the path-view definition and consider blocking — plus the hash-scroll limitation and the dead legacy badge keys). Remaining findings: the direction is sound — dropping the un-attributable rate reconciliation is well argued, the migration's output schema is byte-compatible with the prior view, the Down path is faithful, and no consumers outside api/handlers reference the view; security review is clean (parameterized SQL, sanitized errors, same middleware as siblings). Blocking: the new endpoint evaluates the expensive non-materialized path view twice per request via UNION ALL, with no page cache, refetched every 60s exactly on the hot faulty group. Mediums: affected_pairs double-counting/P+S duplication vs the panel copy's claimed 1:1 (plus an untested empty/no-attribution render), no error state in RootCausePanel, and an undocumented DB-rollback/web-deploy coupling in the migration.

- Question: the removed under-development notice explained that state-collect runs only on jump devices, so endpoints on non-collected devices falsely read unhealthy. If coverage is still jump-only, the new Root causes panel will confidently name endpoints on non-collected devices as faults — its authoritative framing raises the stakes vs the raw path table. Should the panel carry the coverage caveat?
- Question: `LEFT ANY JOIN dz_users_current` (multicast_health_paths.go:187) — a user withdrawn between the path-view scan and the join, or with pre-ingestion bgp_status = '', is labeled 'unhealthy' (fail-open, matching 20260708000002's treatment of '' as connected). Intentional here too?

api/handlers/multicast_health_paths.go
ben-malbeclabs: Fixed in cff315f. The `UNION ALL` is gone — the query now scans `health_publisher_subscriber_path` once and `ARRAY JOIN`s over two candidate endpoint tuples per pair, each gated by the observed flags (publisher when `publisher_endpoint_observed = 0`; subscriber when `subscriber_endpoint_observed = 0 AND publisher_endpoint_observed = 1`). Kept the `LEFT ANY JOIN dz_users_current` since the path view only exposes the pair-level `disconnected`, not which side is down — it's the only source of the per-endpoint BGP state needed for `endpoint_status`. Also gated the whole endpoint (and the frontend query) on `pathFaultCount > 0`, so it doesn't run at all for healthy groups. Didn't add a worker page-cache entry — the single scan plus the fault-count gate should keep it off the hot path; happy to add one if you'd still like the mirror of `/health/paths`.

ben-malbeclabs:
Thanks for the thorough pass — all findings addressed across 0481961 (BGP-down cascade + dead badge keys, shared with Juan's thread) and cff315f (everything else); per-thread replies inline.

Blocking:
- **Query cost:** collapsed the `UNION ALL` to a single scan of the path view via `ARRAY JOIN` over gated candidate tuples; endpoint gated on `pathFaultCount > 0` so it's skipped for healthy groups.
- **Unbounded list:** `LIMIT` + true distinct-endpoint total via `count() OVER ()`.

Mediums/nits: group-by-endpoint (P+S = one combined-role row) with the primary-endpoint fix removing the double-count; "faulting pairs" copy; `RootCausePanel` error state; migration Down/web rollback note; `RATE_STATUS_BADGE` fallback in the delivery panel; `rate_status` derived from the reason; strict TS unions.

Questions:
- **Coverage caveat:** state-collect now runs across the fleet (not jump-only), which is also why the under-development notice's jump-only text was removed — so the Root causes panel isn't naming endpoints on uncollected devices. No caveat added; let me know if you'd like one anyway.
- **Fail-open on `bgp_status = ''`:** yes, intentional and consistent with `20260708000002` treating `''`/`unknown` as connected — a withdrawn/pre-ingestion endpoint reads `unhealthy` rather than `disconnected`.

api/handlers/multicast_health_paths.go
ben-malbeclabs: Fixed in cff315f. Added a `LIMIT` (top 100 by `affected_pairs`) and a true distinct-endpoint `Total` via `count() OVER ()`, so `Total` no longer equals `len(items)`. The panel shows the top N and appends "(showing top N)" when `items.length < total`.

web/src/components/multicast-group-health-tab.tsx
ben-malbeclabs: Fixed in cff315f. Two changes: the single-primary-endpoint attribution (subscriber gated on `publisher_endpoint_observed = 1`) removes the both-unobserved double-count, and the query now groups by endpoint alone with an aggregated role label — a P+S user broken on both sides is one row labeled `publisher+subscriber` instead of two. Copy softened to "faulting pairs" and it now reports the true distinct-endpoint total. Added tests for the down-publisher fan-out (1 root cause, no per-subscriber rows) and the P+S combined row; the both-observed-`disconnected` "attributes nothing" case falls through to the empty-state render ("N faulting pairs, but none could be attributed to a single endpoint").

web/src/components/multicast-group-health-tab.tsx
ben-malbeclabs: Fixed in cff315f. `RootCausePanel` now takes `isError` and renders a distinct "Failed to load root causes." message, so a fetch failure no longer falls through to the "none could be attributed" diagnostic.

indexer/db/clickhouse/migrations/20260709000003_health_rate_presence_only.sql
ben-malbeclabs: Fixed in cff315f. Added a `ROLLBACK COUPLING` note to the migration header, mirroring `20260708000003`: the Down body restores the `reconciled`/`mismatch`/`monitoring_gap`/`group_idle` vocabulary that this PR's web bundle removes, so the migration must be rolled back together with the web deploy.

web/src/components/multicast-delivery-panel.tsx
ben-malbeclabs: Fixed in cff315f. Added `?? RATE_STATUS_BADGE.unknown` at the call site, matching the group-health tab. With the legacy keys pruned this is the guard against an unstyled badge if a stale `reconciled`/`mismatch` value arrives during the deploy window.

indexer/db/clickhouse/migrations/20260709000003_health_rate_presence_only.sql
ben-malbeclabs: Fixed in cff315f. `rate_status` is now derived: `if(rate_status_reason = 'no_data', 'unknown', rate_status_reason)`, so the two can't drift. Confirmed alias reuse in the same SELECT works (mroute rate tests pass against the recreated view).

web/src/lib/api.ts
ben-malbeclabs: Fixed in cff315f. Tightened to `faulting_role: 'publisher' | 'subscriber' | 'publisher+subscriber'` (the combined value the grouped query can now emit) and `endpoint_status: 'disconnected' | 'unhealthy'` to match the Go contract.

api/handlers/multicast_health_paths.go
ben-malbeclabs: Fixed. The subscriber branch is now gated on `publisher_endpoint_observed = 1`, so each pair is attributed to a single primary endpoint: a down/unobserved publisher owns all its pairs (one `disconnected` root cause, `affects N`) with no per-subscriber `unhealthy` rows. Added `TestGetMulticastGroupHealth_PathRootCauses_DownPublisherNoSubscriberFanout` (1 down publisher → 3 subscribers → exactly 1 root cause). A follow-up (cff315f) also collapsed the query to a single scan of the path view and groups by endpoint alone, so a P+S user broken on both sides is one combined-role row rather than two.

ben-malbeclabs:
Thanks Juan — both points addressed (replies in-thread):

- **Root-cause fan-out (BGP-down publisher):** each pair now attributes to a single primary endpoint (subscriber gated on the publisher being observed), so a down publisher is one `disconnected` root cause owning all its pairs — no false per-subscriber rows. Follow-up also made it a single scan + group-by-endpoint (combined role for P+S).
- **Dead badge keys / idle color:** dropped `reconciled`/`mismatch`/`monitoring_gap`/`group_idle` from `DIM_BADGE_CLASS` and aligned `idle` to amber.

Left the `#traffic` mount-only scroll as-is per your note.

web/src/components/user-detail-page.tsx
ben-malbeclabs: Agreed — leaving as-is. The link is essentially always a fresh navigation, so the mount-time scroll covers it. If we later add an in-app hash nav into this page we can key the effect on `useLocation().hash`.

juan-malbeclabs (approved):
All points from my review are resolved in cff315f: BGP-down publisher attribution collapsed to a single primary endpoint (with a fan-out test), dead badge keys removed, and the `idle` color aligned to amber. The hash-scroll limitation stays as an agreed won't-fix. LGTM.

ben-dz (approved):
All findings from both reviews are addressed in 0481961 and cff315f, verified in code: the BGP-down-publisher cascade is fixed via single-primary-endpoint attribution (subscriber gated on the publisher being observed) with a down-publisher fan-out test; the UNION ALL double scan is collapsed to a single ARRAY JOIN scan with LIMIT + a true window-count total; P+S users collapse to one combined-role row; RootCausePanel gained an error state and honest copy; the migration documents the Down-rollback/web-deploy coupling and derives rate_status from the reason; badge fallback and strict TS unions are in; Juan's dead-keys/idle-color cleanup is done. Both open questions were answered (state-collect is now fleet-wide; the bgp_status='' fail-open is intentional and consistent with 20260708000002). CI fully green including the ClickHouse-backed go-test suite. Only leftover: no fixture for the both-observed-disconnected → empty-attribution render — acknowledged, fine as a follow-up.

- Optional follow-up: add a fixture for the both-endpoints-observed 'disconnected' path (stale OIF after a BGP flap) asserting the empty 'none could be attributed' response, so the empty-state render path is exercised.
- Optional follow-up: a worker page-cache entry for /health/path-root-causes mirroring the paths cache, if the shred group's fault-state query shows up in ClickHouse metrics despite the single scan and fault-count gate.

PR #690, opened 2026-07-09
web: location-first path calculator with metro and device search

bgm-malbeclabs:
## Summary of Changes
- Search the Path Calculator by **city/metro or device in a single typeahead** (metros ranked first), so a first-time visitor can find a route without knowing DoubleZero device names — enabling the tool as a sales/discovery asset for Multicast as a Service.
- Metro selections resolve to device paths via the existing `metro-device-paths` endpoint, showing the **best single path** (lowest measured RTT, falling back to fewest hops) with an expandable "all device pairs" view for power users.
- **Mixed metro↔device** queries work by reusing the same endpoint filtered client-side — **no backend changes**; all three modes reuse existing endpoints (`/api/topology/paths`, `/api/topology/metro-device-paths`, `/api/topology`).
- Device-level search is unchanged for power users, and **legacy device-only links remain back-compatible** (new `fromType`/`toType` URL params default to device and are omitted for device selections).
- Added a sales-safe explainer above results and relabeled the map cross-link to "Open in Map".

## Diff Breakdown
| Category    | Files | Lines (+/-)   | Net  |
|-------------|-------|---------------|------|
| Core logic  |     3 | +447 / -245   | +202 |
| Tests       |     1 | +119 / -0     | +119 |
| Scaffolding |     1 | +1 / -0       |   +1 |
| **Total**   |     5 | +567 / -245   | +322 |

Almost entirely a front-end/UX change: new pure logic + one typeahead component, wired into the existing page, with the logic covered by unit tests.

<details>

- `web/src/components/path-calculator-page.tsx` — rewired the page to a unified location model: dual topology fetch (device set + metro join), query dispatch across device/metro/mixed modes, `fromType`/`toType` URL round-trip, best-path + expandable pairs rendering, explainer copy.
- `web/src/lib/path-calculator.ts` — new pure logic: option merge/ranking (`buildLocationOptions`, `filterLocations`), best-pair selection (`pickBestPair`), mixed-query filtering (`filterPairsForDevice`), URL kind parsing (`parseEndpointKind`).
- `web/src/components/topology/LocationSearch.tsx` — new unified metro/device typeahead with a Metro/Device badge per row.
- `web/src/lib/path-calculator.test.ts` — 12 unit tests for the logic above.
- `web/src/components/topology/index.ts` — export `LocationSearch` from the topology barrel.

</details>

## Testing Verification
- Added 12 unit tests covering option merge/ranking (metros-first, exact-match float, exclude/limit), best-pair selection (RTT preference with hop/metric fallback), mixed metro↔device pair filtering, and URL kind parsing incl. the legacy device default. Full web suite passes (197/197).
- End-to-end browser verification of the three query modes and a cold-cache URL round-trip (metro link + legacy device link) to run against the CI preview environment.


web/src/components/path-calculator-page.tsx
```diff
@@ -422,33 +381,29 @@ export function PathCalculatorPage() {
...
+              value={target}
               onChange={updateTarget}
-              devices={devices}
-              excludePK={sourceDevice?.pk}
+              options={options}
+              excludePK={source?.pk}
             />
-
-            {(sourceDevice || targetDevice) && (
+            {(source || target) && (
```
juan-malbeclabs, line 433: Minor UX regression on mobile: this is now the only reset-all button and it's `hidden md:block`. The previous `md:hidden` reset (that used to live in the graph-link block) was removed, so on < md viewports there's no reset-all affordance anymore — users have to clear each field via its individual `X`. Consider making this `block` (or restoring a mobile reset) to keep parity.

web/src/components/path-calculator-page.tsx
```diff
@@ -283,115 +202,147 @@ function PathCard({
...
-        pk: n.data.id,
-        code: n.data.label,
-        status: n.data.status,
-        deviceType: n.data.deviceType,
-      }))
-      .sort((a, b) => a.code.localeCompare(b.code))
-  }, [topology])
+  // Merge metros + devices into one option list. Device set comes from ISIS
+  // (as before); metro linkage (metroPK) is joined from the topology payload.
+  const options: LocationOption[] = useMemo(() => {
```
juan-malbeclabs, line 236: `fetchTopology()` pulls the full `/api/topology` payload (all devices w/ interfaces, links, validators) just to build `metroPKByDevice` and the metros list. It's the pragmatic no-backend-change choice and it's cached (staleTime 60s), so fine for now — but worth a follow-up: a lighter metros/device-metro endpoint would avoid shipping links+validators to this page.

web/src/components/path-calculator-page.tsx
```diff
@@ -283,115 +202,147 @@ function PathCard({
...
   })
 
-  // Reset selected path when results change
-  const paths = pathsResult?.paths ?? []
-  if (selectedPathIndex >= paths.length && paths.length > 0) {
-    setSelectedPathIndex(0)
-  }
+  const paths = result?.paths ?? []
+  const pairs = result?.pairs ?? null
+  if (selectedPathIndex >= paths.length && paths.length > 0) setSelectedPathIndex(0)
```
juan-malbeclabs: Nit (pre-existing, carried over): setState during render. React tolerates it, but since the surrounding block was touched this is a good moment to move it into an effect keyed on `result`.

juan-malbeclabs (approved):
Clean, well-factored change — logic/UI split is good, pure functions are non-mutating, and the unit coverage of the extracted logic is solid. No blocking correctness bugs. Approving with a few minor, non-blocking notes inline (the mobile reset button is the only one I'd actually fix before/after merge).

web/src/components/path-calculator-page.tsx
```diff
@@ -283,115 +202,147 @@ function PathCard({
...
+      pairs: MetroDevicePairPath[] | null
+      error?: string
+    }> => {
+      const s = source!
+      const t = target!
+      if (!anyMetro) {
+        const res = await fetchISISPaths(s.pk, t.pk, 5, service)
+        return { paths: res.paths ?? [], pairs: null, error: res.error }
+      }
+      const fromMetro = s.kind === 'metro' ? s.pk : s.metroPK
```
ben-dz: Same-metro mixed query is a user-reachable dead end: selecting a metro plus a device located in that same metro (excludePK only blocks the identical pk) resolves `fromMetro === toMetro`, and `GetMetroDevicePaths` rejects it (api/handlers/isis.go:2959-2962), so the user sees the raw backend string "from and to must be different metros" after picking a metro and a device. Guard `fromMetro === toMetro` client-side with a purpose-written message (or fall back to intra-metro `fetchISISPaths`, or filter same-metro options from the second typeahead), and extract the guard into path-calculator.ts so it is unit-testable.

web/src/components/path-calculator-page.tsx
```diff
@@ -283,115 +202,147 @@ function PathCard({
...
-    if (fromParam) {
-      const device = devices.find((d) => d.pk === fromParam || d.code === fromParam)
-      if (device) setSourceDevice(device)
-    }
-    if (toParam) {
-      const device = devices.find((d) => d.pk === toParam || d.code === toParam)
-      if (device) setTargetDevice(device)
-    }
-
+    if (initializedFromUrl || isisLoading || topoLoading) return
```
ben-dz, line 250: Neither topology query's `isError` is handled: after retries exhaust, `isLoading` goes false, this effect runs against empty/partial `options`, silently discards the URL `from`/`to`, and latches `initializedFromUrl = true` so a later recovery never re-initializes. The old code's `devices.length === 0` guard never marked init done on failure. Surface `isError` and only latch init once the relevant data actually arrived.

web/src/components/path-calculator-page.tsx
```diff
@@ -422,33 +381,29 @@ export function PathCalculatorPage() {
...
               onChange={updateTarget}
-              devices={devices}
-              excludePK={sourceDevice?.pk}
+              options={options}
+              excludePK={source?.pk}
             />
-
-            {(sourceDevice || targetDevice) && (
+            {(source || target) && (
               <button
```
ben-dz, line 434: Mobile users lost the reset button entirely: the old `md:hidden` reset in the "View in graph" row was removed and this remaining one is `hidden md:block`, so below the md breakpoint there is no way to reset both endpoints at once (only the per-field X). Drop `hidden md:block` or restore a mobile reset affordance.

web/src/components/path-calculator-page.tsx
```diff
@@ -283,115 +202,147 @@ function PathCard({
...
+        return { paths: [], pairs: [], error: 'Selected device has no metro assigned.' }
+      }
+      const res = await fetchMetroDevicePaths(fromMetro, toMetro, service)
+      const pairs = filterPairsForDevice(res.devicePairs ?? [], {
+        sourceDevicePK: s.kind === 'device' ? s.pk : undefined,
+        targetDevicePK: t.kind === 'device' ? t.pk : undefined,
+      })
+      const best = pickBestPair(pairs)
+      // Best pair first so the default card is the recommended path.
+      const ordered = best ? [best, ...pairs.filter((p) => p !== best)] : pairs
```
ben-dz: Metro-mode ordering/labeling is inconsistent: index 0 gets PathCard's unconditional "Shortest" badge but was chosen by `pickBestPair` (lowest measured RTT, possibly more hops), and the remaining pairs keep the backend's metric sort — so in the expanded view "Path 2" is not the second-best pair, and "Path N" implies alternate paths rather than device pairs. Pass a badge/label variant into PathCard ("Best" / "Pair N") and sort the whole pair list by the pickBestPair criterion (extract an `orderPairsBestFirst` helper into path-calculator.ts and pin it with a test).

web/src/components/topology/LocationSearch.tsx
```diff
@@ -0,0 +1,117 @@
...
+              title="Clear"
+            >
+              <X className="h-4 w-4 text-muted-foreground" />
+            </button>
+          </div>
+        ) : (
+          <>
+            <div className="relative">
+              <Search className="absolute left-3 top-1/2 -translate-y-1/2 h-4 w-4 text-muted-foreground" />
+              <input
```
ben-dz, line 72: No keyboard navigation (ArrowUp/ArrowDown/Enter/Escape) and no combobox/listbox ARIA — results are mouse-only. The sibling selectors in this same directory already implement the pattern (DeviceSelector.tsx:93-113, MetroSelector.tsx:87-107, with highlightedIndex + onKeyDown + outside-click); this adds a third, divergent typeahead with the weakest interaction model of the three. Port the DeviceSelector pattern or generalize one of the existing selectors.

web/src/components/path-calculator-page.tsx
```diff
@@ -283,115 +202,147 @@ function PathCard({
...
   )
 
-  // Fetch topology for device list
-  const { data: topology, isLoading: topologyLoading } = useQuery({
+  const { data: isis, isLoading: isisLoading } = useQuery({
     queryKey: ['isis-topology'],
     queryFn: fetchISISTopology,
     staleTime: 60000,
   })
+  const { data: topology, isLoading: topoLoading } = useQuery({
```
ben-dz: First render now blocks on the full /api/topology payload (devices, links, validators) when the page only needs metro pk/code/name — ISIS nodes already carry `metroPK` from the backend (api/handlers/isis.go:104), which the `?? n.data.metroPK` fallback acknowledges. A cold visit pays for data it never uses and the whole page spinners until both queries resolve. Prefer a lightweight metros source, or render the shell and populate the typeahead progressively instead of gating on `topoLoading`.

web/src/components/path-calculator-page.tsx
```diff
@@ -283,115 +202,147 @@ function PathCard({
...
-    error: pathsError,
-  } = useQuery<MultiPathResponse>({
-    queryKey: ['paths', sourceDevice?.pk, targetDevice?.pk, service],
-    queryFn: () => fetchISISPaths(sourceDevice!.pk, targetDevice!.pk, 5, service),
-    enabled: !!sourceDevice && !!targetDevice,
+    data: result,
+    isLoading: resultLoading,
+    error: resultError,
+  } = useQuery({
+    queryKey: ['path-calc', source?.kind, source?.pk, target?.kind, target?.pk, service],
```
ben-dz, line 319: The queryFn couples fetching with presentation (device filtering + best-first reordering) and the cache is keyed suboptimally: `device-in-A → metro B` and `metro A → metro B` are separate cache entries for the identical endpoint response, and `metroPK` is consumed in the queryFn but absent from the queryKey, so a device option that later gains a metroPK can serve a stale "no metro assigned" result until staleTime expires. Key the fetch on the resolved metro pair (['metro-device-paths', fromMetro, toMetro, service]) and do filterPairsForDevice/ordering in a useMemo over the cached response; drop the parallel `paths` array in metro mode.

web/src/components/path-calculator-page.tsx
```diff
@@ -283,115 +202,147 @@ function PathCard({
...
-    }
-
+    if (initializedFromUrl || isisLoading || topoLoading) return
+    const findOpt = (raw: string | null, kind: 'metro' | 'device') =>
+      raw ? options.find((o) => o.kind === kind && (o.pk === raw || o.code === raw)) ?? null : null
+    const s = findOpt(searchParams.get('from'), parseEndpointKind(searchParams.get('fromType')))
+    const t = findOpt(searchParams.get('to'), parseEndpointKind(searchParams.get('toType')))
+    if (s) setSource(s)
+    if (t) setTarget(t)
     setInitializedFromUrl(true)
```
ben-dz, line 260: URL→state sync is one-shot: `initializedFromUrl` latches, so future in-app deep-links into a mounted page and back/forward navigation won't update selections. Latent today (only bare sidebar links exist), but the first feature that deep-links here will silently not work. Consider deriving source/target from searchParams + options instead of state + init effect.

web/src/components/topology/LocationSearch.tsx
```diff
@@ -0,0 +1,117 @@
...
+                type="text"
+                value={search}
+                onChange={(e) => {
+                  setSearch(e.target.value)
+                  setIsOpen(true)
+                }}
+                onFocus={() => setIsOpen(true)}
+                onBlur={() => setTimeout(() => setIsOpen(false), 150)}
```
ben-dz, line 80: The `onBlur` + setTimeout(150) close and the option `onMouseDown` preventDefault are redundant with each other; grabbing the dropdown's scrollbar (max-h-60 overflow-y-auto) blurs the input and closes the list mid-scroll; and the timeout isn't cleared on unmount. Prefer the mousedown-outside listener the sibling selectors use.

web/src/components/path-calculator-page.tsx
ben-dz: Conditional setState during render won't loop, but combined with the parallel paths/pairs indexing it's fragile — a clamp is simpler: `const effectiveIndex = Math.min(selectedPathIndex, Math.max(paths.length - 1, 0))`.

web/src/components/path-calculator-page.tsx
```diff
@@ -283,115 +202,147 @@ function PathCard({
...
-    }
-    if (toParam) {
-      const device = devices.find((d) => d.pk === toParam || d.code === toParam)
-      if (device) setTargetDevice(device)
-    }
-
+    if (initializedFromUrl || isisLoading || topoLoading) return
+    const findOpt = (raw: string | null, kind: 'metro' | 'device') =>
+      raw ? options.find((o) => o.kind === kind && (o.pk === raw || o.code === raw)) ?? null : null
+    const s = findOpt(searchParams.get('from'), parseEndpointKind(searchParams.get('fromType')))
```
ben-dz, line 256: URL edge cases: unknown from/to values are silently dropped while remaining in the address bar; a hand-edited `from == to` reaches the backend and surfaces its raw error (device case pre-existing, metro case new); and `?service=foo` is cast to PathService unvalidated and forwarded until the backend rejects it. Consider clearing unresolvable params on init, validating service against the two known values, and guarding from === to.

web/src/components/path-calculator-page.tsx
```diff
@@ -283,115 +202,147 @@ function PathCard({
...
-  const paths = pathsResult?.paths ?? []
-  if (selectedPathIndex >= paths.length && paths.length > 0) {
-    setSelectedPathIndex(0)
-  }
+  const paths = result?.paths ?? []
+  const pairs = result?.pairs ?? null
+  if (selectedPathIndex >= paths.length && paths.length > 0) setSelectedPathIndex(0)
+
+  // Device pks used for the "View in graph" link (resolve metros via the best/selected pair).
+  const graphSourcePK = anyMetro ? pairs?.[selectedPathIndex]?.sourceDevicePK : source?.pk
```
ben-dz, line 358: `refetchOnWindowFocus` is enabled globally; a focus refetch can reorder `pairs` (new latency → different best) while `selectedPathIndex > 0`, silently repointing the selected card and these "View in graph" PKs at a different device pair.

web/src/lib/path-calculator.ts
```diff
@@ -0,0 +1,112 @@
...
+  opts: { sourceDevicePK?: string; targetDevicePK?: string },
+): MetroDevicePairPath[] {
+  return pairs.filter((p) => {
+    if (opts.sourceDevicePK && p.sourceDevicePK !== opts.sourceDevicePK) return false
+    if (opts.targetDevicePK && p.targetDevicePK !== opts.targetDevicePK) return false
+    return true
+  })
+}
+
+// Re-exported for Task 2 consumers; kept here so all path-calc logic lives in one module.
```
ben-dz: Leftover plan-scaffolding comment: "Re-exported for Task 2 consumers" references a working-plan task numbering that means nothing in the repo, and no consumer imports MetroDevicePairPath/SinglePath from this module (SinglePath is imported only to be re-exported). Drop the re-export and comment.

ben-dz (changes requested):
Well-structured PR: the pure logic in `web/src/lib/path-calculator.ts` is cleanly separated and unit-tested (197/197 web tests pass, tsc/eslint clean), legacy device-only URLs stay back-compatible, and the security review found nothing. One High finding blocks: a mixed metro↔device selection where the device is inside the selected metro resolves `fromMetro === toMetro` and always fails with the raw backend error "from and to must be different metros" — a user-reachable dead end in the headline search-by-city feature. Medium findings: silent URL-param drop when topology/ISIS fetches fail, mobile reset button removed, misleading "Shortest" badge / mixed pair ordering in metro mode, no keyboard navigation in the new typeahead (sibling selectors have it), heavy `/api/topology` fetch gating first render, and queryFn coupling fetch with presentation (cache duplication, `metroPK` missing from the queryKey). Low findings cover one-shot URL sync, dropdown blur mechanics, render-phase setState, URL edge cases, focus-refetch reordering the selected pair, and a leftover "Task 2" scaffolding comment.

- Test coverage: the 12 unit tests are at the right altitude for the lib (URL back-compat, exclude+limit, RTT-vs-hops fallback, pair filtering). Untested and worth covering: the best-first reordering (currently in the component, not the lib), the queryFn dispatch (device/device vs any-metro, missing-metroPK error, and the same-metro guard once added), and the writeEndpoint ↔ init-effect URL round-trip.
- Security review (dz-security-reviewer): no findings. URL params act only as selectors into server-fetched data, all fetch URLs are encoded, the backend validates and parameterizes everything, and all rendering goes through JSX escaping. Optional hardening: allowlist-validate `service` on read and encodeURIComponent the graph-link params.

web/src/components/path-calculator-page.tsx
bgm-malbeclabs: Fixed in c9aaf21 — dropped `hidden md:block` so the field-row reset shows on all viewports (it now appears as soon as either endpoint is set, vs. only when both were). Verified on the preview at 390px.

web/src/components/path-calculator-page.tsx
bgm-malbeclabs: Fixed in c9aaf21 — dropped `hidden md:block`; the reset now shows below the md breakpoint. Verified on the preview at 390px.

web/src/components/path-calculator-page.tsx
bgm-malbeclabs: Fixed in c9aaf21 — moved into a `useEffect` keyed on `result`.

bgm-malbeclabs: Fixed in c9aaf21 — moved the clamp into a `useEffect` keyed on `result` (guards on `result?.paths?.length`, no external deps).

web/src/components/path-calculator-page.tsx
bgm-malbeclabs: Fixed in 4f8cab9 — extracted `resolveMetroPair` into `path-calculator.ts`; a same-metro selection now returns a friendly "Source and destination are in the same metro" message instead of the raw backend error (unit-tested). Verified on the preview with NYC metro + an NYC device.

web/src/components/path-calculator-page.tsx
bgm-malbeclabs: Fixed in 4f8cab9 — the init effect no longer latches while `options` is empty (failed/empty fetch), so the URL from/to aren't silently dropped and a later recovery re-initializes; added an `isError` fetch-error state.

web/src/components/path-calculator-page.tsx
bgm-malbeclabs: Fixed in 4f8cab9 — added `orderPairsBestFirst` (sorts the whole pair list by the pickBestPair criterion, unit-tested) and gave PathCard `label`/`badge` props so metro mode reads "Best" / "Pair N". Verified on the preview: 48 pairs render ascending by measured RTT, Pair 1 badged "Best".

web/src/lib/path-calculator.ts
bgm-malbeclabs: Fixed in 4f8cab9 — dropped the dead re-export and the Task-2 comment (consumers import these types from `@/lib/api`, nothing imports them from this module).

web/src/components/path-calculator-page.tsx
bgm-malbeclabs: Partially addressed: the `from === to` metro case is now guarded by the same-metro fix in 4f8cab9. Unknown-param clearing on init and `service` allowlist validation are deferred to #691.

web/src/components/path-calculator-page.tsx
bgm-malbeclabs: Agreed — deferred to #691 (lighter metros source / progressive render).

web/src/components/path-calculator-page.tsx
bgm-malbeclabs: Deferred to #691 (lighter metros source, or render the shell and populate the typeahead progressively).

web/src/components/path-calculator-page.tsx
bgm-malbeclabs: Deferred to #691 (re-key the fetch on the resolved metro pair and move filtering/ordering into a `useMemo` over the cached response).

web/src/components/topology/LocationSearch.tsx
bgm-malbeclabs: Deferred to #691 (port the DeviceSelector keyboard/ARIA pattern into LocationSearch).

web/src/components/path-calculator-page.tsx
bgm-malbeclabs: Deferred to #691 (derive source/target from searchParams + options rather than a one-shot init effect).

web/src/components/topology/LocationSearch.tsx
bgm-malbeclabs: Deferred to #691 (switch to the mousedown-outside listener the sibling selectors use).

web/src/components/path-calculator-page.tsx
bgm-malbeclabs: Deferred to #691.

ben-dz (approved):

PR #694, opened 2026-07-15
network-health: add the Network Health Reporting page

thijsvanemmerik:
## Summary

Adds the Network Health Reporting page, a public windowed report of how the DoubleZero network performed over a selected time range, organized into three sections:

- **Performance**: the links and devices that were least available over the window (time split across available, maintenance drain, and fault), and committed versus measured latency for links that carry an onchain committed round-trip time.
- **Capacity**: the links closest to their provisioned bandwidth, DIA interface utilization against port speed, the links carrying the most traffic, and the devices closest to their user-slot capacity.
- **Operations**: link and device failures over time and by duration, the links and devices with the most fault time, how quickly links were drained and returned to service, and an incidents and maintenance summary from the ops-management portal.

Every metric supports a selectable window (default 30 days) and is compared against the previous equal-length window. The report can be viewed for the whole network or scoped to a single contributor, and presents the figures neutrally without ranking or grading contributors.

Each panel is served by its own endpoint and cached independently, so the page renders progressively and a slow query on one panel does not hold up the others. Expensive aggregates are pre-computed by a background refresh.

## Testing Verification

- Tests cover outage-episode reconstruction (the sustained-failure floor and episode boundaries), impact-weighted downtime (primary-path and traffic weighting), per-contributor scope isolation across the availability, outage, and drain data, drain and undrain timing, previous-window delta direction, and the JSON contract of each group endpoint.

### DIFF
--- a/api/handlers/network_health.go
+++ b/api/handlers/network_health.go
@@ -624,17 +624,21 @@ const nhGroupDeadline = 35 * time.Second
 // directly (bypassing these handlers) so it is excluded too.
 var nhLiveComputeSem = make(chan struct{}, 4)
 
-// nhAcquireLive tries to reserve a live-compute slot within a short timeout. On
-// success it returns a release func and true; on failure it returns false and the
-// caller must return 503 without launching a query. Release must be deferred.
+// nhAcquireLive reserves a live-compute slot, waiting for one to free rather
+// than shedding immediately. A scoped page load fires ~9 group requests at once
+// and every one computes live (scoped views are never precomputed); with only 4
+// slots, an immediate-shed policy 503'd the requests that lost the race, so the
+// UI showed "Couldn't load" on ~5 of 9 panels. Waiting for a slot instead lets
+// them queue and fill in progressively as the running queries finish. The wait
+// is bounded by ctx (the caller passes the group deadline), so a request still
+// sheds if its deadline expires while queued (genuine sustained overload) or if
+// the client disconnected. On success it returns a release func and true; on
+// failure it returns false and the caller must return 503 without launching a
+// query. Release must be deferred.
 func nhAcquireLive(ctx context.Context) (func(), bool) {
-	timer := time.NewTimer(2 * time.Second)
-	defer timer.Stop()
 	select {
 	case nhLiveComputeSem <- struct{}{}:
 		return func() { <-nhLiveComputeSem }, true
-	case <-timer.C:
-		return nil, false
 	case <-ctx.Done():
 		return nil, false
 	}
@@ -668,17 +672,21 @@ func (a *API) serveNHGroup(
 	)
 
 	writeLive := func(contrib string) {
-		// Bound concurrent live computes; shed load with 503 rather than piling on
-		// another ClickHouse query when the limit is saturated.
-		release, ok := nhAcquireLive(r.Context())
+		// One deadline covers the whole live operation: waiting for a slot plus
+		// the query. Concurrent ClickHouse queries are still bounded to 4, but
+		// waiters queue for a free slot instead of shedding on contention, so a
+		// single page load's ~9 concurrent group requests fill in progressively
+		// rather than 503-ing each other. Only a request whose deadline expires
+		// while queued (sustained overload) or whose client disconnected is shed.
+		ctx, cancel := context.WithTimeout(r.Context(), deadline)
+		defer cancel()
+		release, ok := nhAcquireLive(ctx)
 		if !ok {
 			nhWriteLiveUnavailable(w)
 			return
 		}
 		defer release()
 		w.Header().Set("X-Cache", "MISS")
-		ctx, cancel := context.WithTimeout(r.Context(), deadline)
-		defer cancel()
 		resp := fetch(ctx, start, end, contrib)
 		w.Header().Set("Content-Type", "application/json")
 		if err := json.NewEncoder(w).Encode(resp); err != nil {
--- a/api/handlers/network_health.go
+++ b/api/handlers/network_health.go
@@ -1,0 +1,3453 @@
+package handlers
+
+// Network Health Reporting: a public, windowed report of how the DoubleZero
+// network performed over a time window (default: last 30 days), network-wide.
+//
+// Design notes (see plans/network-health-dashboard/SPEC.md):
+//   - Facts only. This endpoint returns raw measurements; it does not score,
+//     rank, or judge contributors. The frontend must present them neutrally.
+//     One deliberate exception: NHTickets.SelfReportedPct (the share of a
+//     contributor's own incidents that the contributor filed themselves) may be
+//     graded good/warn/bad on that contributor's OWN view. It is the
+//     contributor's own self-fact about their own incidents, shown only on their
+//     own page (never a cross-contributor leaderboard), so it does not judge one
+//     contributor against another.
+//   - Safety: every query carries an inline SETTINGS cap and the whole fetch
+//     runs under a context deadline. Time bounds are passed as bound parameters
+//     (time.Time), never string-interpolated. Rollup/percentile values are
+//     already clean rates, so no counter-wrap guard is needed there; the raw
+//     fact table (traffic bytes) keeps the out_octets_delta > 0 guard.
+//   - The default (30d) view is precomputed by the worker into page_cache and
+//     served on HIT; custom windows compute live under the deadline.
+
+import (
+	"context"
+	"encoding/json"
+	"math"
+	"net/http"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	// The Network Health page is split into independent data-source groups, each
+	// with its OWN page_cache key so it loads and caches on its own. Each suffix is
+	// bumped independently when that one group's payload shape changes, so a stale
+	// blob from an older shape is never served (it would crash the newer frontend).
+	// The bare route /api/network-health serves the Overview group (see
+	// GetNetworkHealth), so its key stands in for the old monolith key.
+	NetworkHealthOverviewCacheKey     = "network_health_overview_30d_v1"
+	NetworkHealthAvailabilityCacheKey = "network_health_availability_30d_v1"
+	NetworkHealthLatencyCacheKey      = "network_health_latency_30d_v1"
+	NetworkHealthCapacityCacheKey     = "network_health_capacity_30d_v1"
+	NetworkHealthOutagesCacheKey      = "network_health_outages_30d_v1"
+	NetworkHealthDrainCacheKey        = "network_health_drain_30d_v1"
+	NetworkHealthTicketsCacheKey      = "network_health_tickets_30d_v1"
+	NetworkHealthImpactfulCacheKey    = "network_health_impactful_30d_v1"
+	// NetworkHealthDeferredCacheKey is the page_cache key for the default 30d
+	// deferred payload (the slow undrain/recovery-health figures, split out of the
+	// main page so the main page loads without waiting on the heavy scan).
+	NetworkHealthDeferredCacheKey = "network_health_deferred_30d_v1"
+	// networkHealthContributorCacheKeyInfix separates a group's version suffix
+	// from the contributor code in a per-contributor cache key (see nhContribKey).
+	networkHealthContributorCacheKeyInfix = "_c_"
+	// NetworkHealthDefaultDays is the default (and cached) window length.
+	NetworkHealthDefaultDays = 30
+
+	networkHealthMaxDays = 92
+	// max_memory_usage caps each query at ~2 GiB so a single heavy query errors
+	// (and degrades to an empty panel via nhGo) instead of breaching the server
+	// total memory limit and taking down sibling queries. Public-page safety.
+	// max_bytes_before_external_group_by / max_bytes_before_external_sort let a
+	// big GROUP BY or ORDER BY spill to disk well before that cap, so a heavy
+	// 30d scan over device_interface_rollup_5m/link_rollup_5m (sorted by
+	// bucket_ts first, so a per-entity filter still scans the whole range)
+	// finishes slower instead of erroring with code 241 ("memory limit
+	// exceeded ... while reading column device_pk/link_pk").
+	// max_execution_time = 55 is chosen against two different callers of these
+	// queries: the page-cache worker's refresh() wraps each cache-refresh call
+	// in its own 60s context (api/worker/workflow.go), so 55s leaves headroom
+	// for the worker to finish every panel on one refresh instead of leaving it
+	// empty until a later tick; live requests (contributor scope, see
+	// GetNetworkHealth) are bounded by their own ~35s request context, which
+	// cancels the ClickHouse query client-side well before 55s, so raising this
+	// shared setting does not relax anything on that path.
+	networkHealthQuerySettings = " SETTINGS max_execution_time = 55, max_memory_usage = 2000000000, max_bytes_before_external_group_by = 1000000000, max_bytes_before_external_sort = 1000000000, timeout_before_checking_execution_speed = 0"
+
+	// networkHealthDeferredQuerySettings is networkHealthQuerySettings with a much
+	// higher max_execution_time (170s) for the deferred recovery-health scan. That
+	// scan (fetchRecoveryHealth over link_rollup_5m) is the one query that could
+	// exceed the main page's 35s/55s budget, so it is served from its own deferred
+	// endpoint (GetNetworkHealthDeferred) under a 170s request deadline. The memory
+	// cap and spill settings are unchanged (still 2 GiB), so this only buys time,
+	// never more memory.
+	networkHealthDeferredQuerySettings = " SETTINGS max_execution_time = 170, max_memory_usage = 2000000000, max_bytes_before_external_group_by = 1000000000, max_bytes_before_external_sort = 1000000000, timeout_before_checking_execution_speed = 0"
+
+	// throughputMaxThreads caps parallelism on the one query that dedups the
+	// whole user-traffic slice of device_interface_rollup_5m (~14.4M rows over
+	// 30d, one dedup group per row). At default parallelism each thread buffers
+	// its own native-stream read of that slice and the query breaches the 2 GB
+	// cap at the source read (code 241, SourceFromNativeStream); 4 threads keeps
+	// the concurrent read buffers small enough to finish under the cap (measured
+	// ~40 ms vs OOM). Appended after networkHealthQuerySettings so the later
+	// max_threads wins. This is the scan the network peak is derived from and the
+	// one that renders the throughput time series, so it fixes both.
+	throughputMaxThreads = ", max_threads = 4"
+)
+
+// nhGo runs one panel query concurrently as best-effort. If it fails or times out
+// (e.g. a heavy scan hitting max_execution_time on the remote proxy) the error is
+// logged with the panel name and the goroutine returns nil, so a single slow or
+// broken panel degrades to an empty box instead of cancelling every sibling query
+// and blanking the whole report. The panel name makes the failure diagnosable.
+func nhGo(g *errgroup.Group, panel string, fn func() error) {
+	g.Go(func() error {
+		if err := fn(); err != nil {
+			logError("network health panel failed", "panel", panel, "error", err)
+		}
+		return nil
+	})
+}
+
+// The Network Health page is served as independent data-source groups, each its
+// own endpoint + cache blob (see the *CacheKey constants and the Get* handlers).
+// Every group is a subset of the old monolith payload; nothing is dropped and no
+// field is owned by more than one group. Each group carries the resolved window
+// (so the frontend can label it), a generated_at stamp, and an optional error
+// (set only on the non-scoped path for the headline-critical groups, so the cache
+// worker keeps the last good blob instead of caching zeros — see the worker
+// entries in api/worker/workflow.go).
+
+// NHOverview is the Overview group: the headline signals plus trend context. It
+// deliberately does NOT own the outage/impactful headline tiles (those come from
+// the Outages and Impactful groups so their heavy scans never block Overview);
+// the frontend headline row cross-reads all three. Contributors is the list that
+// backs the filter-bar dropdown (code + footprint only, no outage breakdown).
+type NHOverview struct {
+	Window       NHWindow        `json:"window"`
+	Headline     NHHeadline      `json:"headline"`
+	Isis         NHIsis          `json:"isis"`
+	Freshness    NHFreshness     `json:"freshness"`
+	Throughput   []NHTsPoint     `json:"throughput_ts"`
+	Contributors []NHContributor `json:"contributors"`
+	GeneratedAt  string          `json:"generated_at"`
+	Error        string          `json:"error,omitempty"`
+}
+
+// NHAvailabilityGroup is the Availability group (no deltas).
+type NHAvailabilityGroup struct {
+	Window             NHWindow         `json:"window"`
+	LinkAvailability   []NHAvailability `json:"link_availability"`
+	DeviceAvailability []NHAvailability `json:"device_availability"`
+	GeneratedAt        string           `json:"generated_at"`
+	Error              string           `json:"error,omitempty"`
+}
+
+// NHLatencyGroup is the Latency group: committed-vs-measured link latency plus
+// the committed-RTT (SLA) adherence tile (no deltas).
+type NHLatencyGroup struct {
+	Window       NHWindow     `json:"window"`
+	LatencyLinks []NHPerfLink `json:"latency_links"`
+	Sla          NHSla        `json:"sla"`
+	GeneratedAt  string       `json:"generated_at"`
+	Error        string       `json:"error,omitempty"`
+}
+
+// NHCapacityGroup is the Capacity group. Capacity is the network SEAT-utilization
+// summary; CapacityLinks is the fullest-links (bandwidth) capacity-planning
+// panel — two different concepts (no deltas).
+type NHCapacityGroup struct {
+	Window        NHWindow         `json:"window"`
+	Capacity      NHCapacity       `json:"capacity"`
+	DeviceSlots   []NHDeviceSlots  `json:"device_slots"`
+	DiaInterfaces []NHDiaInterface `json:"dia_interfaces"`
+	TopLinks      []NHTrafficLink  `json:"top_links"`
+	CapacityLinks []NHCapacityLink `json:"capacity_links"`
+	GeneratedAt   string           `json:"generated_at"`
+	Error         string           `json:"error,omitempty"`
+}
+
+// NHOutagesGroup is the Outages group: the reliability episode scan (which also
+// feeds the headline OutageCount tile + its delta), outage summary, downtime
+// rankings, outages-over-time, and interface-error hotspots. Prev carries the
+// prior-window values for the reliability + outage-summary deltas. The
+// per-contributor footprint lives in the Overview group (the scope dropdown's
+// source), not here.
+type NHOutagesGroup struct {
+	Window           NHWindow         `json:"window"`
+	Reliability      NHReliability    `json:"reliability"`
+	OutageCount      uint64           `json:"outage_count"`
+	OutageCountDelta *float64         `json:"outage_count_delta"`
+	OutageSummary    *NHOutageSummary `json:"outage_summary"`
+	DowntimeLinks    []NHDowntimeRow  `json:"downtime_links"`
+	DowntimeDevices  []NHDowntimeRow  `json:"downtime_devices"`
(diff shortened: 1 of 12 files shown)


api/handlers/network_health.go
```diff
@@ -0,0 +1,3461 @@
...
+	nhGo(g, "device_availability", func() error {
+		rows, err := a.fetchDeviceAvailability(ctx, start, end, contrib)
+		if err != nil {
+			return err
+		}
+		resp.DeviceAvailability = rows
+		return nil
+	})
+	if err := g.Wait(); err != nil {
+		logError("network health availability: partial data", "error", err)
```
ben-dz: **Blocking** — Four of the nine groups cannot report a failed query, so failed queries are served and cached as zeros. `NHAvailabilityGroup`, `NHLatencyGroup`, `NHCapacityGroup` and `NHDrainGroup` declare an `error` field that is never assigned, and `nhGo` swallows each panel error to nil; the frontend's `GroupBoundary` keys off `data.error` and so renders the data branch ("No data in this window", "0 drains", empty tables), and the worker writes the zeroed blob unconditionally so it is later served as `X-Cache: HIT`. Overview and outages additionally gate `resp.Error` on `!scoped` (:1330, :1683), so a contributor view shows "0 link failures" from a failed query. Set the group error flag whenever a panel fails, scoped included, and return `refreshError` from the worker for any group with a failed panel. Same root cause at :1023, where `fetchContribLinkPKs` returns `(nil, false)` for both a query failure and "owns no links".

api/handlers/network_health.go
```diff
@@ -0,0 +1,3461 @@
...
+		outageContrib = contrib
+		scopePubkey = cscope.pubkey
+	}
+
+	// Split tickets into current (created >= start) and prior (priorStart <= created < start).
+	var curTickets, priorTickets []nhRawTicket
+	for _, t := range tickets {
+		ct, ok := nhParseTicketTime(&t.CreatedAt)
+		switch {
+		case !ok || !ct.Before(start):
```
ben-dz: **Blocking** — The tickets "current window" has no upper bound, so it counts tickets created after the window ends. `fetchOpsTicketsSince(priorStart)` filters only `created_at >= priorStart`, `end` is discarded, and everything with `!ct.Before(start)` lands in current — the default view over-counts by up to a day, and a custom historical window (`?start=2026-01-01&end=2026-01-31`) counts every ticket filed since January. `fetchOutageListForTickets` *is* bounded to `[priorStart, end)`, so `total`/`incidents`/`sev1-3` cover a different window than the `outage_count` and `outages_no_ticket` figures shown beside them. Drop tickets with `created >= end` and exclude unparseable timestamps instead of defaulting them into current; no test covers this split.

api/handlers/network_health.go
```diff
@@ -0,0 +1,3461 @@
...
+			if data, err := a.readPageCache(r.Context(), NetworkHealthDeferredContributorCacheKey(code)); err == nil {
+				w.Header().Set("Content-Type", "application/json")
+				w.Header().Set("X-Cache", "HIT")
+				_, _ = w.Write(data)
+				return
+			}
+		}
+
+		// Bound concurrent live computes (see nhLiveComputeSem); shed with 503.
+		release, ok := nhAcquireLive(r.Context())
```
ben-dz: **Blocking** — `/deferred` waits for a live-compute slot with no deadline, so queued requests accumulate without bound. Both paths here (:822, :850) call `nhAcquireLive(r.Context())` and only build the 170s context after the slot is granted, whereas every other group bounds the same wait by its group deadline (`serveNHGroup`, :681) — which is what the queue design relies on. With `WriteTimeout: 0` (api/main.go:853) nothing else caps it: a client that opens connections and holds them parks goroutines indefinitely while four 170s scans run, keeping the shared `nhLiveComputeSem` saturated so every other Network Health live request sheds at its 35s deadline. Any non-default `days` bypasses the cache, so this is reachable unauthenticated. Build the 170s context first and pass it to `nhAcquireLive`.

api/handlers/network_health.go
```diff
@@ -0,0 +1,3461 @@
...
+
+// nhGo runs one panel query concurrently as best-effort. If it fails or times out
+// (e.g. a heavy scan hitting max_execution_time on the remote proxy) the error is
+// logged with the panel name and the goroutine returns nil, so a single slow or
+// broken panel degrades to an empty box instead of cancelling every sibling query
+// and blanking the whole report. The panel name makes the failure diagnosable.
+func nhGo(g *errgroup.Group, panel string, fn func() error) {
+	g.Go(func() error {
+		if err := fn(); err != nil {
+			logError("network health panel failed", "panel", panel, "error", err)
```
ben-dz: **Blocking** — A persistently failing panel pages on-call once per refresh cycle, indefinitely. `nhGo` reports every panel failure through `logError` → `logger.Error`, which logs at ERROR for anything `dberror.IsTransient` rejects — including the ClickHouse code-241 memory-limit errors this file's own comments name as the expected failure mode at the 2 GiB cap. The worker calls these fetchers every cycle, so one broken panel emits an alert-bearing line per cycle forever, for a condition that degrades to an empty box and has no terminal action. Route worker-reachable panel failures through a `logger.Escalator` keyed per panel, or log WARN and let the worker's existing escalator own the alert.

api/worker/workflow.go
```diff
@@ -168,6 +175,26 @@ type Activities struct {
 func (a *Activities) entries() []cacheEntry {
 	api := a.API
 	return []cacheEntry{
+		// The two heavy Network Health groups are scheduled FIRST so they start at
+		// the head of the concurrency-limited batch and have the whole activity
+		// budget to finish their 170s queries, rather than being cut when scheduled
+		// last. Their per-entry timeout is 180s (see cacheEntry.timeout).
+		{name: "network health impactful", key: handlers.NetworkHealthImpactfulCacheKey, timeout: 180 * time.Second, fn: func(ctx context.Context) (any, error) {
```
ben-dz: **Blocking** — These two 180s entries consume the entire page-cache activity budget, starving and escalating unrelated caches. `timeout: 180s` equals `defaultActivityTimeout`, and `RefreshCaches` blocks on `g.Wait()` for all 38 entries — so the batch's wall clock is now bounded below by these two scans (documented as heavier than the 35s page deadline) instead of the previous 60s per-entry ceiling, stretching every other cached page's refresh cadence against the invariant stated at :378. With zero headroom left, a scan past ~178s trips the activity's `StartToCloseTimeout` and every in-flight entry records `errBatchDeadline` (classified non-transient), so topology/status/validators page on-call after three cycles. `maxAttempts = 2` is also unreachable here. Give these two their own activity or an `everyN` cadence, or raise `ActivityTimeout` above their budget.

api/handlers/network_health.go
```diff
@@ -0,0 +1,3461 @@
...
+	End   string `json:"end"`
+	Days  int    `json:"days"`
+	Label string `json:"label"`
+}
+
+type NHHeadline struct {
+	PeakBps                float64  `json:"peak_bps"`
+	JitterImprovePct       float64  `json:"jitter_improve_pct"`
+	DzLossPct              float64  `json:"dz_loss_pct"`
+	OutageCount            uint64   `json:"outage_count"`
```
ben-dz: **Minor** — `NHHeadline.OutageCount`, `NHHeadline.ImpactfulDowntimeHours` and the matching `NHDeltas` fields are never assigned by `FetchNetworkHealthOverviewData`, so the documented public `/api/network-health` payload always reports `"outage_count": 0` for a network that had outages. The frontend reads the Outages/Impactful groups instead, so nothing catches it. Delete the four fields and bump the overview cache key.

api/handlers/network_health.go
```diff
@@ -0,0 +1,3461 @@
...
+
+// nhContribKey builds one contributor's default-window (30d) cache key for a
+// group, so each per-contributor key carries the same version suffix as its
+// group's network-wide key and invalidates together with that group's shape.
+func nhContribKey(base, code string) string {
+	return base + networkHealthContributorCacheKeyInfix + code
+}
+
+// Per-group per-contributor cache-key builders (see nhContribKey).
+func NetworkHealthOverviewContributorCacheKey(code string) string {
```
ben-dz: **Minor** — Nothing writes a per-contributor cache key, so these nine builders, the infix constant, `FetchActiveContributorCodes` and all nine cache-HIT branches are permanently-MISS dead paths. `serveNHGroup`'s doc comment still asserts the default window "is precomputed per contributor by the worker", which sends the next maintainer looking for a writer that does not exist. Delete them along with the stale comments.

api/handlers/network_health.go
```diff
@@ -0,0 +1,3461 @@
...
+	type tev struct {
+		ts    time.Time
+		drain bool // true = drained, false = undrained (soft-drained -> activated)
+	}
+	timeline := map[string][]tev{}
+	drainsByLink := map[string][]time.Time{}
+	var drains, undrains int
+	for _, c := range changes {
+		switch {
+		case c.next == "soft-drained":
```
ben-dz: **Minor** — `computeDrainTiming` recognises only `→ soft-drained` as a drain and only `soft-drained → activated` as an undrain, but `hard-drained` is a real link status the availability query already handles (:3034). A `soft-drained → hard-drained → activated` sequence leaves `drainStart` open forever, so `drains` counts it while `undrains` and `time_drained` silently drop it; `hard-drained → soft-drained` double-counts a drain. Treat any transition into/out of a drained state accordingly, and add both cases to `TestComputeDrainTiming`.

api/handlers/network_health.go
```diff
@@ -0,0 +1,3461 @@
...
+			SELECT link_pk, bucket_ts,
+				sum(is_start) OVER (PARTITION BY link_pk ORDER BY bucket_ts
+					ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS eid
+			FROM (
+				SELECT link_pk, bucket_ts,
+					if(dateDiff('second',
+						lagInFrame(bucket_ts) OVER (PARTITION BY link_pk ORDER BY bucket_ts),
+						bucket_ts) > 300, 1, 0) AS is_start
+				FROM link_rollup_5m
+				WHERE bucket_ts >= ? AND bucket_ts < ? AND isis_down = 1 AND status = 'activated'` + filter + `
```
ben-dz: **Minor** — Two different outage definitions ship as `outage_count` on the same page. `fetchLinkDownEvents` counts `isis_down = 1` runs with no loss condition, no `provisioning = false`, no `FINAL` and no duration floor, and its count renders as "N link-down events" beside the Outages group's `dur_s >= 600` episode count labelled "N link failures". The predicate is copy-pasted about ten times across the file, which is how the two drifted apart. Derive both from the shared `outEpisodeCTE`.

web/src/components/network-health-reporting-page.tsx
```diff
@@ -0,0 +1,2285 @@
...
+        </div>
+      )}
+    </>
+  )
+}
+
+// Traffic-weighted availability, derived on the frontend from the Overview active
+// link count and the traffic-weighted failure hours (same formula the server used:
+// 1 - impactful_hours / (active_links x window_hours)).
+function deriveAvailability(o: NHOverview | undefined, imp: NHImpactful | undefined): number | null {
```
ben-dz: **Minor** — `deriveAvailability` combines `overview.window.days` with `impactful.impactful_downtime_hours` without checking the two describe the same window. Both queries use `keepPreviousData` and the impactful scan can take 170s, so switching 30d→7d divides 30-day failure hours by a 168-hour denominator — the headline availability figure reads far too low until impactful catches up. `NHImpactful.window` is already in the payload; return null unless its start/end match the overview's.

api/handlers/network_health.go
```diff
@@ -0,0 +1,3461 @@
...
+			}
+		}
+		startT, hasStart := nhParseTicketTime(t.StartAt)
+		createdT, hasCreated := nhParseTicketTime(&t.CreatedAt)
+
+		switch t.Type {
+		case "incident":
+			agg.Incidents++
+			if t.RootCause != nil && *t.RootCause != "" {
+				rootCauseCounts[*t.RootCause]++
```
ben-dz: **Minor** — The upstream `root_cause` field is tallied verbatim and published as `NHRootCauseCount.Cause` on an unauthenticated endpoint, while the declared contract is "a fixed enum token (never free text)" (:304). There is no allowlist at the sink, and `TestNHTicketsNoFreeText` asserts JSON field names only — so any operator-entered or newly added upstream value is published as-is. The same ops-ticket source is otherwise kept behind `RequireAuth` for PII (api/main.go:594). Allowlist the seven documented tokens and bucket anything else as `other`.

api/handlers/network_health.go
```diff
@@ -0,0 +1,3461 @@
...
+			}
+			logError("network health: ops tickets pagination stopped early (bad json)", "error", err, "page", page, "collected", len(out))
+			break
+		}
+		tickets := env.Data.Tickets
+		if len(tickets) == 0 {
+			break
+		}
+		for _, t := range tickets {
+			if t.ID != "" && t.CreatedAt >= sinceStr {
```
ben-dz, line 2473: **Minor** — One public `/tickets` request can drive up to 100 sequential credentialed GETs to the ops-management API. The loop's only early exit is "this page's oldest ticket predates the cutoff", and `networkHealthWindow` clamps a window's span to 92 days but not its position, so `?start=1970-01-01&end=1970-04-01` never trips it; an unknown `?contributor=` value is not validated until after the fetch. `/users` is also re-paged uncached on every request. Floor the ops lookback and memoize the users registry for the refresh interval.

api/handlers/network_health.go
```diff
@@ -0,0 +1,3461 @@
...
+// shares of link_rollup_5m, excluding provisioning buckets entirely. Links with
+// no classified buckets in the window (a full gap) are excluded rather than
+// counted as 0% available.
+func (a *API) fetchLinkAvailability(ctx context.Context, start, end time.Time, contrib string) ([]NHAvailability, error) {
+	scope := ""
+	args := []any{start, end}
+	if contrib != "" {
+		scope = " AND l.contributor_pk IN (SELECT pk FROM dz_contributors_current WHERE code = ?)"
+	}
+	q := `WITH agg AS (
```
ben-dz, line 3247: **Minor** — Contributor scoping is applied outside the heavy CTEs — here and in `fetchDeviceAvailability`, `fetchDowntimeLinks/Devices` and `fetchOutageSummary`'s device leg the CTE aggregates every link over the window and the filter only reaches the outer join — so a scoped view costs the same as network-wide on the path that is always a live compute. Separately, `/drain` and `/deferred` each run `fetchLinkDownEvents` + `fetchStatusChanges` for both windows, duplicating four heavy queries per page load. Push the predicate into the CTE.

web/src/components/network-health-reporting-page.tsx
```diff
@@ -0,0 +1,2285 @@
...
+// multicast pub) vs the device's max_users cap.
+function DeviceSlotsPanel({ rows }: { rows: NetworkHealthDeviceSlots[] }) {
+  return (
+    <section className="rounded-lg border border-border p-4">
+      <SectionTitle title="Fullest devices (user slots)" info={DEFS.deviceSlots} />
+      <p className="text-xs text-muted-foreground mb-3">Used vs max_users.</p>
+      <ChartLegend items={[[BAR.blue, 'Unicast'], [BAR.amber, 'Multicast subscriber'], [BAR.slate, 'Multicast publisher']]} />
+      <div className="space-y-2">
+        {rows.map((d) => {
+          const denom = d.max_users || 1
```
ben-dz: **Optional** — Rows the server explicitly marks unknown render as numbers: `d.max_users || 1` draws a >100% bar labelled "5 of 0 max", and `d.util_pct ?? 0` (:1581) shows a DIA interface with `bandwidth = 0` as 0% "of 0.0 G port speed", visually identical to an idle uplink. Render unknown-denominator rows as unknown, or omit them.

web/src/components/network-health-reporting-page.tsx
```diff
@@ -0,0 +1,2285 @@
...
+// INCIDENTS and MAINTENANCE sections below, so no metric is shown twice. Same
+// visual language as HeadlineStrip/KeyFacts (StatCard grid).
+function OutageSummaryStrip({
+  summary,
+  prev,
+}: {
+  summary: NetworkHealthOutageSummary | null | undefined
+  prev?: NetworkHealthOutageSummary | null
+}) {
+  if (!summary) return null
```
ben-dz: **Optional** — `OutageSummaryStrip` returns null when `outage_summary` is null, and because that null comes from `GroupBoundary`'s child the frame doesn't render either — the "Failure summary" panel silently vanishes from the page instead of showing unavailable. Set the group error on that fetch failure and let `GroupBoundary` render it.

api/handlers/network_health.go
```diff
@@ -0,0 +1,3461 @@
...
+		defer cancel()
+		release, ok := nhAcquireLive(ctx)
+		if !ok {
+			nhWriteLiveUnavailable(w)
+			return
+		}
+		defer release()
+		w.Header().Set("X-Cache", "MISS")
+		resp := fetch(ctx, start, end, contrib)
+		w.Header().Set("Content-Type", "application/json")
```
ben-dz: **Optional** — This and the three other inline JSON-write blocks reimplement `writeJSON` (api/handlers/isis.go:212, ~116 existing call sites). Call it instead.

web/src/components/stat-card.tsx
```diff
@@ -13,9 +16,20 @@ interface StatCardProps {
...
+  // Which delta direction is "good" for this metric, so the delta color reads
+  // correctly. 'up' (default): a rise is green (e.g. throughput). 'down': a
+  // fall is green (e.g. outages). 'neutral': no good direction, delta stays
+  // muted (e.g. active links).
+  goodDirection?: 'up' | 'down' | 'neutral'
+  // Optional grade on the value itself: good=green, warn=amber, bad=red. Used
+  // for a metric that is a genuine self-fact grade on the entity's own view
+  // (e.g. a contributor's share of self-reported incidents), never a
+  // cross-entity ranking.
+  tone?: 'good' | 'warn' | 'bad'
```
ben-dz: **Optional** — New-but-unused frontend surface added by this PR: `StatCard.tone`, `LinkBarChart.refLine`, the `bare` prop threaded through four components with no call site setting it, and ~13 unreferenced `DEFS` entries. Delete.

api/handlers/network_health.go
thijsvanemmerik: Implemented as a critical/degraded split. Every failed panel is named in a new `degraded[]` array on the group payload and renders that one panel unavailable. A panel whose failure invalidates the group sets `error`, and the worker returns `refreshError` so no zeroed blob is cached. The `!scoped` gates are removed, so a contributor view reports failures too.

A blanket rule would blank the 9-panel Outages and 5-panel Capacity sections and freeze their caches whenever one sub-panel hit the code-241 cap documented in this file.

`fetchContribLinkPKs` now returns a distinct error, so a query failure no longer reads as "owns no links".

Tests: `TestNHGroupsReportFailedPanels`, `TestNHGroupErrorNotSuppressedWhenScoped`, `TestNHFetchContribLinkPKsFailureIsNotEmptyScope`.

api/handlers/network_health.go
thijsvanemmerik: Fixed. The counted sets are bounded at `end` and unparseable `created_at` is dropped rather than defaulting into current.

Coverage is a separate set, deliberately unbounded by filing time: a ticket filed after the window can still cover an in-window outage, and `computeTicketAggregates` matches on the incident interval. When a cover ticket's `start_at` does not parse, the fallback is bounded by its own `created_at`.

Undated tickets above an absolute floor and a 5% share mark the group degraded; a wholly undated fetch errors so the worker keeps the last good blob. Cache key bumped to `_v2`.

Tests: `TestNHSplitTicketsByWindow`, `TestFetchNetworkHealthTicketsDataWindowBounds`, `TestNHTicketFiledAfterWindowCoversOutage`, `TestComputeTicketAggregates_UnparseableStartAtBoundedByCreatedAt`.

api/handlers/network_health.go
thijsvanemmerik: Fixed. Both paths build the bounded context before `nhAcquireLive`, matching `serveNHGroup`. The deadline is `nhHeavyGroupDeadline` (170s), equal to the queries' `max_execution_time`. All three `nhAcquireLive` call sites are now bounded.

Tests: `TestNetworkHealthDeferredBoundsQueueWait`, `TestServeNHGroupShedsWithoutRunningQuery` (asserts the fetch closure never runs for a request shed while queued).

api/handlers/network_health.go
thijsvanemmerik: Fixed. Panel failures log WARN on the request path, and the alert moved to the worker behind a per-panel `logger.Escalator` that escalates only on a sustained streak.

A ClickHouse code-241 message is pinned as non-transient in `errors_transient_test.go`, which is what makes the downgrade necessary.

Tests: `TestNHPanelFailureLogsWarnNotError`, `TestNetworkHealthDegradedEscalation`.

api/worker/workflow.go
thijsvanemmerik: Fixed. The two heavy groups moved into their own `RefreshHeavyCaches` activity with a 240s `StartToCloseTimeout`, giving the 180s scans 60s of headroom so the single retry is reachable. Batch entries keep the 60s `defaultRefreshTimeout`.

The heavy run is skipped on the continue-as-new boundary, so a just-started scan cannot stall the following drain.

Tests: `TestPageCacheWorkflowSchedulesHeavyRefresh`, `TestPageCacheWorkflowHeavyDoesNotBlockCycles`, `TestHeavyEntriesRegistered`.

api/handlers/network_health.go
thijsvanemmerik: Fixed. The four fields are deleted from `NHHeadline` and `NHDeltas`, the overview cache key is bumped to `_v2`, and the frontend types were updated to match.

api/handlers/network_health.go
thijsvanemmerik: Fixed. The nine builders, the infix constant, `FetchActiveContributorCodes` and the nine cache-HIT branches are removed, and the `serveNHGroup` doc comment no longer claims per-contributor precompute.

api/handlers/network_health.go
thijsvanemmerik: Fixed. `computeDrainTiming` treats drained as a state, so a transition into or out of drained is counted and a soft/hard level change is neither.

`TestComputeDrainTiming` gained six subtests, including `soft_then_hard_then_activated`, `hard_then_soft_no_double_count` and `drain_still_open_across_levels`.

api/handlers/network_health.go
thijsvanemmerik: Fixed. `fetchLinkDownEvents` is rebuilt on a shared `nhOutEpisodeCTE` and `nhOutageBucketPredicate`, so the Drain panel's `outage_count` equals the Outages group's. `fetchImpactfulDowntime` moved onto the same CTE.

The device-partitioned legs share only the bucket predicate: forcing them through a link-partitioned CTE would change `devices_affected` and the device downtime ranking.

Tests: `TestNetworkHealthDrainEventsMatchOutageDefinition`, `TestNetworkHealthOutageDefinitionUsesFinal` (verified to fail when `FINAL` is removed).

web/src/components/network-health-reporting-page.tsx
thijsvanemmerik: Fixed. `sameNHWindow` compares the resolved start and end rather than `days`, since two custom ranges can share a day count, and `deriveAvailability` returns null unless they match.

The mismatch reads as loading only while a fetch is in flight, and as a dash once both queries settle: a persistently failing impactful refresh leaves a frozen blob whose window trails the overview's indefinitely, and a permanent "loading" would misrepresent that.

api/handlers/network_health.go
thijsvanemmerik: Implemented with 11 tokens rather than 7. Paging the full ticket stream returns 11 distinct non-null values: the 7 documented plus `duplicate` (25 tickets), `software` (3), `dz_managed` (1) and `human_error` (1). A 7-token allowlist would relabel 30 tickets as `other`.

Anything outside the set still buckets to `other`, so operator free text cannot reach the payload. The doc comment now points at `nhRootCauseTokens` instead of restating the list, so the two cannot drift again.

Tests: `TestNHRootCauseAllowlist` pins all 11 tokens, case and whitespace normalisation, and free text bucketing to `other`.

api/handlers/network_health.go
thijsvanemmerik: Fixed. The ops lookback is floored, an unknown `contributor` is rejected before the fetch, and the users registry is memoized for the refresh interval. The lookback-floor message now names the cause rather than reporting a generic unavailable.

api/handlers/network_health.go
thijsvanemmerik: Done, with a measurement worth recording: the pushdown is performance-neutral rather than a speedup. Scoped to the largest contributor over 30 days it measured 53ms before and 56ms after, with byte-identical results, because `link_rollup_5m` is sorted on `(bucket_ts, link_pk)` so an entity predicate cannot prune granules. The same will hold for the device-partitioned legs.

The defensible benefit is a single shared episode definition rather than reduced scan work. The duplicate `fetchLinkDownEvents` and `fetchStatusChanges` runs on `/deferred` are removed.

web/src/components/network-health-reporting-page.tsx
thijsvanemmerik: Fixed. A device with no `max_users` renders an empty meter track labelled "of no cap" instead of a bar over 100%.

DIA rows with no port speed are listed separately with their P99 Gbps rather than charted as 0%, since the server has already applied its LIMIT and a dropped row is not replaced by the next one.

web/src/components/network-health-reporting-page.tsx
thijsvanemmerik: Fixed. It renders the `PanelUnavailable` frame instead of returning null, so the section reads as unavailable rather than vanishing.

`IncidentsStrip` and `MaintenanceStrip` are unchanged: their nulls mean "no ops key configured" and "no maintenance in this window", which are not failures.

api/handlers/network_health.go
thijsvanemmerik: Fixed. The inline blocks are gone. Group responses go through `serveNHGroup`, which calls `writeJSON`. The one remaining raw write is the pre-encoded cache blob.

web/src/components/stat-card.tsx
thijsvanemmerik: Fixed. `StatCard.tone`, `LinkBarChart.refLine`, the `bare` prop and the 13 unreferenced `DEFS` entries are deleted.

`toneColorClass` is kept because `MiniStat` uses it, and the `ReferenceLine` import because `GroupedBarChart` sets `refLine={100}` at two call sites.

api/worker/workflow.go
```diff
@@ -558,5 +843,22 @@ func PageCacheWorkflow(ctx temporalworkflow.Context, iteration int, p PageCacheP
...
 	}
 
+	// Drain the outstanding heavy run before continuing as new. A pending activity
+	// is cancelled at the continue-as-new boundary, which would cut a scan
+	// mid-flight, record it as a failure, and leave both heavy blobs unwritten for
+	// that cycle. This is the only place the loop waits on it, and the run it waits
+	// on started at least one cycle back (the last iteration starts none), so it
+	// has already had a full cycle to finish.
+	if heavy != nil {
+		_ = heavy.Get(ctx, nil)
```
ben-dz, line 861: **Minor** — This drain waits on the outstanding heavy scan with no bound, so the 3-second latest-slots refresh can freeze for minutes. The last heavy run may start as late as iteration `threshold-2` (~40s of cycles before the boundary) while its budget is 240s, so `heavy.Get` blocks the single workflow coroutine — and every `RefreshLatestCaches` tick with it — for up to ~200s per continue-as-new window. That is the stall the comment above sets out to avoid; skipping the start on only the final iteration does not cover a run already in flight. Bound the drain with a timer and accept a cancelled scan, or stop starting heavy runs for the last `heavyActivityTimeout / RefreshInterval` iterations.

thijsvanemmerik: Fixed. `heavyStartDue` stops starting runs within `heavyStartLeadCycles` = ceil(heavyActivityTimeout / RefreshInterval) of the boundary (8 at the default 30s cycle), so the run the drain waits on has already resolved and `heavy.Get` no longer blocks.

Took the start-cutoff over bounding the drain with a timer so a near-boundary scan still writes its blobs instead of being cancelled every window; the drain stays a non-blocking safety wait. Lead is capped at threshold-1 so a short window still starts one run at iteration 0.

Tests: `TestHeavyStartDue` pins the lead, the last-eligible start keeping its full budget before the drain, and the short-window cap; `TestPageCacheWorkflowSchedulesHeavyRefresh` now asserts no starts inside the trailing lead.

PR #697, opened 2026-07-16
logging: sweep non-actionable ERROR logs behind transient-aware helpers

### LINKED ISSUE #696: chore: systematic sweep for non-actionable ERROR logs (on-call noise)
## Problem

We keep discovering non-actionable ERROR logs one at a time and fixing them reactively (#679, #681, #683, #688, #695). Each one paged on-call `#alerts`/`#alerts-l2` for something no human needs to act on. They fall into a small number of recognizable patterns, so a **systematic sweep** should find the rest in one pass instead of waiting for each to fire in prod.

**Goal:** no ERROR-level log for a non-actionable condition, anywhere in lake. Only genuinely-actionable failures should reach ERROR (and therefore page). WARN/INFO don't page and are the right home for transient/expected conditions.

*(Reminder on the alert mechanics: alerts fire on `level="ERR"` lines only — prod → `#alerts` (high-sev), staging → `#alerts-l2`. So "fix" = demote to WARN/retry/handle, not silence.)*

## Taxonomy of false positives (what to hunt)

Every one we've fixed maps to one of these. The sweep classifies each ERROR call site against this list:

1. **Not-found / empty result treated as error** — `sql.ErrNoRows`, Neo4j "no records", empty query results logged ERROR when they're an ordinary 404/empty. *(e.g. #695)*
2. **Transient / self-healing infra** — connection reset/refused, EOF, broken pipe, i/o timeout, DNS "server misbehaving", rate-limit (429). Self-heals on retry → WARN, or retry-then-WARN. *(`dberror.IsTransient` already encodes these; #683 applied it in api/handlers.)*
3. **Client-caused** — context canceled (client disconnect), caller deadline, malformed input → WARN/4xx, not ERROR.
4. **Expected lifecycle events** — workflow reattach / continue-as-new, worker/pod shutdown, deploy-time dependency races (Neo4j not ready at startup), Temporal activity reporting to an already-completed workflow → WARN/INFO or suppress. *(e.g. #679, #688, temporal logger demotion)*
5. **Handled fallbacks / degraded-OK** — serving stale cache, optional source unavailable, partial results → WARN/INFO. *(e.g. #679 swap-rate)*
6. **Blip-vs-sustained** — a periodic/background failure that's only actionable when *persistent* should escalate via a consecutive-failure threshold, not ERROR on the first blip. *(e.g. #677/#681/#688 escalation pattern)*

## Where the gaps are (surveyed)

| Area | ERROR call sites | Transient/not-found aware? |
|---|---|---|
| `api/handlers` | 419 `logError` | ✅ yes (#683, #695) — the model to copy |
| `indexer` | **66 raw `.Error()`** | ❌ **no** — largest uncovered surface |
| `api/worker` | 7 raw `.Error()` | partial (has `interrupted`/`dberror`) |
| `slack` | 8 raw `.Error()` | ❌ no |
| `agent` | 2 raw `.Error()` | ❌ no |

Also: **26 `sql.ErrNoRows` references** across api/indexer/admin to audit for the class-1 pattern. And `dberror` currently lives under `api/handlers/` so the indexer can't reuse it.

## Approach

**Phase 0 — Empirical ground truth (highest ROI).** Pull the last 7–14 days of true `level="ERR"` lines from Loki across all lake `service_name`s (prod + staging), dedup by message, rank by frequency. This is the list that *actually* pages. Fix these first, regardless of what static analysis says.

**Phase 1 — Static inventory.** Enumerate every ERROR/fatal call site per component (`slog.Error`, `.Log.Error`, `log.Fatal`, startup panics). Note which route through a transient-aware helper vs. log a raw `err`.

**Phase 2 — Classify** each site against the taxonomy above; flag any that can carry an ErrNoRows / transient / client-cancel / lifecycle error and log it unconditionally at ERROR.

**Phase 3 — Prioritize:** (a) actually-firing in prod (Phase 0), (b) high-frequency-potential on a prod hot path, (c) staging/low-traffic.

**Phase 4 — Fix in batches by pattern, not one at a time:**
- **Systemic lever (biggest win):** promote `dberror` to a shared package (e.g. `pkg/dberror` or `pkg/logging`) and give the indexer + worker + slack a transient-aware log helper mirroring `api/handlers.logError` (transient/not-found → WARN, actionable → ERROR). Retrofitting the 66 indexer sites onto one helper is the main lever.
- **Targeted:** ErrNoRows guards on resolve-then-404 paths; lifecycle events → WARN/INFO; fallbacks → WARN; blip-prone loops → consecutive-failure escalation.

**Phase 5 — Guardrail against regression.** Make the right thing easy and the wrong thing caught: a shared helper new code naturally reaches for, plus a lightweight convention/review-checklist (or custom lint) so a new raw `slog.Error(err)` on these classes gets flagged.

## Non-goals

- Co-tenant repos (shredder/feeds) — separate issue.
- Alert-layer plumbing — already done (#1811/#1818, level-regexp parsing).
- Suppressing genuinely-actionable errors — this is about *non-actionable* noise only.

## Deliverables

- Phase-0 ranked list of currently-firing ERR lines (prod + staging).
- A categorized findings table: call site → class → proposed fix.
- Batched PRs by pattern, with the shared transient-aware log helper as the anchor change.
- A regression guardrail (convention + optional lint).

## Acceptance

`#alerts` (prod) holds at **zero non-actionable pages** over a sustained window, and new ERROR logs for taxonomy classes 1–5 are prevented by construction rather than caught in prod.

nikw9944:
## Summary of Changes

* Systematic sweep of the 79 raw ERROR call sites in indexer, api/worker, slack, and agent. Alerts page on every `level="ERR"` line; #679–#695 fixed non-actionable pages one at a time as they fired — this closes the rest in one pass. Genuinely-actionable terminal failures (startup/shutdown errors, panics, config errors) stay ERROR.
* Anchor change: `dberror` moves to `utils/pkg/dberror`, and `utils/pkg/logger` gains two shared helpers every service now uses — transient-aware `logger.Error` (transient/disconnect-class → WARN, else ERROR; never drops a line) and `logger.Escalator` (per-key consecutive-failure counter: WARN below threshold 3 / transient 10, ERROR at/above, reset on success).
* Retrofits by pattern: indexer view refresh loops and ingest workflows, page-cache query and write legs, Slack socketmode/ChatStream, and Anthropic API calls escalate on sustained failure instead of paging on the first blip; client-disconnect, malformed-input, probe-write, optional-listener, and shutdown-cleanup sites demote to WARN.
* Regression guardrail: a "Logging Levels" convention section in CLAUDE.md (per operator decision, no lint rule).
* Fixes #696

Behavior changes worth knowing:
* Transient causes (deploys, ClickHouse restarts, rate limits — including gRPC `ResourceExhausted`) escalate at 10 consecutive failures; strict causes at 3. The two 5-minute-interval views (validatorsapp, device usage) use the strict threshold for everything so a hard-down dependency pages in ~15 minutes.
* `consecutive_failures` moved to trailing position on the worker/dzingest failure lines — key-based Loki queries are unaffected, but any panel parsing those lines positionally needs updating. Message texts are unchanged everywhere.
* Workflow-side escalator counts reset at continue-as-new (~hourly), delaying escalation by up to 2 iterations across the boundary.

## Testing Verification

* Unit tests pin `logger.Error` classification (transient/disconnect → WARN, actionable → ERROR, no line ever dropped, slog.Attr-safe arg parsing) and `Escalator` behavior (thresholds, reset, per-key independence, mixed-cause flapping, concurrent Fail/Reset under the race detector).
* `TestWriteFailureEscalation` pins that a sustained transient Postgres write outage still reaches ERROR; pre-existing `TestInterruptedEscalation` and `TestLogError_DowngradesTransientToWarn` pass unchanged, pinning the retrofits to prior behavior.
* `dberror` tests cover the word-boundary `eof` match (a "geofence" error no longer classifies transient) and `ResourceExhausted` as a rate limit.


ben-dz (commented):
Well-executed sweep: the shared helpers are correct and race-free, retrofits reproduce prior thresholds and message texts exactly (pinned by tests), redaction/Sentry paths intact, and indexer startup demotions still funnel into a terminal ERROR before shutdown. Two High findings remain, both the failure class the PR set out to prevent — a sustained outage that never pages because a one-shot logger.Error was used where an escalator was needed: (H1) page-cache writes under a hard-down Postgres log WARN forever, and (H2) a sustained Anthropic outage produces zero ERROR lines from the agent path. Mediums cover a missing concurrency test, ~25 hand-rolled Fail/Reset pairs that an Escalator.Observe helper would collapse, ~50-minute paging delay on 5-minute-interval views, an overpromising slack comment, latest-error-class threshold selection, and two missed api/main.go sites.

- M6: Two sites in api/main.go match the sweep's own criteria but were missed (declared scope was indexer/worker/slack/agent, same failure classes): api/main.go:135 logs raw slog.Error on any transient network blip during the per-request S3 asset fetch, and api/main.go:397 keeps the metrics-listener failure at ERROR (non-fatal) while the equivalent indexer site was demoted to WARN in this PR. Route the asset fetch through logger.Error and align the metrics-listener line, here or in an immediate follow-up.
- L9: indexer/pkg/dz/shreds/escrowevents/parser.go:58 (pre-existing, kept by the sweep) — "unknown instruction action" logs ERROR once per matching transaction, so a new on-chain instruction type pages repeatedly until code catches up; if intended, fine, otherwise dedupe per action or route through an escalator.

- `utils/pkg/dberror/dberror.go:76` — low: L2: `strings.Contains(errStr, "eof")` matches any embedded trigram (e.g. "geofence"); this classifier now gates paging fleet-wide. Use "unexpected eof" or word-boundary matching.
- `utils/pkg/dberror/dberror.go:111` — low: L3: The InfluxDB gRPC rate-limit (ResourceExhausted) is not string-classified transient, so it escalates at the strict threshold — yet the pre-PR comment listed InfluxDB rate limits as the transient case this mechanism protects. Add a matching pattern here if the old intent stands (the dzingest comment documents the divergence honestly).

api/worker/workflow.go
```diff
@@ -369,7 +371,9 @@ func (a *Activities) refresh(parentCtx context.Context, name, key string, fn fun
 			if parentCtx.Err() != nil {
 				return
 			}
-			a.Log.Error("cache write failed", "cache", name, "error", err)
+			// Transient-aware: a Postgres blip warns, a genuine write
+			// failure still pages.
+			logger.Error(a.Log, "cache write failed", "cache", name, "error", err)
```
ben-dz: H1: On a sustained Postgres outage this can never page. `WritePageCache` failures like `dial tcp ...: connection refused` classify transient indefinitely, so one-shot `logger.Error` emits WARN on every refresh cycle forever; the write leg never touches the escalator and `a.esc.Reset(key)` at line 363 already cleared the query counter, so no counter covers writes. On main this was a raw `a.Log.Error` that paged on the first failure, and nothing else alerts (handlers fall back silently; PageCacheWorkflow never fails since activities return nil). This contradicts the comment above ("a genuine write failure still pages"). Fix: `a.esc.Fail(a.Log, key+":write", "cache write failed", ...)` with `Reset(key+":write")` after a successful write, plus a test pinning that ≥ transientErrorAfterFailures consecutive transient write failures produce an ERROR.

agent/pkg/workflow/anthropic.go
```diff
@@ -86,7 +87,7 @@ func (c *AnthropicLLMClient) Complete(ctx context.Context, systemPrompt, userPro
 
 	duration := time.Since(start)
 	if err != nil {
-		slog.Error("Anthropic API call failed", "phase", c.name, "duration", duration, "error", err)
+		logger.Error(slog.Default(), "Anthropic API call failed", "phase", c.name, "duration", duration, "error", err)
```
ben-dz: H2: A sustained Anthropic outage now produces zero ERROR lines from the agent path (also line 258 and slack/bot/processor.go:417). Provider timeouts demote via the "context deadline exceeded" disconnect-class match and 429s/connection errors demote as transient, so a total, indefinite provider outage logs only WARN forever — and alerts fire on level="ERR" only. Fix: give AnthropicLLMClient a logger.Escalator keyed by phase (Fail on error, Reset on success), or confirm Sentry/metrics alerting pages on sustained agent failure (which downgrades this to Medium). Also resolve the divergence with api/handlers.internalError, which skips Sentry capture for transient causes while this site captures unconditionally.

utils/pkg/logger/escalate.go
```diff
@@ -0,0 +1,84 @@
...
+	// run is worth paging on.
+	DefaultTransientErrorAfter = 10
+)
+
+// Escalator tracks consecutive failures per key and logs WARN below an
+// escalation threshold, ERROR at/above it, so a periodic or background task
+// pages on-call only when its failure is sustained rather than on the first
+// blip. The zero value is ready to use with the default thresholds.
+//
+// Safe for concurrent use. Also safe inside Temporal workflow code: it holds
```
ben-dz, line 28: M1: "Safe for concurrent use" has no concurrent test, so `make test`'s race detector exercises nothing here; real callers hit it concurrently (worker errgroup, parallel dzingest futures). Add a goroutine-hammer test calling Fail/Reset over overlapping keys and asserting final counts.

utils/pkg/logger/escalate.go
```diff
@@ -0,0 +1,84 @@
...
+	mu       sync.Mutex
+	failures map[string]int
+}
+
+// Fail records a consecutive failure for key and logs msg — with args plus a
+// trailing "consecutive_failures" attribute — at WARN below the escalation
+// threshold, ERROR at/above it. The error in args (the value of the first
+// "error" key) selects the threshold: transient causes use the higher
+// TransientErrorAfter since they self-heal.
+func (e *Escalator) Fail(log Logger, key, msg string, args ...any) {
```
ben-dz, line 60: M2: ~25 call sites hand-roll the Fail/Reset pair (10 views repeat an identical 6-line safeRefresh block; dzingest adds 9 `else { esc.Reset(...) }` branches). Add `func (e *Escalator) Observe(log Logger, key, msg string, err error, args ...any)` — Reset on nil err, Fail otherwise — and each site collapses to one line.

utils/pkg/logger/escalate.go
```diff
@@ -0,0 +1,84 @@
...
+	// DefaultErrorAfter is the default number of consecutive failures of the
+	// same key after which Escalator.Fail logs at ERROR instead of WARN.
+	DefaultErrorAfter = 3
+
+	// DefaultTransientErrorAfter is the default ERROR-escalation threshold for
+	// transient causes (upstream connection blips, timeouts, rate limits; see
+	// dberror.IsTransient). It's higher than DefaultErrorAfter because a brief
+	// transient failure is self-healing and not actionable — only a sustained
+	// run is worth paging on.
+	DefaultTransientErrorAfter = 10
```
ben-dz, line 20: M3: A hard-down dependency whose errors string-classify transient (`connection refused`) pages only after 10 failures × loop interval: ~5 min for the page-cache batch, ~10 min for ingest workflows, but ~50 minutes for the 5-minute-interval views (validatorsapp, device usage), where the view is the only signal for that dependency. Consider a lower TransientErrorAfter (or wall-clock-based escalation) for long-interval loops.

slack/bot/handler.go
```diff
@@ -379,6 +381,11 @@ func (h *EventHandler) handleMessageEvent(ctx context.Context, ev *slackevents.M
 func (h *EventHandler) HandleSocketMode(ctx context.Context, client *socketmode.Client) error {
 	h.log.Info("bot running in socket mode (DMs and channel mentions, thread replies enabled)")
 
+	// connectErrs escalates consecutive connection errors: socketmode
+	// auto-reconnects, so a blip warns, but sustained failure (revoked app
+	// token, broken egress) still reaches ERROR and pages.
```
ben-dz: M4: This comment overpromises — slack-go treats invalid_auth/token_revoked as fatal inside connect(): no EventTypeConnectionError is emitted; client.Run() returns and is logged at raw ERROR in api/main.go (a different path), after which HandleSocketMode blocks on a dead channel (pre-existing gap). The escalator only ever sees recoverable errors, so fix the comment to claim "broken egress" only. The mechanics are otherwise sound: evt.Data implements error, and EventTypeConnected fires on every reconnect so Reset("connect") works.

utils/pkg/logger/escalate.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	e.failures[key]++
+	n := e.failures[key]
+	e.mu.Unlock()
+
+	threshold := e.ErrorAfter
+	if threshold <= 0 {
+		threshold = DefaultErrorAfter
+	}
+	if err := ErrorFromArgs(args); err != nil && dberror.IsTransient(err) {
```
ben-dz, line 73: M5: The threshold is chosen by the latest failure's class, not the streak's: 2 genuine failures followed by transient blips stay WARN at n=3..9 even though the strict threshold was crossed by genuine errors, deferring paging up to the transient threshold under mixed-cause flapping. Acceptable if intentional — document it and pin with a test.

utils/pkg/logger/errlog.go
```diff
@@ -0,0 +1,88 @@
...
+	// the error message.
+	msg := err.Error()
+	return strings.Contains(msg, "context canceled") ||
+		strings.Contains(msg, "context deadline exceeded")
+}
+
+// ErrorFromArgs returns the value of the first "error" key in a slog-style
+// args slice, or nil if none is present.
+func ErrorFromArgs(args []any) error {
+	for i := 0; i+1 < len(args); i += 2 {
```
ben-dz: L1: ErrorFromArgs assumes strict key/value alternation; a slog.Attr in args shifts parity and silently defeats classification (fails safe toward ERROR, i.e. over-pages). Handle slog.Attr by consuming one slot, and consider an explicit-error variant `logger.Err(log, msg, err, args...)` for new call sites.

indexer/pkg/dzingest/workflow.go
```diff
@@ -41,7 +43,13 @@ func RegisterWorkflows(w worker.Worker) {
...
 // Activity failures are logged and the workflow continues to the next iteration.
 func DZIngestWorkflow(ctx temporalworkflow.Context, iteration int) error {
-	logger := temporalworkflow.GetLogger(ctx)
+	log := temporalworkflow.GetLogger(ctx)
+
+	// Activity errors reaching the workflow are Temporal-level (StartToClose
+	// timeouts, scheduling failures) — the activity-side failure counter never
+	// sees them, and timeouts classify as transient, so escalate them at the
+	// strict threshold to keep sustained timeouts visible.
+	esc := &logger.Escalator{TransientErrorAfter: logger.DefaultErrorAfter}
```
ben-dz, line 55: L4: Workflow-side escalator counts reset at continue-as-new (~hourly), delaying escalation up to 2 iterations across the boundary. Harmless at threshold 3, but add a one-line comment so nobody later raises the threshold and silently loses escalation (same in solingest and rollup).

utils/pkg/logger/escalate.go
```diff
@@ -0,0 +1,84 @@
...
+	// ErrorAfter is the consecutive-failure count at which Fail logs ERROR.
+	// Zero means DefaultErrorAfter.
+	ErrorAfter int
+	// TransientErrorAfter is the threshold used instead of ErrorAfter when
+	// the failure's error is transient (dberror.IsTransient). Zero means
+	// DefaultTransientErrorAfter.
+	TransientErrorAfter int
+
+	mu       sync.Mutex
+	failures map[string]int
```
ben-dz, line 46: L5: Document that ErrorAfter/TransientErrorAfter must be set before first use (they're read outside the mutex in Fail) and that keys must be low-cardinality — the map only shrinks via Reset.

utils/pkg/logger/errlog.go
```diff
@@ -0,0 +1,88 @@
...
+//
+// Alerts fire on ERROR lines only, so this is the default way to log a
+// failure that can carry a transient, not-found, or client-caused error.
+// Reserve a raw log.Error call for genuinely-actionable terminal failures.
+//
+// The helper never drops a line: outside a request handler a context
+// deadline is usually a server-side timeout, which deserves a WARN. Request
+// handlers that want to skip logging entirely when the client has gone away
+// should check IsClientDisconnect first (see api/handlers.logError).
+func Error(log Logger, msg string, args ...any) {
```
ben-dz, line 35: L6: `logger.Error(nil, ...)` panics, and the doc doesn't state the edge defaults (no "error" key → ERROR; first-of-multiple wins). One doc sentence covers it; also open with "logs at ERROR or WARN" since the name reads as a level.

api/worker/workflow.go
```diff
@@ -423,22 +417,12 @@ func (a *Activities) interrupted(name, key string, cause error, shuttingDown fun
...
-		threshold = transientErrorAfterFailures
-	}
-	args := append([]any{"cache", name, "consecutive_failures", n, "error", err}, extra...)
-	if n >= threshold {
-		a.Log.Error("cache refresh failed", args...)
-	} else {
-		a.Log.Warn("cache refresh failed", args...)
-	}
+	args := append([]any{"cache", name, "error", err}, extra...)
+	a.esc.Fail(a.Log, key, "cache refresh failed", args...)
```
ben-dz, line 433: L7: `consecutive_failures` moved from mid-line to trailing position in worker/dzingest lines. Key-based Loki queries are unaffected, but worth a note in the PR description in case any saved panel parses positionally.

slack/bot/processor.go
```diff
@@ -413,7 +414,7 @@ func (p *Processor) ProcessMessage(
 	result, err := p.chatRunner.ChatStream(ctx, txt, history, sessionID, onProgress)
 	if err != nil {
 		AgentErrorsTotal.WithLabelValues("workflow", "api").Inc()
-		p.log.Error("API error", "error", err, "message_ts", ev.TimeStamp, "envelope_id", eventID)
+		logger.Error(p.log, "API error", "error", err, "message_ts", ev.TimeStamp, "envelope_id", eventID)
```
ben-dz: L8: New demotion surface — a Slack user's message drives LLM-generated SQL whose ClickHouse/Neo4j error text is substring-classified, so induced failures embedding "timeout"/"eof" tokens stay at WARN, keeping repeated malicious probing below the paging threshold. Mitigated: still logged at WARN, AgentErrorsTotal increments unconditionally, and workspace membership is an internal trust boundary. Classify on a typed/unwrapped sentinel rather than the flattened message if hardening is wanted.

api/worker/workflow.go
nikw9944: Fixed in c58e314. The write leg now escalates under its own `key+":write"` counter (`recordWriteFailure`), with `Reset` after every successful write, so a hard-down Postgres reaches ERROR at the transient threshold. `TestWriteFailureEscalation` pins that `transientErrorAfterFailures` consecutive connection-refused failures produce an ERROR and that the write key is independent of the query leg.

agent/pkg/workflow/anthropic.go
nikw9944: Fixed in c58e314. `AnthropicLLMClient` failures now escalate via a package-level `logger.Escalator` keyed by client name (`recordAPIFailure`): caller-canceled requests (`logger.IsCanceled`) warn without counting, everything else counts, and success resets — so a sustained provider outage reaches ERROR within 3–10 consecutive calls. `slack/bot/processor.go` got the same treatment (`chatErrs` escalator on the ChatStream path). The Sentry divergence is resolved too: capture now skips transient causes, matching `api/handlers.internalError`.

utils/pkg/logger/escalate.go
nikw9944: Added in 39ebef0: `TestEscalator_ConcurrentFailReset` hammers Fail/Reset from 8 goroutines over overlapping keys (race detector coverage), then asserts counting is exact afterwards.

utils/pkg/logger/escalate.go
nikw9944: Added `Escalator.Observe(log, key, msg, err, args...)` in 39ebef0 and retrofitted the views and the dzingest/solingest/rollup workflows onto it in 8ec746d (each site is now the ctx guard plus one Observe line). The worker's `recordFailure`/`interrupted` keep explicit Fail/Reset — their extra-args and deadline-sentinel shape doesn't collapse cleanly.

utils/pkg/logger/escalate.go
nikw9944: Addressed for the flagged loops in 8ec746d: the validatorsapp and device-usage views (5-minute cadence, sole signal for their dependency) now set `TransientErrorAfter` to the strict threshold, paging after ~15 minutes instead of ~50. I left wall-clock-based escalation out — threshold config covers the long-interval loops we have, and a time-based variant would need a clock (worse for workflow determinism).

slack/bot/handler.go
nikw9944: Comment fixed in c58e314 — it now claims recoverable failures (broken egress, Slack-side blips) only and notes that fatal auth failures abort `connect()` and surface through `client.Run()`'s return on a different path.

utils/pkg/logger/escalate.go
nikw9944: Intentional — documented on `Fail` and pinned by `TestEscalator_ThresholdFollowsLatestFailureClass` in 39ebef0 (two genuine failures followed by transient blips stay WARN until the transient threshold). The streak length itself proves the failure is sustained; re-deriving a per-streak class would make the level depend on failure order.

utils/pkg/logger/errlog.go
nikw9944: Fixed in 39ebef0: `ErrorFromArgs` now walks args the way slog does (string key consumes two slots, `slog.Attr` consumes one, badkey consumes one), so an Attr can't shift parity; it also finds an error carried inside an `slog.Attr`. Test added. I skipped the `logger.Err(log, msg, err, ...)` variant for now — no call site needs it and `Observe` already takes an explicit error for the loop shape.

indexer/pkg/dzingest/workflow.go
nikw9944: Comment added to all three workflow escalator declarations in 8ec746d (counts reset at continue-as-new; revisit before raising the threshold).

utils/pkg/logger/escalate.go
nikw9944: Documented in 39ebef0: thresholds must be set before first use (read without the mutex) and keys must be low-cardinality, never derived from request/user data.

utils/pkg/logger/errlog.go
nikw9944: Documented in 39ebef0: the doc now opens with "logs msg at ERROR or WARN", states that args with no "error" key log ERROR, that the first "error" key wins, and that log must be non-nil.

api/worker/workflow.go
nikw9944: Noted in the PR description: key-based queries are unaffected, but any saved panel parsing "cache refresh failed" / "activity refresh failed" lines positionally needs updating.

slack/bot/processor.go
nikw9944: No change, agreed with your mitigations: the line still logs WARN, `AgentErrorsTotal` counts unconditionally, and c58e314 adds the `chatErrs` escalator on this path, so sustained probing (or any sustained failure) now reaches ERROR at the transient threshold regardless of how individual errors classify. Typed/unwrapped-sentinel classification is the dberror hardening deliberately left out of this PR's scope; happy to take it as a follow-up.

nikw9944:
Addressed the review in three commits:

- **39ebef0** (`logger`): `Escalator.Observe` (M2), slog.Attr-safe `ErrorFromArgs` (L1), `IsCanceled`, doc fixes for latest-failure-class threshold selection / set-before-use / low-cardinality keys / ERROR-or-WARN edge cases (M5, L5, L6), concurrency hammer test (M1), mixed-cause flapping pin (M5).
- **c58e314** (`worker, agent, slack`): **H1** — cache writes escalate under `key+":write"` with a pinning test; **H2** — Anthropic API failures escalate per client name (caller cancels don't count) and Sentry capture skips transient causes to match `internalError`; slack `ChatStream` failures escalate too; socketmode comment corrected (M4).
- **8ec746d** (`indexer, api`): views and ingest workflows retrofitted onto `Observe` (M2); validatorsapp + device-usage views escalate transient causes at the strict threshold, ~15 min instead of ~50 at their 5-minute cadence (M3); continue-as-new reset comment in all three workflows (L4); **M6** — `api/main.go` S3 asset fetch routes through `logger.Error` and the metrics-listener/serve lines demote to WARN, matching the indexer.

PR description updated with the `consecutive_failures` position note (L7).

Chose not to change:
- **L8 / dberror substring classification ("eof", word boundaries)**: hardening the classifier to typed/root-cause matching is deliberately out of this PR's scope (the sweep doesn't touch classification semantics); the new `chatErrs` escalator bounds the impact on the slack path. Follow-up candidate.
- **L9 (`escrowevents/parser.go` unknown instruction action)**: kept at ERROR by design — a new on-chain instruction type means a missing decoder and data loss until fixed, and per-action dedupe state keyed by chain-derived strings is its own hazard. Will revisit if it proves noisy in practice.

Lint clean, race tests green locally (Docker-dependent suites left to CI, same as before).

ben-dz (commented):
Round-2 re-check of head 8ec746d: all previously posted findings are verified fixed in code, not just claimed. H1 — cache writes now escalate under a separate `key+":write"` counter (`recordWriteFailure`, reset after each successful write, pinned by `TestWriteFailureEscalation`). H2 — Anthropic and slack ChatStream failures now escalate via `apiFailures`/`chatErrs` escalators (caller-cancels warn without counting, success resets), and Sentry capture skips transient causes matching `api/handlers.internalError`. M1–M6 and the previous L1/L4–L8 all verified: concurrency hammer test, `Escalator.Observe` retrofit across all views and workflows, strict transient thresholds on the two 5-minute views (~15 min to page instead of ~50), corrected socketmode comment, documented+pinned latest-failure-class threshold selection, api/main.go sites swept, slog.Attr-safe `ErrorFromArgs`, doc additions, and the PR-description note on `consecutive_failures` position. `go vet` clean; `go test -race` passes on utils, api/worker, slack/bot, and agent workflow packages; Docker-dependent suites still need CI. Three Low items remain (apparently never posted last round): the `"eof"` substring over-match and missing `ResourceExhausted` rate-limit pattern in the fleet-wide `dberror` classifier, and the per-transaction ERROR on unknown escrow-event instructions.

- L3: indexer/pkg/dz/shreds/escrowevents/parser.go:58 (pre-existing, kept by the sweep) — "unknown instruction action" logs ERROR once per matching transaction, so a new on-chain instruction type pages repeatedly until code catches up; if intended, fine, otherwise dedupe per action or route through an escalator.
- Verified fixed in 39ebef0/c58e314/8ec746d (checked in code, not from replies): H1 write-leg escalation with TestWriteFailureEscalation; H2 apiFailures/chatErrs escalators plus transient-aware Sentry capture; M1 concurrency test; M2 Escalator.Observe retrofit; M3 strict transient thresholds on validatorsapp and telemetry/usage views; M4 socketmode comment; M5 documented+pinned threshold semantics; M6 api/main.go sweep; prior L1/L4–L8 (slog.Attr-safe ErrorFromArgs, continue-as-new comments, docs, PR-description note, chatErrs covering the slack demotion surface — typed-sentinel classification agreed as follow-up).
- Verification: go vet clean on utils, api/worker, slack, agent, and indexer; go test -race passes for utils, api/worker (incl. TestWriteFailureEscalation), slack/bot, and agent workflow packages. Docker-dependent suites (api/handlers, indexer view/rollup) still need CI confirmation.

- `utils/pkg/dberror/dberror.go:76` — low: L1: `strings.Contains(errStr, "eof")` matches any embedded trigram (e.g. "geofence"); this classifier gates paging fleet-wide. Use "unexpected eof" or word-boundary matching.
- `utils/pkg/dberror/dberror.go:111` — low: L2: The InfluxDB gRPC rate-limit (ResourceExhausted) is not string-classified transient, so it escalates at the strict threshold — yet the pre-PR comment listed InfluxDB rate limits as the transient case this mechanism protects. Add a matching pattern here if the old intent stands (the dzingest comment documents the divergence honestly).

nikw9944:
Round-2 Lows addressed in 6d7d20f:

- **L1 ("eof" over-match)**: the connectivity pattern now matches `eof` as a word (`\beof\b` on the lowercased message), so embedded trigrams like "geofence" can't classify an actionable error as transient. Tests cover "unexpected EOF", bare "EOF", and the geofence counter-example.
- **L2 (ResourceExhausted)**: the old intent stands — gRPC `ResourceExhausted` (and "resource exhausted") now classify as rate limits, so InfluxDB throttling gets the transient threshold like every other rate limit. The dzingest `esc` comment no longer documents a divergence.
- **L3 (unknown escrow instruction)**: dedupes per action string — a new on-chain instruction type pages once (ERROR), and every further matching transaction logs WARN until the decoder catches up. Cardinality is bounded by the escrow program's own instruction set.

Lint clean, race tests green on all touched packages; Docker-dependent suites left to CI as before.

ben-dz (approved):
Round-3 re-check of head 6d7d20f: the three Lows remaining from round 2 are verified fixed in code. L1 — `dberror` now matches "eof" with a word-boundary regex (geofence counter-example pinned by test). L2 — gRPC ResourceExhausted classifies as a rate limit, restoring the pre-PR intent for InfluxDB throttling; the worker's deadline sentinels are unaffected (neither matches the new patterns) and TestInterruptedEscalation still passes. L3 — the unknown escrow-instruction log pages once per new action and warns on repeats, with cardinality bounded by the program's instruction set. `go test -race` passes on the touched packages; Docker-dependent suites remain CI-gated as noted since round 1. All findings from rounds 1 and 2 are now resolved — no open findings.

PR #699, opened 2026-07-17
api: cut publisher-check refresh cost and stop serving zeroed totals

### LINKED ISSUE #698: Publisher-check: cut steady-state query cost and stop serving silently-zeroed totals
# Publisher-check: cut steady-state query cost and stop serving silently-zeroed totals

## Context (why)

Prod incident 2026-07-17: lake-api page-cache refreshes of `publisher_check` timed out continuously ~08:55–09:35 EDT ("context deadline exceeded"), paging #alerts (third fire of the `lake-api-errors` rule that week). Root-cause analysis found the publisher-check query chronically runs at the edge of its budgets: live probes of `GET /api/dz/publisher-check?epochs=1` (lighter than the cached `epochs=2` variant) measured **14.7s and 20.2s at a calm moment** against prod. The page-cache worker re-runs the 2-epoch variant every 30s under a 60s per-entry deadline with one retry, at concurrency 8 alongside ~27 other entries, all against a ClickHouse Cloud instance shared with heavy market-data feed ingestion — any contention tips it over. Separately, when the handler's budget expires mid-request, the endpoint returns **200 with silently zeroed totals** (observed in prod: 20.16s response, populated `publishers` array, `total_publishers: 0`).

This PR makes two surgical fixes. It deliberately does NOT try to make the query itself fast (see non-goals).

## Fix 1 (correctness): `FetchPublisherCheckData` must not serve zeroed totals

`api/handlers/publisher_check.go:328-350`: the total-network-stake query and the total-publishers/stake query each `slog.Warn` and continue on error, so a failure (typically context expiry after the main query consumed the budget) yields a 200 response with `total_network_stake` / `total_publishers` / `total_publisher_stake` all zero.

**Change:** propagate both errors (`return nil, fmt.Errorf("total network stake: %w", err)` etc.) instead of Warn-and-continue. All callers already handle an error return from `FetchPublisherCheckData`:
- REST `GetPublisherCheck` (`publisher_check.go:138-151`): transient → one retry with a fresh 20s budget, else 500. A 500 beats silently-wrong zeros.
- Page-cache refresh (`api/worker/workflow.go:182-184`): records the failure and keeps serving the previous complete cached payload.
- v1 `/edge/shreds/publishers/leaders` (`api/v1/shreds_publishers.go:72-78`) and the edge scoreboard: existing error paths.

**Required test** (the test that would have caught this): force the totals query to fail while the main query succeeds, and assert `FetchPublisherCheckData` returns an error rather than a zeroed response. It must fail against the old Warn-and-continue code. Use the repo's existing handler-test idioms (testcontainers ClickHouse) — e.g. seed the main-path tables but omit/drop the table the totals query reads so only that query errors.

## Fix 2 (load): take publisher-check off the 30-second treadmill

Two changes; no user-visible behavior or SQL-semantics change.

### 2a. Per-entry refresh cadence in the page-cache worker

`publisher_check` data changes on epoch timescales (~2 days); refreshing it every 30s makes it the heaviest recurring load on the shared instance. In the slow-batch entry definitions (`api/worker/workflow.go:127-229`), add a per-entry `everyN int` (default 1 = current behavior): an entry with `everyN=N` refreshes only on every Nth slow cycle. Set `publisher_check` to `everyN=4` (≈2 min at the default 30s interval). Maintain the cycle counter sensibly across the workflow's ContinueAsNew boundary — if carrying it is awkward, a counter that resets at continue-as-new is acceptable (worst case: one early refresh per continue-as-new); document the choice in a short comment.

Touch ONLY `publisher_check` unless you verify another entry (e.g. `shreds_rewards`) queries the same `shredder.publisher_shred_stats` table — when in doubt, leave other entries at `everyN=1`.

**Test:** worker unit test that an `everyN=4` entry refreshes on cycles 0, 4, 8 and is skipped on the others, while `everyN=1` entries refresh every cycle (model after the existing workflow tests in `api/worker`).

### 2b. Deadline + cache-first for the uncapped v1 path

`api/v1/shreds_publishers.go:72-

ben-dz:
## Summary of Changes
* Publisher-check no longer returns `200` with silently-zeroed totals when a totals query fails — the error now propagates (REST retries transient failures then `500`; the page-cache worker keeps its last complete payload; v1 returns `500`).
* Page-cache worker refreshes `publisher_check` every 4th slow cycle (~2 min) instead of every 30s, taking the heaviest recurring ClickHouse query off the 30s treadmill; its data only changes on epoch timescales.
* The uncapped v1 `/edge/shreds/publishers/leaders` path now serves default-shape mainnet requests from the page cache and bounds live fallbacks with a 20s deadline, instead of running the heavy query live and unbounded on every call.
* Fixes #698

### Context
Prod incident 2026-07-17: lake-api page-cache refreshes of `publisher_check` timed out continuously (~08:55–09:35 EDT, "context deadline exceeded"), paging #alerts (third fire of `lake-api-errors` that week). The query chronically runs at the edge of its budgets against a ClickHouse Cloud instance shared with heavy market-data ingestion, and when the handler's budget expired mid-request the endpoint returned `200` with `total_publishers: 0`. These are two surgical fixes; they deliberately do not try to make the query itself fast.

### Deploy note
`RefreshCaches`'s Temporal activity signature changes from `(ctx)` to `(ctx, cycle int)`. Across the upgrade, an in-flight `PageCacheWorkflow` scheduled by the old worker may fail the `RefreshCaches` activity until the workflow reaches its continue-as-new boundary (~30 min worst case) or a worker restart re-registers it. The page cache keeps serving the last-written payloads meanwhile, so impact is staleness only, not errors to clients. (`Start` also terminates + restarts the workflow on deploy, which resolves it immediately in the normal deploy path.)

### Follow-ups (out of scope)
* **Pre-aggregate `publisher_shred_stats`** (MV/rollup) to actually speed up the query. That belongs to the shredder pipeline's schema, which this repo does not own or migrate — the real fix for query cost, tracked separately.
* **Run the two cheap totals queries before the heavy main query** (or concurrently) so a budget-expiry failure lands before the main query burns the budget. A perf refinement beyond this PR's error-correctness scope.
* **Sanitize raw error text on the v1 500 path** (`dberror.UserMessage`) — a pre-existing pattern across `api/v1/*`, worth a dedicated sweep.
* **Staleness signal** on cached v1 responses (expose `updated_at` / max-age) — pre-existing behavior, widened slightly here.

The v1 live path is intentionally single-attempt (per the approved plan: live under the same 20s budget as REST, no retry).

## Testing Verification
* `TestFetchPublisherCheck_TotalsQueryError` — forces only the totals `QueryRow` calls to fail while the main `Query` succeeds (via a thin `driver.Conn` wrapper) and asserts an error is returned. Fails against the old warn-and-continue code, which returned a nil error with zeroed totals.
* `TestDueThisCycle` — an `everyN=4` entry refreshes on cycles 0, 4, 8 and is skipped otherwise; default entries refresh every cycle; plus a guard that every `slowBatchEveryN` key maps to a real `entries()` key (catches a rename orphaning the cadence).
* `TestV1EdgeShredsPublishers_ServedFromCache` — a default-shape request is served from a populated cache with no ClickHouse connection (distinctive sentinel values prove the live query was skipped).
* `TestV1EdgeShredsPublishers_NonDefaultShapeBypassesCache` — a `?q=` request bypasses the populated default-shape cache and returns live data (guards a wrong-data regression on the public endpoint).
* `TestFetchPublisherCheckCachedOrLive_LiveDeadline` — the live fallback runs under a deadline bounded by `publisherCheckLiveTimeout`.


api/handlers/publisher_check.go
```diff
@@ -97,6 +97,52 @@ func isDefaultPublisherCheckRequest(r *http.Request) bool {
...
+
+// fetchPublisherCheckCachedOrLive is the testable core of
+// FetchPublisherCheckCachedOrLive: fetch is injected so a test can observe the
+// deadline applied to the live path.
+func (a *API) fetchPublisherCheckCachedOrLive(
+	ctx context.Context, q string, epochsParam, slotsParam int,
+	fetch func(context.Context, string, int, int) (*PublisherCheckResponse, error),
+) (*PublisherCheckResponse, error) {
+	if isMainnet(ctx) && isDefaultPublisherCheckShape(q, epochsParam, slotsParam) {
+		if data, err := a.readPageCache(ctx, "publisher_check"); err == nil {
```
nikw9944: H1: `readPageCache` never checks `updated_at`, so if the page-cache worker stalls (the worker deliberately keeps the last payload on failure), the public v1 endpoint returns 200 with arbitrarily old data and no signal — no `X-Cache` header, no timestamp. Before this PR v1 always computed live, so a worker outage could not affect v1 correctness; the intended staleness is ~2 min but the actual bound is infinite. Fix: read `updated_at` alongside `data` and fall through to the live path when age exceeds a cap (e.g. 3× the effective refresh cadence, ~6 min); optionally expose an `as_of` field or `Age` header to v1 consumers.

api/v1/shreds_publishers.go
```diff
@@ -70,7 +70,7 @@ func registerEdgeShredsPublishers(humaAPI huma.API, api *handlers.API) {
 		Description: "Returns the status of every DoubleZero validator-publisher in the shred multicast group for a recent window (epochs or slots). Includes activated stake, leader/retransmit activity, and validator client/version info. A broader publishers endpoint (covering non-validator publishers) may be introduced later at /edge/shreds/publishers.",
 		Tags:        []string{"Edge/Shreds"},
 	}, func(ctx context.Context, input *EdgeShredsPublishersInput) (*EdgeShredsPublishersOutput, error) {
-		resp, err := api.FetchPublisherCheckData(ctx, input.Q, input.Epochs, input.Slots)
+		resp, err := api.FetchPublisherCheckCachedOrLive(ctx, input.Q, input.Epochs, input.Slots)
```
nikw9944, line 77: M1: `huma.Error500InternalServerError("failed to fetch shreds publishers", err)` serializes `err.Error()` into the public `errors[].message` (huma v2.37.3). This PR adds two new propagation sources — the totals errors at publisher_check.go:386 and :401 — which can carry internal table/column names and `dial tcp` hostnames/ports to unauthenticated clients. Since the new propagation points land here, sanitize in this PR: `huma.Error500InternalServerError(dberror.UserMessage(err))` and log the raw error server-side, as the REST handler does at publisher_check.go:195.

api/handlers/publisher_check.go
```diff
@@ -97,6 +97,52 @@ func isDefaultPublisherCheckRequest(r *http.Request) bool {
...
 }
 
+// publisherCheckLiveTimeout bounds a live (uncached) publisher-check run. Matches
+// the REST handler's per-attempt budget.
+const publisherCheckLiveTimeout = 20 * time.Second
+
+// isDefaultPublisherCheckShape reports whether the given parameters match the
+// cached default (no filter, epochs=2, no slots). Parameter-level mirror of
+// isDefaultPublisherCheckRequest for callers that don't have an *http.Request.
+func isDefaultPublisherCheckShape(q string, epochsParam, slotsParam int) bool {
```
nikw9944, line 111: M2: the default-shape definition is now encoded in four places that must stay in sync: `isDefaultPublisherCheckRequest` (:86), `isDefaultPublisherCheckShape` (:107), the worker's hardcoded `("", 2, 0)` (api/worker/workflow.go:204), and huma's `default:"2"` tag (api/v1/shreds_publishers.go:53). A change to the default epochs in one spot silently serves a mismatched cached shape. Extract a `defaultPublisherCheckEpochs` const and make `isDefaultPublisherCheckRequest` delegate to `isDefaultPublisherCheckShape` after parsing.

api/handlers/publisher_check.go
```diff
@@ -97,6 +97,52 @@ func isDefaultPublisherCheckRequest(r *http.Request) bool {
...
+	return q == "" && epochsParam == 2 && slotsParam == 0
+}
+
+// FetchPublisherCheckCachedOrLive returns publisher-check data, serving the
+// default-shape mainnet request (no filter, epochs=2, no slots) from the page
+// cache when a cached payload exists and otherwise running the query live under
+// publisherCheckLiveTimeout. It lets uncapped callers (e.g. the v1 edge endpoint)
+// avoid an unbounded live run of the heavy query on every request while reusing
+// the same cache the REST handler serves.
+func (a *API) FetchPublisherCheckCachedOrLive(ctx context.Context, q string, epochsParam, slotsParam int) (*PublisherCheckResponse, error) {
```
nikw9944, line 120: M3: `GetPublisherCheck` and `FetchPublisherCheckCachedOrLive` are parallel cache-or-live implementations with divergent semantics — REST retries transient errors with a fresh 20s budget (:185-:191), v1 does not; on cache-payload drift REST serves the raw drifted bytes while v1 warns and runs the 20s live query per request. Refactor `GetPublisherCheck` onto `fetchPublisherCheckCachedOrLive` (returning a `fromCache bool` for the `X-Cache` header) and move the transient retry into the shared live path.

api/handlers/publisher_check.go
```diff
@@ -325,13 +371,19 @@ func (a *API) FetchPublisherCheckData(ctx context.Context, q string, epochsParam
...
+	// the page-cache worker keeps its last complete payload, and other callers
+	// (v1 edge, scoreboard) return 500.
 	var totalNetworkStake int64
 	err = a.envDB(ctx).QueryRow(ctx,
 		`SELECT COALESCE(SUM(activated_stake_lamports), 0)
 		 FROM solana_vote_accounts_current
 		 WHERE epoch_vote_account = 'true' AND activated_stake_lamports > 0`).Scan(&totalNetworkStake)
 	if err != nil {
-		slog.Warn("publisher check: total network stake query failed", "error", err)
+		return nil, fmt.Errorf("total network stake: %w", err)
```
nikw9944: M4: all three queries run sequentially under one 20s budget, so a slow main query starves the totals and now converts a previously-degraded 200 into a 500 for filtered (non-default-shape) v1/REST requests with no cache fallback. The PR lists reordering as a follow-up, but running the two cheap totals queries first (or concurrently via errgroup) is small and removes the most likely production trigger of this new error path.

api/handlers/v1_shreds_publishers_test.go
```diff
@@ -283,3 +283,69 @@ func TestV1EdgeShredsPublishers_EpochsAndSlotsParams(t *testing.T) {
...
+		Epoch:               4242,
+		MaxSlot:             999999,
+		TotalNetworkStake:   123456789,
+		TotalPublishers:     7,
+		TotalPublisherStake: 42,
+		Publishers: []handlers.PublisherCheckItem{
+			{PublisherIP: "10.9.9.9", DZUserPubkey: "cached-user"},
+		},
+	}
+	require.NoError(t, api.WritePageCache(t.Context(), "publisher_check", cached))
```
nikw9944, line 323: M5: both new cache tests upsert the production key `"publisher_check"` into the package-shared test Postgres (`SetupPostgresForTest` has no per-test isolation) and never delete it. They avoid racing only because `ServedFromCache` is serial while `NonDefaultShapeBypassesCache` is `t.Parallel()` — an undocumented ordering contract; the persisted epoch-4242 sentinel will leak into any future pg-backed test that reads this key. Add `t.Cleanup` deleting the row in both tests and a comment on why `ServedFromCache` must not be parallel.

api/worker/workflow.go
```diff
@@ -111,6 +111,27 @@ type cacheEntry struct {
...
 
+// slowBatchEveryN sets a per-entry slow-refresh cadence: an entry whose key maps
+// to N refreshes only on every Nth slow cycle; keys absent from the map refresh
+// every cycle (the default). publisher_check reads shredder.publisher_shred_stats,
+// the heaviest recurring query on the shared ClickHouse, yet its data only changes
+// on epoch timescales (~2 days) — refreshing every 4th cycle (~2 min at the default
+// 30s interval) removes it from the 30s treadmill with no user-visible staleness.
+// edge_scoreboard also reads that table but backs a live-tail view, so it stays at
+// every cycle.
+var slowBatchEveryN = map[string]int{
```
nikw9944: M6: `slowBatchEveryN` is a side map keyed by strings that must match `entries()` keys — the orphan-key hazard is real enough that a guard test exists solely for it. Put the cadence on the entry itself (`cacheEntry.everyN int`), which removes both the drift failure mode and the guard test.

api/handlers/publisher_check.go
```diff
@@ -325,13 +371,19 @@ func (a *API) FetchPublisherCheckData(ctx context.Context, q string, epochsParam
 		publishers = []PublisherCheckItem{}
 	}
 
+	// Propagate errors from the totals queries rather than warn-and-continue:
+	// a failure here (typically context expiry after the main query consumed the
+	// budget) would otherwise return a 200 with silently-zeroed totals. Callers
+	// handle the error: the REST handler retries transient failures then 500s,
+	// the page-cache worker keeps its last complete payload, and other callers
```
nikw9944: L1: this comment claims the "scoreboard" calls `FetchPublisherCheckData` and returns 500; the scoreboard reads the page cache directly (edge_scoreboard.go:677) and warn-continues. The actual callers are the REST handler, the worker entry, and v1 via `FetchPublisherCheckCachedOrLive`. Fix the comment.

api/handlers/publisher_check.go
```diff
@@ -97,6 +97,52 @@ func isDefaultPublisherCheckRequest(r *http.Request) bool {
...
+}
+
+// fetchPublisherCheckCachedOrLive is the testable core of
+// FetchPublisherCheckCachedOrLive: fetch is injected so a test can observe the
+// deadline applied to the live path.
+func (a *API) fetchPublisherCheckCachedOrLive(
+	ctx context.Context, q string, epochsParam, slotsParam int,
+	fetch func(context.Context, string, int, int) (*PublisherCheckResponse, error),
+) (*PublisherCheckResponse, error) {
+	if isMainnet(ctx) && isDefaultPublisherCheckShape(q, epochsParam, slotsParam) {
```
nikw9944, line 136: L3: `q` is not trimmed here (the REST handler trims at :161), so `q=%20` bypasses the cache and forces a live run, and shape semantics diverge between the two entry points. Apply `strings.TrimSpace` before `isDefaultPublisherCheckShape`.

api/handlers/publisher_check.go
```diff
@@ -97,6 +97,52 @@ func isDefaultPublisherCheckRequest(r *http.Request) bool {
...
+				// A populated but unparseable cache entry indicates payload drift;
+				// fall through to a live query but surface it so it's diagnosable.
+				slog.Warn("publisher check: cached payload unmarshal failed", "error", uerr)
+			} else {
+				return &resp, nil
+			}
+		}
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, publisherCheckLiveTimeout)
```
nikw9944: L4: only the live path gets `publisherCheckLiveTimeout`; the cache read at :129 runs on the caller's unbounded ctx. Wrap the whole function body (or the `readPageCache` call) in the deadline.

api/handlers/publisher_check.go
```diff
@@ -97,6 +97,52 @@ func isDefaultPublisherCheckRequest(r *http.Request) bool {
...
+	defer cancel()
+	return fetch(ctx, q, epochsParam, slotsParam)
```
nikw9944: L5: a Postgres outage silently converts every default-shape request into a 20s heavy ClickHouse query. Wrap the live fallback in `singleflight` so concurrent default-shape misses share one query.

api/worker/workflow.go
```diff
@@ -111,6 +111,27 @@ type cacheEntry struct {
...
+	"publisher_check": 4,
```
nikw9944: L6: at `everyN=4`, a persistently failing refresh takes ~6 min to cross `errorAfterFailures=3` (~20 min for transient causes) versus ~90s/~5 min before — slower detection of a genuinely broken refresh. Acknowledge this in the comment or scale the threshold by cadence in `recordFailure`.

api/worker/workflow.go
```diff
@@ -471,7 +495,12 @@ func PageCacheWorkflow(ctx temporalworkflow.Context, iteration int, p PageCacheP
 	fastCtx := temporalworkflow.WithActivityOptions(ctx, fastActOpts)
 
 	for iteration < p.ContinueAsNewThreshold {
-		_ = temporalworkflow.ExecuteActivity(ctx, (*Activities).RefreshCaches).Get(ctx, nil)
+		// iteration doubles as the slow-refresh cycle counter (see slowBatchEveryN).
```
nikw9944: L7: the `RefreshCaches(ctx, cycle)` signature change is safe in practice — deploy terminates and restarts the workflow (pagecache.go:70), Go SDK replay doesn't compare activity inputs, and mixed-version workers zero-fill/drop the extra payload — but none of that is stated. Add a sentence noting the mixed-fleet behavior during rolling deploys.

api/handlers/v1_shreds_publishers_test.go
```diff
@@ -283,3 +283,69 @@ func TestV1EdgeShredsPublishers_EpochsAndSlotsParams(t *testing.T) {
...
+	require.Len(t, resp.Publishers, 1)
+	assert.Equal(t, "cached-user", resp.Publishers[0].DZUserPubkey)
+}
+
+// TestV1EdgeShredsPublishers_NonDefaultShapeBypassesCache asserts that a
+// non-default-shape request (a q filter) runs live and is NOT served the
+// default-shape cached payload. Guards against a regression in
+// isDefaultPublisherCheckShape serving unfiltered cached data for a filtered
+// request on the public endpoint.
+func TestV1EdgeShredsPublishers_NonDefaultShapeBypassesCache(t *testing.T) {
```
nikw9944, line 346: L8: no test pins that a non-mainnet v1 request (`X-DZ-Env: testnet`) bypasses the cache — the `isMainnet(ctx)` gate at publisher_check.go:128 is only exercised implicitly via the default-env path. Add a case with the header set and a populated cache.

nikw9944 (changes requested):
The three changes are correct and well-tested at the unit level: totals errors now propagate instead of yielding zeroed 200s, the cadence gate works and is pinned by tests, and the Temporal signature change is safe under the terminate-and-restart deploy model. No injection, authz, or cache-poisoning issues. Two things are worth fixing before merge: routing the public v1 endpoint through the TTL-less page cache trades a bounded live query for unbounded, unsignalled staleness during worker outages (H1), and the newly-added error propagation points leak raw ClickHouse error text on the public v1 500 path while sanitization is deferred to a follow-up (M1).

- M7 (security reviewer; pre-existing): the live-query DoS bound rests on a spoofable rate-limit key. Any client can bypass the cache with `?epochs=3` or any `q` and force the heavy query live; the new 20s deadline bounds each request (an improvement — pre-PR this path was unbounded), but the aggregate bound is the per-IP limiter, and `GetIPFromRequest` (api/handlers/auth_middleware.go:105) trusts client-controlled `X-Forwarded-For`/`X-Real-IP`. Fix: rightmost-untrusted-hop XFF parsing behind a trusted proxy, or a server-side concurrency cap (semaphore/singleflight) around live `FetchPublisherCheckData` runs.
- L2 (security reviewer): api/handlers/edge_scoreboard.go:672 — comment still says the `publisher_check` entry is "refreshed by the worker every 30s"; with `slowBatchEveryN` it is now ~2 min. Update the comment.

api/handlers/publisher_check.go
ben-dz: Fixed in `0de9684`. Added `readPageCacheWithAge` (reads `updated_at`); the cached-or-live path now falls through to live when age exceeds `publisherCheckStaleAfter` (6 min ≈ 3× the effective cadence). Pinned by `TestV1EdgeShredsPublishers_StaleCacheRunsLive`.

api/v1/shreds_publishers.go
ben-dz: Fixed in `0de9684`. Now returns `huma.Error500InternalServerError(dberror.UserMessage(err))` and logs the raw error server-side, matching the REST handler.

api/handlers/publisher_check.go
ben-dz: Fixed in `0de9684`. Extracted `DefaultPublisherCheckEpochs` as the single source of truth and removed `isDefaultPublisherCheckRequest` — the REST handler parses params then goes through the same `isDefaultPublisherCheckShape`. The worker entry uses the const; the huma `default` tag can't reference it, so it carries a keep-in-sync comment.

api/handlers/publisher_check.go
ben-dz: Fixed in `0de9684`. Both entry points now share `fetchPublisherCheckCachedOrLive`, which returns `fromCache` (REST sets `X-Cache`) and runs the transient retry inside the shared live path. Payload drift now behaves identically on both (warn + live).

api/handlers/publisher_check.go
ben-dz: Fixed in `0de9684`. The main query and the two totals queries now run concurrently via `errgroup`, removing the ordering dependency that turned tail-starvation into a 500. `TestFetchPublisherCheck_TotalsQueryError` still passes.

api/handlers/v1_shreds_publishers_test.go
ben-dz: Fixed in `0de9684`. Both cache tests register `t.Cleanup` to delete the `publisher_check` key, and I documented why `ServedFromCache` must stay serial.

api/worker/workflow.go
ben-dz: Fixed in `0de9684`. Cadence now lives on `cacheEntry.everyN`; dropped the string-keyed `slowBatchEveryN` map and its orphan-guard test.

api/handlers/publisher_check.go
ben-dz: Fixed in `0de9684`. Comment rewritten — the callers are the REST handler, the worker entry, and v1 via `FetchPublisherCheckCachedOrLive`; the scoreboard reads the cache directly and warn-continues, so it's no longer named here.

api/handlers/publisher_check.go
ben-dz: Fixed in `0de9684`. `q` is trimmed in the shared path before the shape check, so `q=%20` no longer diverges between entry points.

api/handlers/publisher_check.go
ben-dz: Fixed in `0de9684`. The shared path bounds the cache read under a deadline and runs each live attempt under `publisherCheckLiveTimeout` — nothing runs on the caller's unbounded ctx anymore.

api/handlers/publisher_check.go
ben-dz: Fixed in `0de9684`. Concurrent default-shape misses collapse via singleflight, and a semaphore additionally caps aggregate live concurrency (see the M7 reply).

api/worker/workflow.go
ben-dz: Fixed in `0de9684`. Added the tradeoff to the `publisherCheckEveryN` comment: counters advance once per due cycle, so escalation is ~4× slower (~6 min strict / ~20 min transient). Acceptable for epoch-scale data, and the cached-or-live readers cap staleness independently.

api/worker/workflow.go
ben-dz: Fixed in `0de9684`. Added a comment noting the mixed-fleet behavior during rolling deploys: Start terminates+restarts the workflow, the Go SDK doesn't compare activity inputs on replay, and a mixed-version worker zero-fills/drops the extra arg — no history divergence.

api/handlers/v1_shreds_publishers_test.go
ben-dz: Fixed in `0de9684`. Added `TestV1EdgeShredsPublishers_NonMainnetBypassesCache` — an `X-DZ-Env: testnet` request with a populated cache runs live instead of serving the mainnet payload.

ben-dz:
Thanks for the thorough review — all points addressed in `0de9684`, plus merge `72abb8f` (resolving the #697 `dberror`/logging relocation conflict) and `95163ec` (a test fix). CI is green. Per-thread replies cover the inline items; the two from the review body:

**M7 (pre-existing DoS bound on a spoofable key).** Added a server-side semaphore (`maxConcurrentPublisherCheckLive`) that caps concurrent live runs of the heavy query across all callers, independent of the client-controllable per-IP key — plus singleflight collapsing default-shape stampedes (L5). I deliberately left the `X-Forwarded-For` trust handling in `GetIPFromRequest` alone: fixing it correctly (rightmost-untrusted-hop behind a trusted proxy) changes rate-limiting for every endpoint, so it belongs in its own change. Happy to file that follow-up.

**L2.** `edge_scoreboard.go` comment updated — the `publisher_check` entry now refreshes on a slow cadence (~2 min, `publisherCheckEveryN`), not every 30s.

One heads-up: the branch had drifted into a merge conflict with `main` (#697 moved `dberror` to `utils/pkg/dberror` and reworked failure escalation), which was silently blocking PR CI. That's merged and resolved.

api/handlers/publisher_check.go
```diff
@@ -79,42 +80,140 @@ type PublisherCheckResponse struct {
...
 	}
-	if q.Get("slots") != "" {
-		return false
+
+	// Only the default shape stampedes (a Postgres outage sends every default
+	// request live at once); collapse those. Filtered shapes are varied and
+	// low-volume, so singleflight would rarely help and could mask distinct work.
+	if isDefaultPublisherCheckShape(q, epochsParam, slotsParam) {
+		key := string(EnvFromContext(ctx))
+		v, err, _ := a.pubCheckSF.Do(key, func() (any, error) { return run() })
```
nikw9944: M1 (new): `run` closes over the first caller's `ctx`, so if the singleflight winner's request is canceled (client disconnect, earlier deadline), the shared query fails with `context.Canceled` and every collapsed waiter gets that error — one unrelated disconnect turns into 500s for all concurrent default-shape callers, in exactly the cache-miss stampede this path protects against. Fix: run the shared fetch on a context detached from the winner — `context.WithoutCancel(ctx)` (values like the env survive) plus its own `publisherCheckLiveTimeout` so a detached run can't hold a semaphore slot forever — or use `DoChan` and let each caller select on its own ctx.

api/handlers/publisher_check.go
```diff
@@ -79,42 +80,140 @@ type PublisherCheckResponse struct {
...
-			_, _ = w.Write(data)
-			return
-		}
+func (a *API) fetchPublisherCheckLiveGuarded(
+	ctx context.Context, q string, epochsParam, slotsParam int,
+	fetch func(context.Context, string, int, int) (*PublisherCheckResponse, error),
+) (*PublisherCheckResponse, error) {
+	sem := a.publisherCheckLiveSem()
+	select {
+	case sem <- struct{}{}:
```
nikw9944, line 213: L1 (new): the semaphore acquire waits on the caller's ctx, which for REST/huma requests may carry no deadline — the per-attempt `publisherCheckLiveTimeout` is only applied after acquisition — so a burst of distinct filtered requests parks goroutines indefinitely. Bound the wait (e.g. `context.WithTimeout(ctx, publisherCheckLiveTimeout)` around the acquire) so queued requests fail fast.

nikw9944 (approved):
All 15 findings from the first review round are resolved and verified in code at 95163ec: the staleness cap (`readPageCacheWithAge` + 6-min `publisherCheckStaleAfter`), v1 error sanitization via `dberror.UserMessage`, the shared cache-or-live path with `fromCache`, concurrent totals via errgroup, entry-level `everyN` cadence, server-side live-concurrency semaphore, and all test-hygiene items — each pinned by new tests, CI green. The response commit introduces one new medium: the singleflight added for stampede protection runs the shared query under the winning caller's context, so one client disconnect can 500 every collapsed waiter — exactly in the cache-miss scenarios it protects against. Two low notes remain (unbounded semaphore queueing; the pre-existing spoofable X-Forwarded-For rate-limit key, deferred by the author to a dedicated follow-up).

- All 15 prior findings (H1, M1–M7, L1–L8) verified resolved in code at 95163ec — staleness cap with `TestV1EdgeShredsPublishers_StaleCacheRunsLive`, sanitized v1 errors, single `DefaultPublisherCheckEpochs` source of truth, shared cache-or-live path, concurrent totals, `t.Cleanup` + documented serial/parallel contract in the pg-backed cache tests, `cacheEntry.everyN`, server-side concurrency cap, corrected comments, trimmed `q`, bounded cache read, documented escalation tradeoff and rolling-deploy safety, and the non-mainnet bypass test. The 95163ec WaitGroup test fix is sound.
- L2 (carried, pre-existing): `GetIPFromRequest` (api/handlers/auth_middleware.go) still trusts client-controlled `X-Forwarded-For`/`X-Real-IP` as the rate-limit key. The new `maxConcurrentPublisherCheckLive` semaphore now bounds this endpoint's aggregate query load independent of that key, so the residual risk is request queueing rather than ClickHouse load. The author commits to a dedicated follow-up for the XFF trust handling; keep visibility until that lands.

api/handlers/publisher_check.go
ben-dz: Fixed in `59f8836`. The collapsed default-shape run now executes on `context.WithoutCancel(ctx)` (keeps the env value `envDB` routes on) with its own `publisherCheckLiveTimeout`, so a winner's disconnect can't cancel the shared query, and callers wait via `DoChan` and select on their own ctx. Pinned by `TestFetchPublisherCheckLive_DefaultShapeDetachedFromCaller` (fetch ctx stays live after the caller cancels).

api/handlers/publisher_check.go
ben-dz: Fixed in `59f8836`. The semaphore acquire now waits on `context.WithTimeout(ctx, publisherCheckLiveTimeout)` rather than the caller's raw ctx, so a burst of distinct filtered requests fails fast instead of parking on a deadline-less request. Pinned by `TestFetchPublisherCheckLive_AcquireBounded` (a done ctx returns promptly on a saturated semaphore, without running the fetch).

ben-dz:
Thanks for the approve. Both new findings from this round are fixed in `59f8836`, CI green:

- **M1 (new):** the singleflight-collapsed default-shape run no longer inherits the winner's context — it runs on `context.WithoutCancel(ctx)` with its own deadline, and callers wait via `DoChan` selecting on their own ctx. A winner disconnect can no longer 500 the collapsed waiters.
- **L1 (new):** the semaphore acquire is now bounded by `publisherCheckLiveTimeout`, so a burst of distinct filtered requests fails fast instead of parking goroutines on a deadline-less caller ctx.

Both are pinned by new deterministic tests (`..._DefaultShapeDetachedFromCaller`, `..._AcquireBounded`).

The remaining low — the pre-existing spoofable `X-Forwarded-For` rate-limit key in `GetIPFromRequest` — I'm still leaving for a dedicated follow-up, since fixing it right (rightmost-untrusted-hop behind a trusted proxy) changes rate limiting for every endpoint. Happy to file that issue.

PR #701, opened 2026-07-20
api: move tiredsolid scoreboard feed config to postgres

armcconnell:
## Summary

- Moves the tiredsolid scoreboard's feed list out of hardcoded Go slices and into a Postgres config table (`feed`, `label`, `display_order`, `enabled`). Adding, removing, or reordering a feed is now a row change instead of a deploy.
- Collapses the previous two lists (shown set + withheld set) into a single allow-list. Feeds with no enabled row no longer surface anywhere, including the recent-races strip.
- Config load failures now propagate instead of quietly caching an empty payload, so the refresher keeps serving the last good result and escalates.
- Adds an empty state so an unseeded environment renders nothing rather than a 0.0% headline.
- Redacts vendor names from two existing design docs.

The migration ships the table **empty on purpose** — rows are environment config, seeded out of band, so values never live in the repo. Until an environment is seeded its scoreboard renders empty. Changes take effect on the next cache refresh (~60s for the 1h view, ~10min for 24h/7d), no restart.

## Testing Verification

- `go test ./api/handlers/` green, 0 failures.
- New tests cover allow-list filtering, display order, disabled rows, unlisted feeds staying hidden, malformed feed ids being skipped, config-load failure surfacing as an error, and the empty-config path.
- Confirmed the headline math is unchanged: totals previously summed only the configured set, so moving the restriction into SQL narrows the scan without moving the math.

### DIFF
--- a/api/config/migrations/00016_add_hyperliquid_scoreboard_entry.sql
+++ b/api/config/migrations/00016_add_hyperliquid_scoreboard_entry.sql
@@ -1,0 +1,14 @@
+-- +goose Up
+-- Scoreboard feed allow-list. Only enabled rows are raced, counted, or displayed on the
+-- Hyperliquid scoreboard. Intentionally seeded with no rows: the rows are environment
+-- config, inserted out of band, so they never live in this repository.
+CREATE TABLE hyperliquid_scoreboard_entry (
+    feed          TEXT PRIMARY KEY,
+    label         TEXT NOT NULL,
+    display_order INT  NOT NULL DEFAULT 0,
+    enabled       BOOLEAN NOT NULL DEFAULT TRUE,
+    updated_at    TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+-- +goose Down
+DROP TABLE IF EXISTS hyperliquid_scoreboard_entry;
--- a/api/config/migrations/00016_add_hyperliquid_scoreboard_entry.sql
+++ b/api/config/migrations/00016_add_hyperliquid_scoreboard_entry.sql
@@ -4,7 +4,7 @@
 -- config, inserted out of band, so they never live in this repository.
 CREATE TABLE hyperliquid_scoreboard_entry (
     feed          TEXT PRIMARY KEY,
-    label         TEXT NOT NULL,
+    label         TEXT NOT NULL CHECK (length(label) BETWEEN 1 AND 64),
     display_order INT  NOT NULL DEFAULT 0,
     enabled       BOOLEAN NOT NULL DEFAULT TRUE,
     updated_at    TIMESTAMPTZ NOT NULL DEFAULT NOW()
--- a/api/handlers/hyperliquid_scoreboard.go
+++ b/api/handlers/hyperliquid_scoreboard.go
@@ -103,37 +103,6 @@ func (a *API) loadHyperliquidScoreboardEntries(ctx context.Context) hyperliquidE
 	return e
 }
 
-// hyperliquidCompetitors lists the non-DoubleZero feeds shown on the scoreboard,
-// in display order, mapping each raw feed name to its label.
-var hyperliquidCompetitors = []struct{ Feed, Label string }{
-	{"hyperliquid_public_bbo", "Public API"},
-	{"hydromancer_bbo", "Hydromancer"},
-	{"dwellir_l2book_bbo", "Dwellir"},
-	{"quicknode_l2book_bbo", "QuickNode"},
-}
-
-// hyperliquidExcludedFeeds are competitor feeds withheld from every scoreboard measurement
-// (per-competitor, per-node, headline totals, and recent races) because their upstream data is
-// currently unreliable. hyperpc_shared_bbo (HypeRPC) arrives a median ~100s stale during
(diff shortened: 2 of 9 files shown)


packethog (approved):

ben-dz (commented):
Two mediums worth addressing before or shortly after merge: the live request path now returns a 500 during a Postgres outage where it previously served from ClickHouse (config was in code, and the page cache is also in Postgres, so there is no fallback), and the DZ-lost arm of the unconfigured-feed exclusion clause has no test. Two lows: the 500 body now carries raw Postgres error text (swap in the existing internalError() helper), and the new table should attach the existing updated_at trigger. The sanitizer, allow-list clause semantics, and last-good cache preservation on config-load failure all check out.

- Low — api/handlers/hyperliquid_scoreboard.go:736: the 500 body returns raw err.Error(), and this PR routes the first Postgres-sourced errors into it — pgx connect errors embed internal detail (failed to connect to host=… user=… database=…), so during a Pg incident an uncached request (e.g. ?symbol=BTC) discloses infra names to anonymous clients. The repo already has internalError() (api/handlers/errors.go:45) for exactly this; swap it in.

api/handlers/hyperliquid_scoreboard.go
```diff
@@ -209,23 +275,30 @@ func (a *API) FetchHyperliquidScoreboardData(ctx context.Context, window, symbol
...
+
+	// The configured feed allow-list drives which feeds are raced, counted, and labelled.
+	// A load failure propagates as an error so the caller (page-cache refresher, request
+	// handler) does not mistake it for "nothing configured" and overwrite a last-good cached
+	// payload with an empty one. Zero configured rows is not an error — with nothing
+	// configured there is no scoreboard to compute, so serve the empty-but-valid payload
+	// rather than scanning for feeds we would then discard.
+	entries, err := a.loadHyperliquidScoreboardEntries(ctx)
+	if err != nil {
+		return nil, err
```
ben-dz, line 289: The live request path now hard-fails during a Postgres outage where it previously served. Before this PR a Pg outage made readPageCache miss and the handler still served the scoreboard live from ClickHouse (config was in code); now the live path dies here with a 500 — and there is no cached fallback either, because the page cache also lives in Postgres. Other pages degrade to live-serving during a Pg outage; this one now 500s. The propagate-don't-blank choice is right for the refreshers, but if scoreboard uptime through Pg maintenance matters, consider a process-local last-good copy of hyperliquidEntries (e.g. an atomic.Pointer on API, updated on every successful load) as a fallback when the load fails.

api/handlers/hyperliquid_scoreboard_test.go
```diff
@@ -191,16 +216,180 @@ func TestHyperliquidScoreboard_HeadlineAndCompetitors(t *testing.T) {
...
+
+// A feed with no config row at all never surfaces — including in the recent-races strip,
+// which previously could show an unlisted feed under its raw name.
+func TestHyperliquidScoreboard_UnconfiguredFeedNeverSurfaces(t *testing.T) {
+	api := apitesting.NewTestAPIBarePg(t, testChDB, testPgDB)
+	createFeedsTable(t, api)
+	insertEntry(t, api, "feed_a_bbo", "Feed A", 1, true)
+
+	insertPairwise(t, api, "tyo-rec1", "tyo", "BTC", 10, 1, "tob_gcp_tyo_hl_mainnet1", "feed_a_bbo", 1.0)
+	insertPairwise(t, api, "tyo-rec1", "tyo", "ETH", 20, 2, "tob_gcp_tyo_hl_mainnet1", "feed_unlisted_bbo", 1.0)
```
ben-dz, line 295: The DZ-lost arm of the unconfigured-feed exclusion is unpinned: this test (and TobFeedEntryIgnored) only inserts winner=tob_*, loser=feed_unlisted_bbo. The reverse row (feed=feed_unlisted_bbo, loser_feed=tob_* — an unconfigured competitor beating DZ) exercises the loser_feed IN (…) side of the OR and is excluded by the code but asserted nowhere. One extra insertPairwise pins the clause's other arm.

api/config/migrations/00016_add_hyperliquid_scoreboard_entry.sql
```diff
@@ -0,0 +1,14 @@
+-- +goose Up
+-- Scoreboard feed allow-list. Only enabled rows are raced, counted, or displayed on the
+-- Hyperliquid scoreboard. Intentionally seeded with no rows: the rows are environment
+-- config, inserted out of band, so they never live in this repository.
+CREATE TABLE hyperliquid_scoreboard_entry (
+    feed          TEXT PRIMARY KEY,
+    label         TEXT NOT NULL CHECK (length(label) BETWEEN 1 AND 64),
+    display_order INT  NOT NULL DEFAULT 0,
+    enabled       BOOLEAN NOT NULL DEFAULT TRUE,
+    updated_at    TIMESTAMPTZ NOT NULL DEFAULT NOW()
```
ben-dz, line 10: updated_at freshness relies on operators remembering `updated_at = NOW()` in the runbook UPDATE. The repo's update_updated_at_column() trigger function (migrations 00001/00003) is attached to sibling tables for this; attach it here so a hand-typed UPDATE can't leave a stale timestamp.

PR #703, opened 2026-07-22
indexer: publish serviceability feeds to ClickHouse

nikw9944:
## Summary of Changes

* Publish the Generalized Payments **feed** objects to ClickHouse via the serviceability view: a new `dz_feeds_current` dimension (feed SKU → metro + joinable multicast groups), `dz_users_current.feed_pks` (JSON array of the EdgeSeat feeds whose per-feed seats a connected user consumed — a user may hold seats on multiple feeds), and `dz_access_passes_current.feed_seats` (per-feed seat billing lifecycle on EdgeSeat passes).
* Serviceability added these onchain (`Feed` account, `User.FeedPks`, `AccessPass.FeedSeats`); lake pinned an SDK from before they existed, so this also bumps the doublezero SDK to the current `main` pseudo-version, which includes the `feed_pk` → vec-of-`feed_pks` change (doublezero#4080).
* New entity follows the `dz_access_passes` template; new columns follow the recent `bgp_status` template — appended **last** in `PayloadColumns()` and placed `AFTER` the prior last column to satisfy the positional staging-insert contract. Three goose migrations, one concern each.
* Conventions kept uniform: onchain "exchange" → `metro_pk` (matches `dz_devices`); variable-length lists stored as JSON strings defaulting to `'[]'` (`groups`, `feed_pks`, `feed_seats` — matches `publishers`/`mgroup_*_allowlist`), with nil slices normalized to `[]` (never `null`) in the schema layer.
* Rollout: `feed_pks` and `feed_seats` are hashed payload columns, so the first refresh after the migration rewrites one history row per existing user and access pass (same accepted one-time churn as the `bgp_status` rollout). Purely additive; goose `Down` migrations restore the prior views and drop the columns.
* Fixes malbeclabs/infra#2068

## Testing Verification

* Unit tests over the converters: `convertFeeds` (base58 of pubkeys, JSON `groups` including empty → `[]`), `convertUsers` (multi-feed vec → base58 array, empty vec for non-EdgeSeat users), `convertAccessPasses` (exact `feed_seats` JSON including no-seats → `[]`).
* Integration tests (testcontainers ClickHouse): `ReplaceFeeds` round-trip (insert, rename-update, tombstone-on-missing), `dz_users` round-trip asserting the `feed_pks` JSON array and nil → `[]` persistence, `dz_access_passes` round-trip asserting `feed_seats` JSON and empty → `[]`, and the `Refresh` "stores all data" fixture extended with a feed, a user carrying `FeedPks`, and an EdgeSeat pass carrying `FeedSeats`, asserted end-to-end through the `_current` views.
* Evals: `SQL_CONTEXT.md` is agent context, and CI's `evals-short` runs the eval suite in `-short` mode, where the live-model cases skip (it verifies the prompt still embeds and the eval package compiles). To cover model behavior with the new feeds paragraph, the filtered live-model SQL-generation evals (`go test -tags evals -run 'TestLake_Agent_Evals_Anthropic_SQLGeneration'`) were run locally against the composed `GENERATE.md` + `SQL_CONTEXT.md` prompt.
* The container-backed tests cannot run in the build sandbox (no Docker), so local verification was limited to building all packages and vet/gofmt after the SDK bump; the container tests run in CI. Post-merge, verify on staging via the data MCP that `dz_feeds_current` populates and `feed_pks`/`feed_seats` carry data on a feed-bearing environment.


indexer/pkg/dz/serviceability/schema.go
```diff
@@ -493,13 +501,51 @@ func (s *AccessPassSchema) ToRow(ap AccessPass) []any {
...
+		"owner_pubkey:VARCHAR",
+		"code:VARCHAR",
+		"name:VARCHAR",
+		"metro_pk:VARCHAR",
+		"groups:VARCHAR",
+	}
+}
+
+func (s *FeedSchema) ToRow(f Feed) []any {
+	groupsJSON, _ := json.Marshal(f.Groups)
```
nikw9944: **M1**: `FeedSchema.ToRow` marshals `f.Groups` without the nil→`[]` guard that this same PR (abe12fc) added to `AccessPassSchema.ToRow` for `FeedSeats` (schema.go:485). A nil `Groups` persists the literal string `null` into `dz_feeds_current.groups`, violating the migration's documented "JSON array of base58 multicast-group PKs" contract and breaking `JSONExtract`/array queries. Latent today — `convertFeeds` always passes a non-nil slice — but `Store.ReplaceFeeds` is exported and abe12fc deliberately placed the equivalent guard in `ToRow` so the schema layer guarantees the shape regardless of caller. Apply the same guard here: `if f.Groups == nil { ... []string{} }`.

indexer/pkg/dz/serviceability/view.go
```diff
@@ -561,6 +596,36 @@ func convertUsers(onchain []serviceability.User) []User {
...
 			BgpStatus:   bgpStatusString(user.BgpStatus),
+			FeedPK:      pubkeyOrEmpty(user.FeedPk),
+		}
+	}
+	return result
+}
+
+// pubkeyOrEmpty returns the base58 pubkey, or "" for the zero pubkey. The
+// serviceability SDK uses the zero pubkey to mean "no value" (e.g. User.FeedPk
+// on non-EdgeSeat/unicast users), and "" keeps `WHERE col != ”` filters sane.
```
nikw9944: **L1**: This comment contains a Unicode right double-quote (U+201D): `WHERE col != ”` should be `WHERE col != ''` (two ASCII single quotes). The PR description lists this typo as fixed, but it's still present — the only U+201D in the changed files.

nikw9944 (commented):
Two Mediums and one Low; nothing blocking.

**M1**: `FeedSchema.ToRow` marshals `f.Groups` without the nil→`[]` guard this PR's own abe12fc added to `AccessPassSchema.ToRow` for `feed_seats` — a nil `Groups` persists the literal string `null` into `dz_feeds_current.groups`. Latent today (convertFeeds always passes non-nil), but the same three-line guard closes it.

**M2**: The deferred `entity_changes_v` `feed_pk` gap is a reasonable follow-up (matches the `20260709000001` bgp_status precedent), but please file and link the follow-up issue before merge so it actually ships.

**L1**: The Unicode-quote typo in the `pubkeyOrEmpty` comment (view.go:607) that the PR description lists as fixed is still present.

- **M2** (medium): The deferred `entity_changes_v` update for `feed_pk` is confirmed real — the users branch fires on any `attrs_hash` change (which now includes `feed_pk`) but `changed_fields` only enumerates status/kind/client_ip/dz_ip/device/tunnel_id/bgp_status, so `feed_pk`-only transitions render as contentless "updated (0 fields)" timeline events. Deferring matches the `20260709000001` bgp_status precedent and the published data is correct, so keeping it out of this PR is fine — but that precedent worked because the follow-up shipped. Please file the follow-up issue (add `feed_pk` to the users `changed_fields` array; consider whether `dz_feeds` itself belongs in the view) and link it from this PR before merge.

indexer/pkg/dz/serviceability/schema.go
nikw9944: Fixed in bd9fda6 — `FeedSchema.ToRow` now normalizes nil `Groups` to `[]string{}` before marshaling (schema.go:534), same pattern as `AccessPassSchema.ToRow` for `FeedSeats`. Covered by a unit test asserting `groups` renders as `[]` for a nil slice.

indexer/pkg/dz/serviceability/view.go
nikw9944: Fixed in bd9fda6 (comment rewritten in a435f68 so gofmt stops reintroducing it) — the `pubkeyOrEmpty` comment now uses ASCII quotes. Verified with a grep over all changed files: no U+201D remains.

nikw9944:
Re **M2**: confirmed — the users branch of `entity_changes_v` fires on any `attrs_hash` change but `changed_fields` doesn't enumerate `feed_pk`, so a `feed_pk`-only transition renders as "updated (0 fields)". Deferring per the `20260709000001` bgp_status precedent.

Follow-up to file and link before merge (filing is pending operator action):

> **indexer: add feed_pk to entity_changes_v users changed_fields**
> Recreate `entity_changes_v` with `feed_pk` added to the users `changed_fields` comparison (lag window + `if(prev != cur, 'feed_pk', '')`), following `20260709000001_entity_changes_bgp_status.sql`. While in there, decide whether `dz_feeds` belongs in `entity_changes_v` as its own entity section (feeds emit no timeline events at all today).

**M1** and **L1** are already fixed on the branch (bd9fda6); replied on the threads with specifics.

agent/pkg/workflow/prompts/SQL_CONTEXT.md
```diff
@@ -485,6 +485,8 @@ The `dz_multicast_groups_current` table stores multicast group metadata:
 - `subscribers` column: `["group_pk1","group_pk2"]` — groups this user subscribes to
 - Multicast users have `kind = 'multicast'`
 
+**Feeds (SKUs)** map a feed to a metro and its joinable multicast groups: `dz_feeds_current` has `pk`, `code`, `name`, `metro_pk` (joins `dz_metros_current.pk`), and `groups` (JSON array of multicast-group PKs). `dz_users_current.feed_pks` links a connected EdgeSeat user to its feeds (JSON array of feed PKs; a user may hold seats on multiple feeds) and is `'[]'` for non-EdgeSeat/unicast users, so filter unset feeds with `feed_pks != '[]'` and expand with `JSONExtract(feed_pks, 'Array(String)')`; `dz_access_passes_current.feed_seats` is a JSON array of per-feed seat billing (`feed_pk`, `max_users`, `current_users`, `window_end`, `terminates_at`, ...) on EdgeSeat passes.
```
ben-dz, line 488: The `feed_seats` field list doesn't state that `window_end`/`terminates_at` are unix seconds, so the agent has to guess the unit when wrapping them in `toDateTime`/`fromUnixTimestamp`. Add "(unix seconds)" after the two field names. Fine to fold into the `entity_changes_v` follow-up.

ben-dz (approved):
The code looks good — nothing in the diff should block. The mechanical contracts all hold: the positional staging-insert order is correct for all three tables (`subscribers`→`bgp_status`→`feed_pks` and `flags`→`feed_seats` are each physically last, and the new feeds tables match `PayloadColumns()` order), both recreated views are their prior definition plus one column with nothing dropped and no `SELECT *` consumers, the Down migrations drop the column from history *and* staging, and the SDK bump doesn't shift any already-published field (`FeedSeats` sits mid-struct in `AccessPass` but is only read under the `EdgeSeat` tag). The ClickHouse-backed tests flagged as unverified in the description did run in CI — `TestMain` hard-exits when the testcontainer can't start, so there's no silent-skip path.

Two non-code items to close before merging:

1. **Update the PR description.** It still says `feed_pk` in 8 places and `feed_pks` in none. The rollout note points an operator at a column that doesn't exist, and the "One High deferred for operator decision" paragraph — the decision record for the `entity_changes_v` deferral — names the wrong column.
2. **File the `entity_changes_v` follow-up issue.** It still doesn't exist (no match in `malbeclabs/infra` or `malbeclabs/lake` beyond #2068). The deferral matches the `20260708000001` → `20260709000001` bgp_status precedent, but that precedent worked because the follow-up shipped.

One nit worth folding into that follow-up: `SQL_CONTEXT.md:488` doesn't state that `feed_seats.window_end`/`terminates_at` are unix seconds, so the agent has to guess the unit when wrapping them in `toDateTime`.

- Before merging: the description still says `feed_pk` in 8 places and `feed_pks` in none. Beyond the mismatch already noted, the rollout note asks an operator to verify a column that doesn't exist, and the "One High deferred for operator decision" paragraph — the decision record for the `entity_changes_v` deferral — names the wrong column. Worth fixing since the body may become the merge commit message.
- Before merging: the `entity_changes_v` follow-up issue still doesn't exist — no open or closed match for `entity_changes feed_pks` in `malbeclabs/infra` or `malbeclabs/lake`, only #2068. One detail to capture when filing it: the contentless "updated (0 fields)" event fires once per *existing* user at rollout, not only on later `feed_pks`-only transitions, because the `latest` CTE compares against the stored `attrs_hash` (`indexer/pkg/clickhouse/dataset/dim_write.go:676`) rather than recomputing it from the now-defaulted column.

Jared-dz:
Schema sanity check: the load-bearing bits hold up. I traced every ADD/DROP COLUMN ever applied to the users and access-passes tables on main — `bgp_status` and `flags` are physically last, so the `AFTER` placements match the appended `PayloadColumns()` order. The July 22 timestamps landing behind main's July 23 migration are fine because the goose provider runs with `WithAllowOutofOrder(true)`. Also ran the container-backed tests you deferred to CI: full serviceability suite passes with `-race` against a real ClickHouse locally, including the `ReplaceFeeds` round-trip and the `feed_pks`/`feed_seats` persistence tests.

One question: `SQL_CONTEXT.md` is agent context, and the testing section doesn't mention evals. The change is additive doc so risk is low, but did a short-mode or filtered eval run happen?

nikw9944:
@Jared-dz Good catch — CI's `evals-short` alone doesn't cover this. In `-short` mode the live-model SQL-generation evals skip themselves (`sql_generation_literal_test.go:24`), so the check only proves the eval package compiles and `SQL_CONTEXT.md` still embeds via `prompts.PromptsFS`.

A filtered live-model run has now been done against this branch's head (1af7a57):

```
go test -tags evals -run 'TestLake_Agent_Evals_Anthropic_SQLGeneration' -v ./agent/evals/
```

All 5 cases pass (2 literal-generation, 3 preserve-query), each composing the production `GENERATE.md` prompt with the updated `SQL_CONTEXT.md` — including the new feeds/`feed_pks`/`feed_seats` paragraph — against live Haiku. The doc's data-shape claims (`'[]'` default, `feed_pks != '[]'`, `JSONExtract(feed_pks, 'Array(String)')`) are separately asserted against real ClickHouse by the integration tests.

Also added an evals bullet to the PR body's Testing Verification describing this.

PR #704, opened 2026-07-22
indexer: fix permission-events refresh timeout poison loop; classify InfluxDB throttle as transient

ben-dz:
## Summary

- **Break the permission-events poison loop.** Steady-state `RefreshPermissionEvents` accumulated every watched account's events into one end-of-cycle `InsertEvents`. The per-account cursor is derived from the fact table (`max(slot)` per `permission_pk`), so it only advances when events are persisted — and nothing persisted until that single terminal insert. When a cycle's fetch+insert exceeded the 5-minute Temporal activity `StartToCloseTimeout`, the insert was cancelled, no cursor advanced, and the next cycle re-did identical work: a permanent poison loop. Now each account is inserted as soon as it finishes fetching, so a mid-cycle timeout still advances the completed accounts and the backlog drains instead of restarting every cycle. This mirrors the checkpoint-per-chunk pattern `BackfillRefresh` already uses.
- **Classify InfluxDB Cloud throttle signatures as transient.** `dberror.Classify` matched the singular `"resource exhausted"` but not InfluxDB's plural `"Resources exhausted"` (`strings.Contains` won't match across the extra `s`), and had no `"request too large"` pattern — so `RefreshTelemetryUsage`'s recurring `request too large: … Resources exhausted: Heap exhausted` classified as Unknown/non-transient, contradicting the documented intent that InfluxDB resource-exhaustion is transient (escalate at 10, not 3). Added both patterns with tests.

## Context

On 2026-07-22 the prod **Lake Indexer: Errors** alert flapped ~hourly (21:21–23:01 UTC). Root cause was `lake_testnet` / `RefreshPermissionEvents` timing out every ~300s with `context deadline exceeded` on the batch insert (14 failures over ~100 min), while mainnet-beta was healthy. The first change fixes that outage-adjacent behavior; the second is an independent correctness cleanup surfaced during the investigation.

A third finding — that testnet/devnet secondary-indexer errors page the **prod** on-call because the alert has no network dimension — is tracked separately in `malbeclabs/infra` (alert-scoping change): https://github.com/malbeclabs/infra/issues/2071

## Testing Verification

- Local build/test could not run in the agent environment (no Go toolchain present); CI must exercise build/lint/test.
- `dberror` change is covered by new table cases asserting `Classify` → `ErrorTypeRateLimit` and `IsTransient` → true for both `"Resources exhausted: Heap exhausted"` and the full `request too large: …` message.
- Behavior of the permission-events refresh is otherwise unchanged: all accounts are still attempted per cycle (plain `errgroup`, no early cancel), partially-decoded accounts still persist what decoded cleanly, and a failed account is still re-fetched next cycle (cursor never advances past un-indexed signatures). Idempotency is preserved via the fact table's ReplacingMergeTree dedup, so per-account inserts are safe to repeat.

### DIFF
--- a/indexer/pkg/dz/serviceability/permissionevents/view.go
+++ b/indexer/pkg/dz/serviceability/permissionevents/view.go
@@ -200,52 +200,56 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error) 
 		return result, fmt.Errorf("get high water marks: %w", err)
 	}
 
+	// Persist each account's events as soon as it finishes fetching, rather than
+	// accumulating every account into one end-of-cycle insert. The per-account cursor is
+	// derived from the fact table (max(slot) per permission_pk), so committing an account
+	// immediately advances its cursor. If the refresh later times out — the Temporal activity
+	// has a bounded StartToCloseTimeout — the accounts already committed are not re-fetched
+	// next cycle. A single terminal batch instead discards the whole cycle's work on timeout
+	// and re-fetches it forever: a poison loop once the backlog exceeds one timeout window.
 	var (
-		mu        sync.Mutex
-		allEvents []PermissionEventRow
+		mu            sync.Mutex
+		totalInserted int64
 	)
 	var g errgroup.Group
 	g.SetLimit(maxConcurrentFetches)
 	for _, pda := range pdas {
 		g.Go(func() error {
-			events, err := v.fetchAccountEvents(ctx, pda, hwms[pda.String()])
+			events, fetchErr := v.fetchAccountEvents(ctx, pda, hwms[pda.String()])
+			// Persist whatever decoded cleanly before surfacing any fetch error (idempotent
+			// via ReplacingMergeTree). A failed account isn't cursor-advanced past its
+			// un-indexed signatures, so it is simply re-fetched next refresh — no silent gap.
 			if len(events) > 0 {
+				if err := v.store.InsertEvents(ctx, events); err != nil {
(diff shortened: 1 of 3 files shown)


juan-malbeclabs (approved):
Approving — correct, minimal, and well-documented. Both changes verified.

**Poison-loop fix (`permissionevents/view.go`)**
- Sound. Committing each account's events inside its goroutine advances that account's fact-table cursor immediately, so a mid-cycle Temporal timeout keeps completed accounts' progress instead of discarding the whole cycle. Mirrors the existing checkpoint-per-chunk pattern in `BackfillRefresh`.
- Idempotency via ReplacingMergeTree makes per-account re-inserts safe; a failed account returns its `fetchErr` without advancing its cursor, so it's re-fetched next cycle with no silent gap.
- `totalInserted` accounting and the single `ViewRefreshTotal(error)` increment via `g.Wait()` are both correct.

**`dberror` classification**
- Verified: `errStr` is lowercased, so the new lowercase patterns match the capitalized prod messages. The plural `"resources exhausted"` is genuinely required — `strings.Contains("resources exhausted", "resource exhausted")` is false because of the extra `s`. New table tests cover both signatures.

**Non-blocking note**
- Insert concurrency rises from 1 batched insert/cycle to up to `maxConcurrentFetches` (10) concurrent `InsertEvents` calls, i.e. more (smaller) inserts → more ClickHouse parts to merge. Correct tradeoff for the fix given the small watched-account set; just flagging it as the real behavioral change beyond timeout resilience. `store.InsertEvents`' Info log also now fires per-account per cycle rather than once.

No changes required.

PR #706, opened 2026-07-23
indexer: drain permission-events accounts in durable chunks to break testnet poison loop

ben-dz:
## Summary of Changes
- Break the permission-events poison loop for single-account backlogs. Since 2026-07-22 21:07 UTC, testnet `RefreshPermissionEvents` timed out every cycle: one Permission PDA receives ~3 tx/s of `AddMulticastGroupSubAllowlist` instructions, which reference the PDA but are not permission-management ops, so they decode to zero audit rows. The fact-derived high-water mark could therefore never advance, and every 5-minute activity re-fetched the same (growing) backlog from scratch — #704 checkpointed per account, but couldn't make progress *within* an account, and its partial-persist fallback ran on the already-cancelled activity context so nothing landed.
- The steady-state watch now drains each account oldest-first in 200-signature chunks: decode a chunk concurrently, insert it, then advance a new durable per-account cursor (`dz_permission_events_account_cursor`) that moves through row-less transactions. A refresh cut short by cancellation or its deadline re-does at most one chunk of work. The cursor always wins over the fact-derived high-water mark once present (the hwm remains only as the pre-upgrade fallback) — comparing slots would let stale fact rows, e.g. surviving a ledger reset, shadow the cursor forever.
- A deadline budget stops starting new chunks within 30s of the activity deadline: stop-with-progress is a success (next cycle resumes from the cursor; reported via a new `partial` refresh status so a never-converging drain stays visible), stop-with-zero-progress is an error so a starved account escalates instead of no-op succeeding. Signature pagination — which is O(remaining backlog) and runs before any chunk work — checks the same budget per page and fails fast with a distinct error rather than burning the window first.
- A chunk that fails decode is dropped whole rather than partially persisted — decode order is not cursor order, so a partial-chunk insert could advance the fact-derived mark past never-decoded signatures, silently and permanently dropping audit rows (the previous code had this latent gap; a regression test now pins it).
- Not-found transactions are handled by age: a `getTransaction` null more than 300 slots below the account's fetch tip is pruned history — skipped with a warning and counted via `doublezero_data_indexer_permission_events_skipped_tx_total` instead of wedging the drain forever. Near the tip a null is usually a load-balanced backend lagging finalization (the retry client only retries transport failures, never null results), so the chunk fails and the un-advanced cursor retries the signature next refresh — no audit row is silently lost to a transient miss.
- `getTransaction` concurrency is bounded at 10 view-wide by a shared semaphore. Pre-PR per-account decode was sequential (effective peak 10); the chunked rework alone would have allowed up to 10 accounts × 10 fetches = 100 in flight — introduced and closed within this PR, now pinned by a regression test.
- During a multi-cycle drain, freshness (`source_max_event_ts`) advances to the oldest committed chunk frontier across backlogged accounts, not to `now` — a backlogged account's newest events stay unindexed until the drain converges.
- The admin `--truncate` recovery path now also resets the account cursor table; leaving it behind would make the steady-state watch resume from a stale cursor after a reset.

## Diff Breakdown
| Core logic   |     3 | +305 / -95  | +210 |
| Tests        |     4 | +731 / -27  | +704 |
| Scaffolding  |     2 | +36  / -1   |  +35 |

Mostly new test coverage around a focused rework of the per-account drain path.

<details>

- [`indexer/pkg/dz/serviceability/permissionevents/view.go`](https://github.com/malbeclabs/lake/pull/706/files#diff-459842cd4012d0dffff4899fb743474d212ab8c0ad9bdeeb226cc0843a0066fa) — per-account fetch rebuilt as `drainAccount`: oldest-first chunked decode+insert+cursor checkpoint, budget-aware pagination and chunking, shared decode semaphore, age-gated not-found handling
- [`indexer/pkg/dz/serviceability/permissionevents/store.go`](https://github.com/malbeclabs/lake/pull/706/files#diff-186946680acee51d57959380f946a72d78c441db31b50d12e9b32c8b03ac8fc8) — durable per-account cursor read/write (`GetAccountCursors`/`SetAccountCursor`), tie-broken by `(updated_at, last_slot)`
- [`indexer/pkg/dzingest/backfill_permission_events.go`](https://github.com/malbeclabs/lake/pull/706/files#diff-e5f4720c4b8c9b48b0a3082d5cb351be90a7bb0b346b7b297bb548e909615b07) — `--truncate` also resets the account cursor table

</details>

## Testing Verification
- Twelve integration tests (testcontainers ClickHouse + fake Solana RPC); those pinning new behavior were verified failing first:
  - a backlog cancelled mid-drain resumes from the committed chunk instead of re-fetching everything, with the cursor advancing through row-less transactions (the exact prod failure mode)
  - a mid-chunk decode failure persists nothing and a retry backfills all events — the old code persisted a newest-first partial prefix and permanently skipped the older half
  - a near-tip not-found fails the chunk with nothing persisted, and the event lands once the lagging backend catches up; an old (pruned) not-found is skipped so the drain can't wedge
  - the durable cursor wins over a higher fact-derived high-water mark (backfill overshoot / ledger-reset remnants)
  - budget exhaustion with zero progress surfaces an error with no decode work started; budget stop after a committed chunk returns success with freshness clamped to the committed frontier, and the next refresh drains the remainder; pagination that exhausts the budget fails fast before any decode (deterministic via a fake clock advanced by the RPC stub)
  - peak in-flight `getTransaction` stays ≤ 10 with three accounts draining concurrently — validated to bite by temporarily bypassing the semaphore (peak observed: 30)
  - `TruncatePermissionEvents` clears the fact table and both cursor tables from seeded rows — validated to bite by temporarily dropping the account cursor from the truncate list
- Upgrade path covered: accounts indexed before the cursor table existed resume from the fact-derived high-water mark.
- Package suites run repeatedly with `-race`; full repo suite run with `-race` (surfaced a same-millisecond cursor-read tie in the new test, fixed with the `(updated_at, last_slot)` tie-break).
- Note for deploy: the new table is created by the existing migration flow (`--migrations-enable`). After rollout, the testnet whale (~10⁵ signatures and growing) drains at roughly 30–100 tx/s of decode, so the alert should quiet within a few refresh cycles' worth of drain; stopping the testnet spammer (`BUtAWK…` from 64.225.32.240) is a separate follow-up.

### DIFF
--- a/indexer/pkg/dz/serviceability/permissionevents/view.go
+++ b/indexer/pkg/dz/serviceability/permissionevents/view.go
@@ -19,16 +19,23 @@ import (
 )
 
 const (
-	// maxConcurrentFetches limits parallel getTransaction calls per refresh.
+	// maxConcurrentFetches limits in-flight getTransaction calls across the whole view
+	// via a shared semaphore: accounts drain concurrently, but the RPC never sees more
+	// than this many transaction fetches at once.
 	maxConcurrentFetches = 10
 	// maxSignaturesPerRequest is the Solana RPC page limit.
 	maxSignaturesPerRequest = 1000
-	// scanChunkSize is how many signatures the full-program backfill scan fetches,
-	// decodes, and durably writes before advancing the program scan cursor. Chunking
-	// (oldest-first) bounds how much work is lost if the backfill is cancelled — a
-	// full-history sweep that exceeds the Temporal activity timeout resumes from the last
+	// scanChunkSize is how many signatures the program backfill scan and the per-account
+	// drain decode and durably write before advancing their cursor. Chunking
+	// (oldest-first) bounds how much work is lost if a refresh is cancelled — a
+	// backlog that exceeds the Temporal activity timeout resumes from the last
 	// completed chunk instead of restarting from scratch.
 	scanChunkSize = 200
+	// drainBudgetMargin is how much of the refresh context's deadline the per-account
+	// drain leaves unspent when deciding to start another chunk: enough to decode,
+	// insert, and checkpoint that chunk. When less than this remains, the drain stops
+	// and the next refresh continues from the durable cursor.
+	drainBudgetMargin = 30 * time.Second
 	// permissionAccountType is the serviceability account_type discriminator (first byte
 	// of account data) for Permission accounts — AccountType::Permission = 15 in
 	// state/permission.rs. Used to enumerate Permission PDAs via getProgramAccounts.
@@ -74,6 +81,11 @@ type View struct {
 	store     *Store
 	refreshMu sync.Mutex
 
+	// decodeSem bounds in-flight getTransaction calls view-wide (see
+	// maxConcurrentFetches); per-chunk errgroup limits alone would multiply
+	// by the number of concurrently draining accounts.
+	decodeSem chan struct{}
+
 	readyOnce sync.Once
 	readyCh   chan struct{}
 
@@ -91,10 +103,11 @@ func NewView(cfg ViewConfig) (*View, error) {
 		return nil, fmt.Errorf("failed to create store: %w", err)
 	}
 	return &View{
-		log:     cfg.Logger,
-		cfg:     cfg,
-		store:   store,
-		readyCh: make(chan struct{}),
+		log:       cfg.Logger,
+		cfg:       cfg,
+		store:     store,
+		decodeSem: make(chan struct{}, maxConcurrentFetches),
+		readyCh:   make(chan struct{}),
 	}, nil
 }
 
@@ -159,8 +172,11 @@ func (v *View) safeRefresh(ctx context.Context) {
 //  2. per PDA, getSignaturesForAddress since that account's high-water mark and decode only
 //     those transactions.
 //
-// Because a Permission PDA is only ever referenced by permission instructions, this fetches
-// only permission transactions rather than the whole program's history. New grants are
+// This fetches only transactions touching Permission PDAs rather than the whole program's
+// history. Note a Permission PDA is not only referenced by permission-management
+// instructions: other serviceability instructions (e.g. multicast allowlist ops) reference
+// it too and decode to zero audit rows — which is why resume progress is tracked in a
+// durable per-account cursor rather than derived from indexed rows. New grants are
 // picked up once the PDA is discovered (its create tx references the account). The only gap
 // is an account created and deleted between two discovery polls (a deleted account no longer
 // appears in getProgramAccounts) — BackfillRefresh's full-history program scan is the
@@ -191,61 +207,77 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error) 
 		return result, nil
 	}
 
-	// Resume each account from the newest slot already indexed for it. The fact table holds
-	// every permission event for a given PDA, so max(slot) per permission_pk is a sound
-	// per-account cursor (unlike a program-wide scan, which the fact table can't anchor).
+	// Resume each account from its durable cursor, falling back to the fact-derived
+	// high-water mark for accounts indexed before the cursor table existed. When both
+	// exist take the newer: the program-wide backfill can insert rows beyond the cursor.
 	hwms, err := v.store.GetHighWaterMarks(ctx)
 	if err != nil {
 		metrics.ViewRefreshTotal.WithLabelValues(metricSource, "error").Inc()
 		return result, fmt.Errorf("get high water marks: %w", err)
 	}
+	cursors, err := v.store.GetAccountCursors(ctx)
+	if err != nil {
+		metrics.ViewRefreshTotal.WithLabelValues(metricSource, "error").Inc()
+		return result, fmt.Errorf("get account cursors: %w", err)
+	}
+	resume := func(pk string) HighWaterMark {
+		cur, hasCur := cursors[pk]
+		if hwm, ok := hwms[pk]; ok && (!hasCur || hwm.Slot > cur.Slot) {
+			return hwm
+		}
+		return cur
+	}
 
-	// Persist each account's events as soon as it finishes fetching, rather than
-	// accumulating every account into one end-of-cycle insert. The per-account cursor is
-	// derived from the fact table (max(slot) per permission_pk), so committing an account
-	// immediately advances its cursor. If the refresh later times out — the Temporal activity
-	// has a bounded StartToCloseTimeout — the accounts already committed are not re-fetched
-	// next cycle. A single terminal batch instead discards the whole cycle's work on timeout
-	// and re-fetches it forever: a poison loop once the backlog exceeds one timeout window.
+	// Drain each account in durable oldest-first chunks (see drainAccount). A refresh cut
+	// short by the Temporal activity timeout keeps every committed chunk, so even a single
+	// account whose backlog exceeds one timeout window makes forward progress each cycle
+	// instead of re-fetching the same work forever: the poison-loop failure mode.
 	var (
 		mu            sync.Mutex
 		totalInserted int64
+		totalPending  int
 	)
 	var g errgroup.Group
 	g.SetLimit(maxConcurrentFetches)
 	for _, pda := range pdas {
 		g.Go(func() error {
-			events, fetchErr := v.fetchAccountEvents(ctx, pda, hwms[pda.String()])
-			// Persist whatever decoded cleanly before surfacing any fetch error (idempotent
-			// via ReplacingMergeTree). A failed account isn't cursor-advanced past its
-			// un-indexed signatures, so it is simply re-fetched next refresh — no silent gap.
-			if len(events) > 0 {
-				if err := v.store.InsertEvents(ctx, events); err != nil {
-					return fmt.Errorf("insert permission events for %s: %w", pda.String(), err)
-				}
-				mu.Lock()
-				totalInserted += int64(len(events))
-				mu.Unlock()
-			}
-			if fetchErr != nil {
-				v.log.Warn("serviceability/permission-events: failed to fetch account",
-					"permission_pk", pda.String(), "error", fetchErr)
-				return fetchErr
+			inserted, pending, drainErr := v.drainAccount(ctx, pda, resume(pda.String()))
+			mu.Lock()
+			totalInserted += inserted
+			totalPending += pending
+			mu.Unlock()
+			if drainErr != nil {
+				v.log.Warn("serviceability/permission-events: failed to drain account",
+					"permission_pk", pda.String(), "error", drainErr)
+				return fmt.Errorf("drain account %s: %w", pda.String(), drainErr)
 			}
 			return nil
 		})
 	}
-	if err := g.Wait(); err != nil {
+	// Committed chunks are durable regardless of the refresh outcome — report them, and
+	// let freshness advance on partial progress (overstates by at most one cycle).
+	err = g.Wait()
+	result.RowsAffected = totalInserted
+	if totalInserted > 0 || err == nil {
+		fetchedAt := time.Now().UTC()
+		result.SourceMaxEventTS = &fetchedAt
+	}
+	if err != nil {
 		metrics.ViewRefreshTotal.WithLabelValues(metricSource, "error").Inc()
 		return result, fmt.Errorf("refresh permission accounts: %w", err)
 	}
 
-	result.RowsAffected = totalInserted
-	fetchedAt := time.Now().UTC()
-	result.SourceMaxEventTS = &fetchedAt
-
 	v.markReady()
-	metrics.ViewRefreshTotal.WithLabelValues(metricSource, "success").Inc()
+	// "partial" = every account made progress but at least one stopped at the refresh
+	// budget with backlog left. A drain that stays partial forever is a stalled drain —
+	// the metric is what keeps that visible, since each cycle reports success.
+	if totalPending > 0 {
+		metrics.ViewRefreshTotal.WithLabelValues(metricSource, "partial").Inc()
+		v.log.Info("serviceability/permission-events: drain stopped at refresh budget",
+			"inserted", totalInserted, "pending_signatures", totalPending, "watched_accounts", len(pdas))
+	} else {
+		metrics.ViewRefreshTotal.WithLabelValues(metricSource, "success").Inc()
+	}
 
 	if totalInserted > 0 {
 		v.log.Info("serviceability/permission-events: indexed new events",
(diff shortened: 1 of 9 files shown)


indexer/pkg/dz/serviceability/permissionevents/view.go
```diff
@@ -378,16 +410,85 @@ func (v *View) discoverPermissionAccounts(ctx context.Context) ([]solana.PublicK
...
+		cur := HighWaterMark{TxSignature: newest.Signature.String(), Slot: newest.Slot}
+		if err := v.store.SetAccountCursor(ctx, pda.String(), cur); err != nil {
+			return inserted, len(sigs) - start, fmt.Errorf("set account cursor: %w", err)
+		}
+	}
+	return inserted, 0, nil
+}
+
+// fetchAccountSignatures returns all signatures touching pda newer than resume, newest-first.
+func (v *View) fetchAccountSignatures(ctx context.Context, pda solana.PublicKey, resume HighWaterMark) ([]*rpc.TransactionSignature, error) {
```
armcconnell, line 540: **non-blocking:** `fetchAccountSignatures` pages the whole backlog before any budget check runs. For a backlog large enough that paging back to the cursor alone approaches the activity deadline, pagination alone can exhaust the budget every cycle — the poison-loop symptom relocated from decode to pagination rather than eliminated. Failure here is loud (hits the same zero-progress error), so it's not silent, but this scaling ceiling isn't tested or bounded. Given the PR cites a ~10⁵-signature and growing backlog, worth confirming pagination volume stays well under this ceiling, or chunking pagination against the budget too.

---
_Generated by [Claude Code](https://claude.ai/code)_

indexer/pkg/dz/serviceability/permissionevents/view.go
```diff
@@ -444,8 +538,23 @@ func (v *View) decodeChunk(ctx context.Context, sigs []*rpc.TransactionSignature
 	g.SetLimit(maxConcurrentFetches)
 	for _, sig := range sigs {
 		g.Go(func() error {
+			// The view-wide semaphore bounds total RPC pressure: without it, chunks
+			// draining in parallel across accounts would each fetch at the group limit.
+			select {
+			case v.decodeSem <- struct{}{}:
```
armcconnell, line 608: **non-blocking:** no regression test verifies peak in-flight `GetTransaction` calls stay ≤ `maxConcurrentFetches` through `decodeSem`. This bound is one of the PR's two headline fixes (closing the 10 accounts × 10 fetches = 100-in-flight hazard); without a test, a future change could silently reintroduce it. Consider a fake-RPC high-water-mark counter asserting peak concurrency.

---
_Generated by [Claude Code](https://claude.ai/code)_

indexer/pkg/dzingest/backfill_permission_events.go
```diff
@@ -54,11 +57,14 @@ func (a *Activities) TruncatePermissionEvents(ctx context.Context) error {
...
 	}
-	if err := conn.Exec(ctx, "TRUNCATE TABLE fact_dz_permission_events"); err != nil {
-		return fmt.Errorf("truncate fact_dz_permission_events: %w", err)
-	}
-	if err := conn.Exec(ctx, "TRUNCATE TABLE dz_permission_events_scan_cursor"); err != nil {
-		return fmt.Errorf("truncate dz_permission_events_scan_cursor: %w", err)
+	for _, table := range []string{
+		"fact_dz_permission_events",
+		"dz_permission_events_scan_cursor",
+		"dz_permission_events_account_cursor",
```
armcconnell, line 63: **non-blocking:** `dz_permission_events_account_cursor` truncation is an untested string literal — no test seeds a cursor row and asserts `TruncatePermissionEvents` clears it. This is a rarely-exercised, high-blast-radius recovery path; a typo or table rename here would reproduce the exact stale-cursor gap the comment above warns about, and would only surface in production.

---
_Generated by [Claude Code](https://claude.ai/code)_

indexer/pkg/dz/serviceability/permissionevents/view.go
```diff
@@ -191,61 +207,77 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error)
...
-	fetchedAt := time.Now().UTC()
-	result.SourceMaxEventTS = &fetchedAt
-
 	v.markReady()
-	metrics.ViewRefreshTotal.WithLabelValues(metricSource, "success").Inc()
+	// "partial" = every account made progress but at least one stopped at the refresh
+	// budget with backlog left. A drain that stays partial forever is a stalled drain —
+	// the metric is what keeps that visible, since each cycle reports success.
+	if totalPending > 0 {
+		metrics.ViewRefreshTotal.WithLabelValues(metricSource, "partial").Inc()
```
armcconnell, line 296: **nit:** `ViewRefreshTotal` gains a third status value (`partial`) alongside `success`/`error`/`panic`. Nothing in this repo currently alerts on that ratio, but flag it for whoever owns any downstream dashboard/alert that assumes only the prior set of statuses.

---
_Generated by [Claude Code](https://claude.ai/code)_

armcconnell (approved):
No blocking issues. Non-blocking: pagination-before-budget-check scaling ceiling, missing concurrency-bound and truncate-path tests. Nit: PR description overstates pre-PR concurrency (was capped at 10, not 100 — the 100-in-flight case is introduced and closed within this same PR); new `partial` refresh-status value.

---
_Generated by [Claude Code](https://claude.ai/code)_

indexer/pkg/dz/serviceability/permissionevents/view.go
```diff
@@ -444,8 +545,23 @@ func (v *View) decodeChunk(ctx context.Context, sigs []*rpc.TransactionSignature
 	g.SetLimit(maxConcurrentFetches)
 	for _, sig := range sigs {
 		g.Go(func() error {
+			// The view-wide semaphore bounds total RPC pressure: without it, chunks
+			// draining in parallel across accounts would each fetch at the group limit.
+			select {
+			case v.decodeSem <- struct{}{}:
+			case <-ctx.Done():
```
nikw9944, line 609: A single null `getTransaction` response permanently drops an audit row. `rpc.ErrNotFound` in solana-go v1.19.0 just means one backend returned a JSON-RPC `null` result, and the `rpc.NewWithRetries` client (`jsonrpc.WithRetry`) only retries transport/HTTP failures — a `null` is a successful response and is never retried. Behind the load-balanced `LedgerPublicRPCURL`, `null` happens transiently (signature listed by one backend, `getTransaction` routed to another that lags finalization or has partial history), and the steady-state drain processes signatures right up to the finalized tip every 5 minutes, where that window is widest. On the first `null` the chunk succeeds and the durable cursor advances past the signature — a permanent silent gap, the class this PR is built to prevent; the comment itself names "inconsistent upstream history" (the recoverable case) as a cause it skips.

Fix: retry the not-found within the chunk (bounded, short delay), and/or only skip when `sig.Slot` is well below the finalized tip; for a near-tip not-found, return the error so the chunk fails and the un-advanced cursor retries next refresh. A genuinely pruned old signature still gets skipped by the age gate, so the drain can't wedge.

indexer/pkg/dz/serviceability/permissionevents/view.go
```diff
@@ -191,61 +207,77 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error)
...
 		return result, fmt.Errorf("get high water marks: %w", err)
 	}
+	cursors, err := v.store.GetAccountCursors(ctx)
+	if err != nil {
+		metrics.ViewRefreshTotal.WithLabelValues(metricSource, "error").Inc()
+		return result, fmt.Errorf("get account cursors: %w", err)
+	}
+	resume := func(pk string) HighWaterMark {
+		cur, hasCur := cursors[pk]
+		if hwm, ok := hwms[pk]; ok && (!hasCur || hwm.Slot > cur.Slot) {
```
nikw9944: `resume()` lets a stale fact-derived high-water mark permanently shadow the durable cursor after a ledger reset. The cursor read was designed to survive resets (`store.go`: `updated_at` stays authoritative "where slots restart low"), but this raw slot comparison undoes it: post-reset with retained fact rows, the old-ledger hwm slot exceeds any new-ledger cursor slot forever, so every refresh resumes from a dead signature, `until` never matches, and the account's entire new-ledger history is re-paged and re-decoded every cycle — unbounded growing rework the cursor can never win back.

Fix: return the cursor whenever one exists (`if hasCur { return cur }`), keeping the fact hwm only as the pre-cursor-upgrade fallback. Cost is bounded and idempotent: after a program-wide backfill inserts rows beyond the cursor, one drain pass re-decodes those signatures once and the cursor catches up.

indexer/pkg/dz/serviceability/permissionevents/view.go
```diff
@@ -191,61 +207,77 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error)
...
+				v.log.Warn("serviceability/permission-events: failed to drain account",
+					"permission_pk", pda.String(), "error", drainErr)
+				return fmt.Errorf("drain account %s: %w", pda.String(), drainErr)
 			}
 			return nil
 		})
 	}
-	if err := g.Wait(); err != nil {
+	// Committed chunks are durable regardless of the refresh outcome — report them, and
+	// let freshness advance on partial progress (overstates by at most one cycle).
```
nikw9944: "(overstates by at most one cycle)" is wrong during a multi-cycle drain: `SourceMaxEventTS` advances to `now` every partial cycle while a backlogged account's newest events stay unindexed for the whole drain (days for the testnet whale), so the overstatement is bounded by the remaining backlog, not one cycle. Reword, or clamp to the newest committed chunk's block time if honest freshness is preferred.

nikw9944 (changes requested):
The chunked-drain design is sound and well tested, but two cursor-advancement paths need fixing before merge: the not-found skip treats a per-backend transient `null` as permanent audit-row loss (H1), and the resume fallback lets a stale fact-derived high-water mark permanently shadow the new durable cursor after a ledger reset (M1).

indexer/pkg/dz/serviceability/permissionevents/view.go
ben-dz: Addressed in 75a748b: pagination now checks the same deadline budget per page and fails fast with a distinct error (`refresh budget exhausted during signature pagination`) instead of paging the whole backlog before the zero-progress check catches it. Covered by `TestLake_PermissionEvents_View_PaginationRespectsBudget` (fake clock advanced per page, asserts zero decode work starts).

The O(remaining backlog) pagination cost itself is inherent to newest-first pages + oldest-first cursor advance, so the ceiling stays (loudly) rather than being bounded — for the current whale that's ~10² pages against a ~10³-page budget. If a backlog ever outgrows that, the distinct error names the situation directly.

indexer/pkg/dz/serviceability/permissionevents/view.go
ben-dz: Added in 75a748b: `TestLake_PermissionEvents_View_BoundsConcurrentTransactionFetches` drains three accounts concurrently with a high-water-mark counter in the fake RPC and asserts peak in-flight `GetTransaction` ≤ `maxConcurrentFetches`. Verified it bites: with the semaphore acquire bypassed, the observed peak is exactly 30 (3 accounts × inner limit 10) and the test fails.

indexer/pkg/dzingest/backfill_permission_events.go
ben-dz: Added in 75a748b: `TestLake_DZIngest_TruncatePermissionEvents_ResetsFactAndCursors` (new dzingest testcontainers harness) seeds a row in the fact table and both cursor tables and asserts all three are empty after the activity runs. Verified it bites by temporarily dropping the account-cursor table from the truncate list — the surviving row fails the test.

indexer/pkg/dz/serviceability/permissionevents/view.go
ben-dz: Documented in 75a748b — `ViewRefreshTotal`'s help text now enumerates the status values including `partial`. Nothing in infra currently alerts on refresh status (the lake-indexer alert is log-based), and the PR body flags the new value for dashboard owners.

indexer/pkg/dz/serviceability/permissionevents/view.go
ben-dz: Right — the retry client only retries transport failures, so a per-backend `null` was being converted into a permanently skipped audit row with the cursor advanced past it. Fixed in 67dd9bc with the age gate you suggested: a not-found within `notFoundSkipSlotLag` (300 slots) of the account's fetch tip fails the chunk, so the un-advanced cursor retries it next refresh; only signatures well below the tip are skipped as pruned history (still WARN + counted). The same threshold is threaded through the backfill's `decodeChunk` call.

Tests (both verified failing first): `TestLake_PermissionEvents_View_NearTipNotFoundFailsChunkAndRetries` — a near-tip null fails the refresh with nothing persisted, then the event lands once the lagging backend catches up; `TestLake_PermissionEvents_View_SkipsUnretrievableOldTransaction` — an old missing signature still skips so the drain can't wedge.

indexer/pkg/dz/serviceability/permissionevents/view.go
ben-dz: Fixed in 67dd9bc exactly as suggested: the durable cursor now always wins when present, and the fact-derived high-water mark remains only as the pre-cursor-upgrade fallback. Agreed the post-backfill cost is the right trade — one idempotent re-decode of the overshoot span versus unbounded re-paging after a ledger reset.

`TestLake_PermissionEvents_View_DurableCursorWinsOverStaleHighWaterMark` pins it (cursor at slot 100 + fact row at slot 105 → resume from the cursor; verified failing first — the old code resumed from 105).

indexer/pkg/dz/serviceability/permissionevents/view.go
ben-dz: Clamped in 67dd9bc (the comment was wrong, as you said): on a partial cycle `SourceMaxEventTS` now advances to the oldest committed chunk frontier across backlogged accounts (block time of the committed chunk tail), and to `now` only on a full clean pass. The budget-stop test asserts the exact frontier value.

nikw9944 (approved):
All three of my round-one findings are verified fixed in the current code, each with a regression test: the not-found skip is now age-gated so a near-tip null fails the chunk and retries instead of permanently dropping an audit row (H1); the durable cursor always wins resume over the fact-derived high-water mark, closing the ledger-reset shadow loop (M1); and partial-cycle freshness is clamped to the committed chunk frontier instead of overstating to now (L1). The other reviewer's four non-blocking findings are also verifiably addressed (pagination budget check, concurrency-bound regression test, truncate test, metric help text). The two new commits introduce no new issues above Low, and all CI including go-test passes on the new head.

Issue #708, opened 2026-07-23
indexer: telemetry-usage ingest never advances past maxTime — interface-counter data permanently ~1h stale


nikw9944:
## Summary

Interface-counter data (`fact_dz_device_interface_counters`) is permanently ~1 hour stale in production. The telemetry-usage refresh loop never ingests past `maxTime` in steady state, so `max(event_ts)` continuously lags `now()` by ~1h, advancing only in ~5-minute steps when the "data too old" branch fires.

This was diagnosed while working #702 (baseline-query CPU cost). It is a separate data-freshness bug and gets its own PR.

## Root cause

In `View.Refresh` (`indexer/pkg/dz/telemetry/usage/view.go`):

1. The incremental branch sets `queryStart = maxTime − 5m` (a deliberate overlap to catch late-arriving rows).
2. The catch-up cap (added in #665/#671, `maxCatchupChunks = 1`) then bounds the end:
   `queryEnd = queryStart + QueryChunk × maxCatchupChunks = queryStart + 5m = maxTime`.

Because the cap (5m) equals the overlap (5m), an incremental refresh can never query past `maxTime` — it only re-reads already-written rows (which dedup out) and inserts nothing. `maxTime` advances only when it falls behind the 1h query window and the "data too old" branch fires, ingesting a single ~5m chunk of hour-old data. The result is a sawtooth: `now() − max(event_ts)` oscillates ~3,450s..3,800s.

**Prod confirmation (2026-07-23):** `max(event_ts)` lagged `now()` by ~3,800s, stepping down ~5m at a time — consistent with the above.

## Fix direction

The catch-up cap must bound *new* data relative to `maxTime`, not the total span including the overlap. Roughly:

```
queryEnd = min(now, maxTime + maxCatchupChunks × QueryChunk)
```

so the steady-state span becomes overlap + one chunk (~10m → 2 Flux queries per refresh).

## Interaction with #665/#671 (must be checked in that PR, not folded into #702)

Allowing progress past `maxTime` changes ingest behavior under the same constraints #671 was tuned for:

- The steady-state span grows from one Flux query to two (~10m), which needs a fresh look at the 5m dzingest activity `StartToCloseTimeout` and the per-request HTTP timeout, plus the OOM headroom from #665. The overlap or chunk count may need retuning.

[...]

PR #709, opened 2026-07-23
indexer: make telemetry-usage baseline cache watermark-based; shorten lookback to 2d

### LINKED ISSUE #702: indexer: telemetry-usage argMaxIf baseline is now the #1 cluster CPU consumer (~20%) — shorten/cache/materialize
## Summary

With the co-tenant shredder MV fix landed (shredder#85 / shredder#86, −34% cluster CPU on 2026-07-21), the **indexer's telemetry-usage sparse-counter baseline query is now the #1 CPU consumer on the entire shared ClickHouse cluster**: **~747K CPU-sec/day ≈ 20% of the whole cluster** (~3.65M/day post-fix), reading **136 TiB/day**.

This is also the query that **timed out (code 159, 60s) during the June backfill**, leaving imperfect forward-fill baselines in the backfilled region — so fixing it is both the top perf item we own and a reliability fix.

## The query

`queryBaselineCountersFromClickHouse` — `indexer/pkg/dz/telemetry/usage/view.go:978`:

```sql
SELECT device_pk, intf,
       argMaxIf(in_discards,   event_ts, in_discards   IS NOT NULL) as in_discards_val,
       argMaxIf(in_errors,     event_ts, in_errors     IS NOT NULL) as in_errors_val,
       argMaxIf(in_fcs_errors, event_ts, in_fcs_errors IS NOT NULL) as in_fcs_errors_val,
       argMaxIf(out_discards,  event_ts, out_discards  IS NOT NULL) as out_discards_val,
       argMaxIf(out_errors,    event_ts, out_errors    IS NOT NULL) as out_errors_val
FROM fact_dz_device_interface_counters
WHERE event_ts >= ? AND event_ts < ?   -- windowStart - 7 DAYS .. windowStart
  AND (in_discards IS NOT NULL OR ... OR out_errors IS NOT NULL)
GROUP BY device_pk, intf
```

Measured (24h, `system.query_log`, 2026-07-22): **1,865 runs/day (~every 46s), 400 CPU-sec/run, ~75 GiB read/run**. It re-scans a **7-day window** of a billions-of-rows fact table to recover "last non-null value per (device, intf)" — then a 5-minute cache TTL (`baselineCacheTTL`) throws that work away.

Note: 1,865 runs/day is ~2× what a 5-min TTL across 3 envs (mainnet-beta/testnet/devnet ≈ 864/day) predicts — worth confirming where the extra misses come from (per-env view instances, non-realtime `windowStart` bypassing the cache, restarts) as part of this work.

## Fix options (cheapest first — they compose)

1. **Shorten the lookback.** The code comment itself says the indexer writes every few minutes, so the last non-null value "should be well within" the window — 7 days is enormously conservative. 1 day cuts the scan ~7×; even 2 days cuts ~3.5×. One-line change, no schema.
2. **Lengthen `baselineCacheTTL`** (5 min → 30–60 min) and/or carry baselines forward in memory from each refresh's own output (`convertRowsToUsage` already computes end-of-window values — the comment at view.go:975 says the cache is *usually* warmed from there; verify why it isn't in practice).
3. **Materialize the baseline (durable fix):** an incremental MV into an `AggregatingMergeTree` keyed `(device_pk, intf)` with `argMaxStateIf(...)` columns, so "last non-null per interface" is maintained on insert and the baseline read becomes a few-MB `argMaxMerge` over current state — O(interfaces), not O(7 days of rows). Kills both the CPU cost and the 60s-timeout failure mode for good (backfill included).

## Sibling (same pattern, smaller)

`argMax(rtt_us, (epoch, sample_index, event_ts))` link-latency (`indexer/pkg/dz/telemetry/latency/store.go:66,229`) — **~321K CPU-sec/day, 28 TiB/day** (#2 lake query). Same materialized-current-state fix applies; worth doing together if we go with option 3.

## Context

- Cluster CPU tracking: malbeclabs/infra#1837
- Co-tenant fixes: malbeclabs/shredder#85 (✅ verified −95%), malbeclabs/hyperliquid#201 (open)
- Backfill timeout history: the June device-interface-counters backfill hit this query's 60s limit repeatedly

nikw9944:
## Summary of Changes
* Make the telemetry-usage sparse-counter baseline cache **watermark-based** instead of wall-clock-TTL-based, so steady-state refreshes reuse it and the `argMaxIf` ClickHouse scan runs only on startup/backfill — not on nearly every refresh.
* The old `isRealtime` gate compared `windowStart` to `now`, but `windowStart` is always ~1h stale, so the cache never hit and the 7-day, billions-of-rows scan ran ~1,865×/day (~20% of the shared cluster CPU, 136 TiB/day read) — and this is the query that hit the 60s `max_execution_time` during the June backfill.
* Shorten the raw lookback **7d → 2d** (`baselineLookback`), carry baselines across contiguous backfill chunks, and add `doublezero_data_indexer_clickhouse_baseline_query_total{dz_env}` to make the fix verifiable in prod.
* Fixes #702

The underlying ingest-staleness bug that keeps `windowStart` ~1h stale is filed separately as #708 (its fix interacts with #665/#671 tuning and is independently observable/revertible; the watermark cache is designed to keep hitting in both the current stale regime and after #708 lands).

## Testing Verification
* New integration tests (testcontainers ClickHouse): cache hit at/near watermark (incl. `windowStart` above the watermark), bypass for historical `windowStart`, the 2-day lookback bound, and baseline carry across two contiguous backfill chunks asserting the scan runs exactly once.
* New pure-logic test for the merge semantics: a key silent for a window keeps its carried baseline, an empty window advances the watermark without wiping the cache (guards against the 0-key InfluxDB fallback), and a nil end-of-window struct is a no-op.
* Deliberately not changed: the cache-hit guard stays unbounded on the forward side — the live "data too old" catch-up path requires it (source stopped ⇒ ClickHouse has nothing newer), and the only unsafe shape (a large forward jump into a region another writer populated) can't arise from the admin backfill loop, which runs one contiguous ascending range per process with a fresh cache. Documented in the guard's comment.


indexer/pkg/dz/telemetry/usage/store_backfill.go
```diff
@@ -71,12 +78,25 @@ func (v *View) BackfillForTimeRange(ctx context.Context, startTime, endTime time
...
+	// next contiguous backfill chunk reuses them instead of re-running the
+	// expensive historical ClickHouse scan (which timed out repeatedly during the
+	// June backfill). The admin backfill loop processes ascending contiguous
+	// chunks in one process, so the next chunk's startTime == this chunk's endTime
+	// hits the watermark cache in queryBaselineCountersFromClickHouse. A
+	// non-contiguous backward call simply misses the cache and re-queries — a safe
+	// fallback (see the guard's doc for the forward-jump caveat). Done only after
+	// the rows are durably in ClickHouse (mirrors Refresh), so a failed insert
+	// can't leave the cache holding unpersisted values.
+	carryBaselines := func() { v.updateBaselineCache(endBaselines, endTime) }
```
armcconnell, line 115: **blocking:** the `len(rows) == 0` early return above (line 70-77, unchanged by this diff) never reaches `carryBaselines`/`updateBaselineCache`, unlike the `len(usage) == 0` branch here (line 98-99) and `Refresh()`'s empty-window handling (view.go ~line 492), both of which explicitly advance the watermark on an empty window. A backfill chunk with zero InfluxDB rows for its whole range — a real scenario for old/gappy historical data, exactly what this PR is meant to make robust — leaves `baselineCacheWatermark` stale, contradicting the documented invariant that it tracks "the data end... the cached baselines represent." Currently masked by the cache guard being unbounded on the forward side, but it's untested (none of the 5 new tests in baseline_cache_test.go cover a zero-row chunk) and becomes a real bug if that guard is ever tightened. Fix: build an empty `CounterBaselines{}` before the `len(rows) == 0` return and call `v.updateBaselineCache(emptyBaselines, endTime)` there too. Add a test: a mock InfluxDB returning 0 rows for one chunk sandwiched between two data-bearing chunks, asserting the watermark still advances and the third chunk still cache-hits.

---
_Generated by [Claude Code](https://claude.ai/code)_

indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -980,27 +1043,40 @@ func (v *View) convertRowsToUsage(rows []map[string]any, baselines *CounterBasel
...
+	// Historical windowStart values (backfill of an old region) fall far below the
+	// watermark and bypass the cache, re-querying ClickHouse as before.
+	//
+	// The guard is intentionally unbounded on the forward side: a live refresh that
+	// falls behind resumes at windowStart = now-QueryWindow, far above the watermark,
+	// and must still cache-hit (the source stopped, so ClickHouse has nothing new).
+	// The one unsafe shape — a large FORWARD jump into a region another writer has
+	// since populated — cannot arise from the admin backfill loop, which runs a
+	// single contiguous ascending range per process with a fresh (nil) cache.
+	if v.baselineCache != nil && !windowStart.Before(v.baselineCacheWatermark.Add(-baselineCacheMaxLag)) {
```
armcconnell: **non-blocking:** the single-writer safety argument here is an operational convention, not a code-enforced invariant — nothing stops an operator from running a manual admin backfill over a range a live indexer process has already cached, in which case the live process's in-memory cache keeps serving pre-correction values indefinitely (no external-write detection). Reasonable to accept given the two callers that exist today, but worth a one-line note (here or in an ops runbook) that re-backfilling a range covered by a running live process requires restarting that process.

---
_Generated by [Claude Code](https://claude.ai/code)_

indexer/pkg/dz/telemetry/usage/store_backfill.go
```diff
@@ -23,6 +23,13 @@ func (v *View) BackfillForTimeRange(ctx context.Context, startTime, endTime time
 		return nil, fmt.Errorf("start time (%s) must be before end time (%s)", startTime, endTime)
 	}
 
+	// Serialize with the refresh loop: both read and write the shared baseline
+	// cache/watermark. In production backfill runs in a standalone admin process
+	// that never starts the refresh loop, but taking the lock keeps the cache
+	// invariant enforced in code rather than by that call-site convention.
+	v.refreshMu.Lock()
+	defer v.refreshMu.Unlock()
```
armcconnell, line 34: **non-blocking:** holding `refreshMu` for the whole `BackfillForTimeRange` call (including the InfluxDB query and ClickHouse insert) is safe today only because backfill always runs in a separate one-shot admin process/`View` instance, never sharing a process with a running `Refresh()` loop. Not a defect in the current call path, but if this is ever invoked from a long-lived service instead, a multi-day historical backfill would block every live refresh cycle for its duration — worth keeping in mind for future callers.

---
_Generated by [Claude Code](https://claude.ai/code)_

indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -499,6 +524,44 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error)
...
+	mergeBaselineMap(v.baselineCache.InDiscards, endBaselines.InDiscards)
+	mergeBaselineMap(v.baselineCache.InErrors, endBaselines.InErrors)
+	mergeBaselineMap(v.baselineCache.InFCSErrors, endBaselines.InFCSErrors)
+	mergeBaselineMap(v.baselineCache.OutDiscards, endBaselines.OutDiscards)
+	mergeBaselineMap(v.baselineCache.OutErrors, endBaselines.OutErrors)
+	v.baselineCacheWatermark = watermark
+}
+
+// mergeBaselineMap upserts each non-nil src entry into dst.
+func mergeBaselineMap(dst, src map[string]*int64) {
```
armcconnell, line 590: **nit:** `mergeBaselineMap` only ever upserts, never deletes, so a decommissioned device/interface's baseline entry lives in the map for the life of the process (bounded by restart, low severity given typical churn) — worth a one-line doc note that this is intentionally unbounded rather than an oversight.

---
_Generated by [Claude Code](https://claude.ai/code)_

indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -191,11 +191,27 @@ func (cfg *ViewConfig) Validate() error {
...
+// data watermark this process processed, so its validity is judged against that
+// watermark, not against wall-clock time. Steady-state incremental refreshes
+// query [maxTime-5m, ...], so windowStart lands ~5m below the watermark; the
+// margin also covers capped catch-up steps.
+//
+// It must exceed the refresh overlap (5m, see queryStart in Refresh) plus one
+// capped catch-up step (QueryChunk × maxCatchupChunks, 5m); 15m leaves headroom.
+// If those constants grow past this margin the cache stops hitting and every
+// refresh scans ClickHouse again — watch doublezero_data_indexer_clickhouse_baseline_query_total.
+const baselineCacheMaxLag = 15 * time.Minute
```
armcconnell, line 235: **nit:** `baselineCacheMaxLag` only bounds the backward direction (how stale the watermark may be relative to `windowStart`) — the forward direction is explicitly unbounded by design (per the guard's doc at line ~1069). A name like `baselineCacheBackwardSlack`, or an explicit note here, would reduce the chance a future reader assumes it bounds staleness symmetrically.

---
_Generated by [Claude Code](https://claude.ai/code)_

armcconnell (changes requested):
Watermark-based cache redesign is sound and well-tested for its primary paths, but the zero-row backfill branch doesn't advance the watermark, which is untested and undermines the invariant the cache relies on. One blocking issue, three non-blocking/nits.

---
_Generated by [Claude Code](https://claude.ai/code)_

indexer/pkg/dz/telemetry/usage/view.go
ben-dz: Forward-unbounded cache guard serves stale baselines → silent permanent corruption.

Any `windowStart >= watermark - 15m` hits, no matter how far above. Within one writer this is safe (`maxTime <= watermark` always holds, so `windowStart <= watermark`). But a forward jump (`windowStart >> watermark`) arises when a *second* writer advances ClickHouse's max `event_ts` past this process's watermark. Concrete path: the live indexer freezes its in-memory watermark at T0 during an ingestion stall (refresh errors before `updateBaselineCache`); an operator runs the admin backfill in "continue from where we left off" mode (`backfill_device_interface_counters.go:84-87`, `startTime=MaxTime endTime=now`) writing `[T0, now)`; on recovery the live indexer's `GetMaxTimestamp` returns backfill's recent max, `windowStart = maxTime-5m ≈ now`, the guard hits and returns T0 baselines. Each interface's first post-gap sparse value is then delta'd against the pre-outage baseline, double-counting the gap; the row has a fresh `event_ts` so ReplacingMergeTree never corrects it. This is a regression — the old TTL would have expired and re-scanned.

The comment's safety argument only covers the backfill process's *own* nil cache, not the live indexer's warm cache. Fix: gate the forward hit on ClickHouse reality — pass in `maxTime` (already fetched at view.go:334) and hit only when `maxTime <= watermark + overlap`; otherwise miss and re-scan (a 2-day query now). This keeps the single-writer recovery hit (`maxTime == watermark`) while forcing a re-scan when another writer filled the gap. Add a stale-watermark + newer-`GetMaxTimestamp` test asserting the scan runs.

indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -486,6 +503,14 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error)
 	insertDuration := time.Since(insertStart)
 	v.log.Info("telemetry/usage: inserted data to clickhouse", "rows", len(usage), "duration", insertDuration.String())
 
+	// Update the baseline cache only after the rows are durably in ClickHouse. A
+	// failed insert must not leave the cache holding unpersisted end-of-window
+	// values: the next refresh's overlap re-reads those same rows and would
+	// compute their first sparse delta against a poisoned baseline. Merge (not
+	// replace) so interfaces that reported no rows this window keep their carried
+	// baseline instead of being dropped and losing forward-fill.
+	v.updateBaselineCache(endBaselines, queryEnd)
```
ben-dz, line 537: The commit-after-insert invariant (cache updated only after a successful insert, so a failed insert can't poison the next refresh's deltas) — a headline self-review fix — has no test. Add a failing-`InsertInterfaceUsage` (or canceled-context) case asserting `baselineCache`/`baselineCacheWatermark` are unchanged, so a future reorder of `updateBaselineCache` above the insert can't regress silently.

indexer/pkg/dz/telemetry/usage/view.go
ben-dz, line 590: `mergeBaselineMap` is upsert-only and the cache never expires, so an interface silent longer than `baselineLookback` keeps its carried baseline — the opposite of the "loses its baseline (one nil delta on its next report)" semantics documented at view.go:196-198. Memory risk is low (bounded by real device×interface cardinality). Consider evicting entries older than `baselineLookback` to match the documented behavior.

ben-dz (changes requested):
One blocking correctness hole plus a few smaller items.

**H1 (blocking):** The watermark cache-hit guard (`view.go:1069`) is unbounded on the forward side. Within a single writer `windowStart` can never exceed the watermark, so a forward jump only happens when a *second* writer advances ClickHouse past this process's watermark — the concrete case being the admin backfill's auto "continue from where we left off" mode writing the realtime region while a degraded live indexer holds a stale frozen watermark. On recovery the live indexer cache-hits stale baselines and computes the first post-gap sparse delta against a pre-outage value, silently and permanently corrupting error/discard deltas (fresh `event_ts`, so ReplacingMergeTree never fixes it). This is a regression: the old wall-clock TTL forced a correct re-scan. Fix: gate the forward hit on ClickHouse reality — `maxTime` (already fetched at view.go:334) `<= watermark + overlap`; otherwise re-scan (now only a 2-day query).

**Medium:** cache-hit + failed dedup query re-emits overlap rows with end-of-window baselines producing negative sparse deltas (M1); the commit-after-insert invariant the PR self-review added is untested (M2).

**Low:** the cache never evicts, contradicting the documented "loses its baseline" semantics (L1); the admin backfill emits the new metric with `dz_env=""` (L2).

The watermark approach itself is sound and the concurrency is correct.

- `indexer/pkg/dz/telemetry/usage/view.go:447` — medium: When `GetMaxTimestampsByKey` fails, `Refresh` proceeds with `alreadyWritten = nil` and re-processes the 5m overlap. With a cache hit the baselines are the end-of-window values (as of the watermark ≈ maxTime), so an overlap row at `t < maxTime` carrying a sparse increment computes `delta = v(t) - v(watermark) <= 0` (no regression guard for sparse counters at view.go:977-980), and the re-inserted row replaces the correct one via ReplacingMergeTree. Silent and permanent, though narrowly triggered (dedup query fails while others succeed; a sparse increment lands in the overlap and isn't consumed as the first-row baseline at view.go:933-943). Same shape is latent for a non-ascending backfill re-run (`store_backfill.go:81` always nil). Fix: on `alreadyWritten` fetch failure, bypass the cache (force a ClickHouse re-scan) or skip that refresh cycle — the H1 maxTime gate does not cover this path.
- `admin/internal/admin/backfill_device_interface_counters.go:155` — low: The backfill `ViewConfig` omits `DZEnv`, so the new `doublezero_data_indexer_clickhouse_baseline_query_total` metric fires under an empty `dz_env` label for backfill scans. The metric was added to make backfill/startup scan behavior observable — set `DZEnv` on the backfill view.

indexer/pkg/dz/telemetry/usage/store_backfill.go
nikw9944: Fixed in 3d86d55. The `len(rows) == 0` branch now calls `updateBaselineCache(baselines, endTime)` — merging the start-of-window baselines (which are also the end-of-window state for an empty chunk, and richer than an empty struct when the chunk 1 scan populated them) and advancing the watermark. Added `TestLake_TelemetryUsage_BaselineCache_BackfillZeroRowChunkAdvancesWatermark`: a zero-row chunk sandwiched between two data-bearing chunks, asserting the watermark advances, carried baselines survive, and the third chunk still cache-hits (one scan total).

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: Addressed in 3d86d55, in two parts. The gap-filling case (backfill writing past a live process's watermark) is now detected in code: the cache hit is gated on ClickHouse's max `event_ts`, and foreign writes past `watermark+overlap` discard the cache and force a re-scan. The remaining undetectable case — re-backfilling an already-cached *historical* range, which doesn't advance max `event_ts` — is called out in the guard's doc with the restart requirement you suggested.

indexer/pkg/dz/telemetry/usage/store_backfill.go
nikw9944: Added a note in 3d86d55 on the lock's doc: the lock spans the InfluxDB query and ClickHouse insert, so a long-lived service sharing a View with the refresh loop would have every live refresh blocked for the backfill's duration.

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: Documented in 3d86d55: `mergeBaselineMap`'s doc now states the upsert-only, never-delete behavior is intentional — bounded by real device×interface cardinality, cleared by restart — and spells out the consequence for interfaces silent longer than `baselineLookback`.

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: Addressed in 3d86d55. `baselineCacheMaxLag`'s doc now states it bounds only the backward direction, and the forward direction is no longer unbounded: it's gated on ClickHouse's max `event_ts` (see the other thread on the guard). Kept the name since the asymmetry is now explicit in both docs and code.

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: Fixed in 3d86d55 as suggested: `queryBaselineCountersFromClickHouse` now takes `chMaxTime` (the `maxTime` Refresh already fetched) and only hits when `maxTime <= watermark + refreshOverlap`; a forward excursion proves foreign writes and forces a re-scan. One step further than the suggestion: foreign writes discard the whole cache rather than just bypassing it — the post-refresh `updateBaselineCache` merges, so keys silent through the foreign-filled gap would otherwise carry pre-gap values back in and poison their next delta. Backfill passes nil (gating on the global max would defeat the chunk carry; safe per its fresh-cache contiguous-ascending contract, documented at the call site). Added `TestLake_TelemetryUsage_BaselineCache_ForeignWriteForcesRescan` (stale watermark + newer max event_ts → scan runs, cache invalidated) and kept the single-writer recovery hit covered in the existing hit test with an explicit `chMaxTime`.

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: Added in 3d86d55: `TestLake_TelemetryUsage_BaselineCache_FailedInsertLeavesCacheUnchanged` — the mock InfluxDB returns convertible rows then cancels the refresh context, so conversion succeeds and `InsertInterfaceUsage` fails; asserts `Refresh` errors and `baselineCache`/`baselineCacheWatermark` are untouched.

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: Chose a doc note over eviction in 3d86d55: `baselineLookback`'s doc already carries the "unless the in-memory cache still carries it forward" caveat, and `mergeBaselineMap` now documents the upsert-only retention as intentional with that consequence spelled out. Eviction would need per-key timestamps to fix a low-severity, restart-bounded behavior whose alternative (one nil delta) isn't clearly better than delta-against-carried-value; happy to add it if you disagree.

nikw9944:
Addressed both reviews in 3d86d55 (branch also rebased onto main).

**Blocking**
- Zero-row backfill chunk now advances the watermark (merging the start-of-window baselines, which are the end-of-window state for an empty chunk), with a sandwiched-empty-chunk test.
- The cache guard's forward side is now gated on ClickHouse's max `event_ts`: writes past `watermark+overlap` prove a second writer filled the gap, which discards the whole cache (not just bypasses it — the post-refresh merge would otherwise carry pre-gap values for silent keys back in) and forces a re-scan. Test covers stale watermark + newer max `event_ts`. Backfill skips the gate (nil `chMaxTime`) since the global max is the live indexer's realtime data; safe per its fresh-cache contiguous-ascending contract, documented at the call site.

**Medium**
- M1: a failed `GetMaxTimestampsByKey` now fails the refresh instead of proceeding without dedup — with cached end-of-window baselines, re-emitting overlap rows would write negative sparse deltas that ReplacingMergeTree keeps.
- M2: added a failed-insert test asserting the cache and watermark stay untouched.

**Low/nits**
- L2: admin backfill now passes `--dz-env` through to the view so the baseline query metric carries a real `dz_env` label.
- L1 / merge nit: documented upsert-only retention as intentional rather than adding eviction (see thread).
- Doc notes: `refreshMu` scope in backfill, `baselineCacheMaxLag` bounds only the backward direction, and re-backfilling an already-cached historical range still requires restarting the live indexer (the gate can't see writes that don't advance max `event_ts`).

Not changed: no eviction for the baseline map (doc note instead, rationale in the thread — will add if you disagree).

ben-dz (approved):
All five findings from the prior review are resolved in commit `3d86d55`, verified in the current code and pinned by new tests.

- **H1 (forward-jump stale baselines):** fixed — `queryBaselineCountersFromClickHouse` now gates the forward side on ClickHouse's max `event_ts` (`chMaxTime`); a foreign write past `watermark+refreshOverlap` discards the cache and forces a re-scan. Pinned by `ForeignWriteForcesRescan`. No false trigger in steady state (`maxTime <= watermark` invariant).
- **M1 (failed dedup → negative deltas):** fixed — a failed `GetMaxTimestampsByKey` now fails the refresh instead of proceeding without dedup.
- **M2 (commit-after-insert untested):** fixed — `FailedInsertLeavesCacheUnchanged` asserts the cache/watermark are untouched when the insert fails.
- **L1 (no eviction / doc mismatch):** addressed — retention is now documented as intentional and reconciled with `baselineLookback`'s caveat.
- **L2 (backfill `dz_env=""`):** fixed — `DZEnv` threaded from the `--dz-env` flag into the backfill view.

The new commit also correctly advances the watermark on zero-row backfill chunks (pinned by a new test) and introduces no new issues. `go-build`/`go-lint` pass; the ClickHouse integration tests are still pending on this head.

indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -191,11 +191,38 @@ func (cfg *ViewConfig) Validate() error {
...
+//
+// It must exceed refreshOverlap plus one capped catch-up step (QueryChunk ×
+// maxCatchupChunks, 5m); 15m leaves headroom. If those constants grow past this
+// margin the cache stops hitting and every refresh scans ClickHouse again —
+// watch doublezero_data_indexer_clickhouse_baseline_query_total.
+//
+// This bounds only the backward direction (windowStart below the watermark).
+// The forward direction is bounded separately, by comparing ClickHouse's max
+// event_ts against the watermark (see queryBaselineCountersFromClickHouse).
+const baselineCacheMaxLag = 15 * time.Minute
```
armcconnell, line 235: **non-blocking:** the doc comment says `baselineCacheMaxLag` (15m) must exceed `refreshOverlap + QueryChunk*maxCatchupChunks`, but nothing enforces that against `cfg.QueryChunk`, which is operator-settable (default only applied when `<= 0`, `view.go:181`). If `QueryChunk` is later bumped for an unrelated reason, the cache silently regresses to always-miss — quietly undoing this PR's fix, with only a rising `doublezero_data_indexer_clickhouse_baseline_query_total` (a brand-new metric with no established baseline/alert yet) as a signal. Consider asserting this invariant in `ViewConfig.Validate()` instead of relying solely on the comment.

---
_Generated by [Claude Code](https://claude.ai/code)_

indexer/pkg/dz/telemetry/usage/store_backfill.go
```diff
@@ -71,12 +94,25 @@ func (v *View) BackfillForTimeRange(ctx context.Context, startTime, endTime time
...
+	// next contiguous backfill chunk reuses them instead of re-running the
+	// expensive historical ClickHouse scan (which timed out repeatedly during the
+	// June backfill). The admin backfill loop processes ascending contiguous
+	// chunks in one process, so the next chunk's startTime == this chunk's endTime
+	// hits the watermark cache in queryBaselineCountersFromClickHouse. A
+	// non-contiguous backward call simply misses the cache and re-queries — a safe
+	// fallback (see the guard's doc for the forward-jump caveat). Done only after
+	// the rows are durably in ClickHouse (mirrors Refresh), so a failed insert
+	// can't leave the cache holding unpersisted values.
+	carryBaselines := func() { v.updateBaselineCache(endBaselines, endTime) }
```
armcconnell, line 115: **nit:** on a cache-hit, `baselines` here is the same `*CounterBaselines` pointer as `v.baselineCache`; the zero-rows path (`view.go`, `v.updateBaselineCache(baselines, endTime)`) then merges it back into itself via `mergeBaselineMap`. It's safe (existing-key overwrites only, no insert/delete during the range), but worth a one-line comment here noting the aliasing so the next reader doesn't have to re-derive that it's safe.

---
_Generated by [Claude Code](https://claude.ai/code)_

indexer/pkg/dz/telemetry/usage/baseline_cache_test.go
```diff
@@ -0,0 +1,382 @@
...
+	// Single-writer maxTime: ClickHouse's max event_ts sits at/below the watermark.
+	chMaxTime := watermark.Add(-1 * time.Minute)
+
+	// windowStart slightly below the watermark (normal 5m overlap).
+	got, err := view.queryBaselineCountersFromClickHouse(context.Background(), watermark.Add(-5*time.Minute), &chMaxTime)
+	require.NoError(t, err)
+	require.NotNil(t, got.InErrors[key])
+	require.Equal(t, int64(100), *got.InErrors[key], "should return cached value, not ClickHouse value")
+
+	// windowStart above the watermark ("data too old" branch): still a hit while
```
armcconnell, line 85: **nit:** this only exercises a +5m forward jump. The guard's forward side is deliberately left unbounded for single-writer stall-recovery (per the PR description's "disputed Medium" note) — a case with a much larger jump (e.g. +6h, still below `chMaxTime`) would lock in that documented behavior explicitly instead of leaving it implicit.

---
_Generated by [Claude Code](https://claude.ai/code)_

armcconnell (approved):
Watermark-cache/invalidation/merge logic and locking check out correct under trace (incl. the self-merge aliasing case); build and vet pass. 3 non-blocking/nit comments, no blockers.

---
_Generated by [Claude Code](https://claude.ai/code)_

armcconnell (approved):
Watermark cache, forward gate, merge-not-replace, and commit-after-insert ordering are all sound and well-covered by tests. No changes required.

---
_Generated by [Claude Code](https://claude.ai/code)_

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: Done in 0a48031: `ViewConfig.Validate` now rejects a `QueryChunk` where `refreshOverlap + QueryChunk×maxCatchupChunks` reaches `baselineCacheMaxLag`, with a test at the boundary (10m chunk). `baselineCacheMaxLag`'s doc points at the check. Neither binary sets `QueryChunk` explicitly, so nothing existing is affected.

indexer/pkg/dz/telemetry/usage/store_backfill.go
nikw9944: Done in 0a48031: the zero-row path's comment now spells out that `baselines` may alias `v.baselineCache` on a hit, and why the self-merge is safe (existing-key overwrites only, no insert/delete while ranging).

indexer/pkg/dz/telemetry/usage/baseline_cache_test.go
nikw9944: Done in 0a48031: added a +6h forward-jump case to the cache-hit test with max event_ts unmoved, locking in the deliberately-unbounded single-writer stall-recovery hit.

nikw9944:
Addressed the three follow-up comments from the approving review in 0a48031: the `QueryChunk`-vs-`baselineCacheMaxLag` invariant is now enforced in `ViewConfig.Validate` (with a boundary test) instead of comment-only, the zero-row backfill path documents the cache-aliasing self-merge and why it is safe, and the cache-hit test now locks in the unbounded single-writer forward jump (+6h with unmoved max `event_ts`). No behavior change for existing configs — neither binary sets `QueryChunk`.

PR #711, opened 2026-07-23
indexer: anchor telemetry-usage catch-up cap at maxTime so ingest advances past ~1h stale

### LINKED ISSUE #708: indexer: telemetry-usage ingest never advances past maxTime — interface-counter data permanently ~1h stale

nikw9944:
## Summary of Changes
* Anchor the telemetry-usage catch-up cap at `maxTime` (the ingest watermark) instead of `queryStart`, so incremental refreshes query past `maxTime` and `fact_dz_device_interface_counters` stops being permanently ~1h stale.
* The cap had equalled the 5m re-read overlap (`queryStart = maxTime − overlap`, cap = `QueryChunk × maxCatchupChunks` = 5m → `queryEnd = maxTime`), so every incremental refresh re-read only rows that dedup out and inserted nothing; `max(event_ts)` sawtoothed ~3,450–3,800s behind `now()`.
* Give `RefreshTelemetryUsage` a dedicated 10m activity `StartToCloseTimeout` with `MaximumAttempts: 1` — steady state now runs 2 Flux queries (~8m worst case) which no longer fits the shared 5m deadline; a single attempt avoids a ~30m loop stall since the workflow reschedules every iteration anyway.
* Fixes #708

## Testing Verification
* Added regression tests in `dztelemusage` (fake clock + per-test ClickHouse + InfluxDB mock capturing query spans): incremental refresh queries one chunk past `maxTime` (fails on `main`); steady state advances exactly one chunk per refresh (monotonic, no sawtooth); catch-up cap preserved for out-of-window/empty `maxTime` (the #665 memory bound); near-`now` tail clamps `queryEnd` to `now`.
* Post-deploy: `now() − max(event_ts)` on `fact_dz_device_interface_counters` should drop from ~3,500–3,800s to single-digit minutes and stay flat; `doublezero_data_indexer_clickhouse_baseline_query_total` should remain ~restarts-only (confirms no #702 baseline-cache interaction).
* Note: ClickHouse-backed tests require Docker and were not run locally (no Docker daemon in the dev environment); they run in CI. Build, `go vet`, and `golangci-lint` pass on the changed packages.

## Note on review deviation
The approved plan kept the existing retry policy; the architecture review flagged that `MaximumAttempts: 3` × the new 10m deadline widens the worst-case ingest-loop stall to ~30m (from ~15m) with no benefit, since `telemUsageEveryN = 1` reschedules the activity every ~1 minute. Reduced to `MaximumAttempts: 1`, which also keeps a hung Flux query from stacking attempts on `refreshMu`. The pre-existing zero-rows gap-stall (Medium, self-heals when `maxTime` ages out of the window) is documented in-code but not otherwise changed — out of scope for this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Non-actionable failures no longer log at ERROR (which pages on-call): indexer view refreshes and ingest workflows escalate WARN→ERROR only on sustained consecutive failures; slack, agent, and worker log sites classify transient and client-caused errors as WARN; probe-write, optional-listener, and shutdown-cleanup failures demote to WARN (#696)
 - `dberror` moved from `api/handlers/dberror` to `utils/pkg/dberror` so all services can share transient-error classification (#696)
 
+### Fixed
+
+- Indexer interface-counter data (`fact_dz_device_interface_counters`) is no longer permanently ~1h stale: the telemetry-usage catch-up cap is now anchored at `maxTime` (the ingest watermark) instead of `queryStart`, which had made the cap equal the 5m re-read overlap so an incremental refresh could never query past `maxTime` — every cycle re-read only rows that dedup out and inserted nothing, and `max(event_ts)` sawtoothed ~1h behind `now()`. Steady state now ingests the overlap plus one chunk (~10m, 2 Flux queries) and advances every refresh. `RefreshTelemetryUsage` gets a dedicated 10m activity `StartToCloseTimeout` (was the shared 5m) to bound the two-query worst case plus ClickHouse work, preventing the #665/#671 timeout loop. The post-downtime catch-up memory bound is unchanged (#708)
+
 ### Added
 
 - Shared `utils/pkg/logger` helpers: transient-aware `logger.Error`/`logger.Warn` and consecutive-failure `logger.Escalator`; logging-level convention documented in CLAUDE.md (#696)
--- a/indexer/pkg/dz/telemetry/usage/flux_influx_client.go
+++ b/indexer/pkg/dz/telemetry/usage/flux_influx_client.go
@@ -21,8 +21,11 @@ type FluxInfluxDBClient struct {
 
 // defaultFluxHTTPTimeout is the HTTP client timeout for Flux queries.
 // The influxdb-client-go/v2 default is 20s which is too short for pivot queries
-// over a 1-hour window of interface counter data. The Temporal activity
-// StartToCloseTimeout is 5 minutes, so 4 minutes gives a comfortable margin.
+// over a 1-hour window of interface counter data. A steady-state telemetry-usage
+// refresh runs 2 chunked queries (overlap re-read + one new chunk), so ~8m of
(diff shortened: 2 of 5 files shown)


ben-dz (approved):
lgtm. The catch-up cap is correctly re-anchored at `max(queryStart, maxTime)`: steady state now spans `[maxTime−overlap, maxTime+chunk]` and advances one chunk per refresh, while the nil/old-`maxTime` fallback to `queryStart` preserves the one-chunk post-downtime memory bound. The `now` clamp, the 2-chunk worst-case span (≤~8m Flux under the new 10m `StartToCloseTimeout`), and the baseline-cache gate all remain sound, and the regression tests genuinely fail on `main`.

PR #712, opened 2026-07-24
indexer: stop transient rollup/permission-events blips from paging

ben-dz:
## Summary

Two prod indexer ERROR lines overnight (2026-07-24) each tripped the aggregate "Lake Indexer: Errors (prod)" Grafana pager (which fires on *any* `level="ERR"` line) and self-resolved within ~5 minutes. Both were transient, self-healing conditions that shouldn't log at ERROR. This is alert-noise reduction, aligned with the logging-levels guidance in `CLAUDE.md` (transient/not-found/lifecycle conditions log WARN, not ERR).

- **RollupLinks (devnet):** a native-TLS `read: EOF` from ClickHouse Cloud (`:9440`) during the link-latency query. The activity auto-retries (Temporal `MaximumAttempts: 3`) and self-heals, but the Temporal SDK worker logs `Activity error.` at ERROR on the non-final `Attempt=1`. `temporalLogger.Error` now demotes to WARN when the carried error is transient (`dberror.IsTransient`), mirroring the existing context-cancellation demotion. Sustained failure still pages via the workflow-side `Escalator`. This is a recurring class (same EOF-from-`:9440` hit `timing statswriter` on 2026-05-29).
- **RefreshPermissionEvents (testnet):** a `getTransaction` null near an account's fetch tip — transient by design (a load-balanced backend lagging finalization; the un-advanced cursor retries next refresh, per #706). It propagated as a plain `rpc.ErrNotFound`, which `dberror.Classify` files as `Unknown`, so the activity escalator paged at the strict threshold (3 refreshes, ~15 min) instead of the transient threshold (10, ~50 min) used for other self-healing causes. The near-tip return is now marked transient via a new `dberror.ErrTransient` sentinel (still unwrapping to `rpc.ErrNotFound`). The old/pruned skip branch is unchanged.

## Follow-up (out of scope)

The near-tip age-gate (`skipNotFoundBelow`) is derived from the account's own newest signature (`sigs[0].Slot`). On a **low-traffic** PDA whose tip never advances, a genuinely-unretrievable near-tip null (e.g. a sig lost across a testnet ledger reset) would never age into the skip branch and would keep failing — now paging every ~50 min instead of ~15. No evidence this happened in the incident (both errors self-resolved). A durable fix would age-gate against ledger/wall-clock time or bound the retries before skipping; worth a separate change if it recurs.

## Testing Verification

- `dberror.IsTransient` returns true for an error wrapping `ErrTransient` and remains false for a plain not-found; the wrapped error still `errors.Is` its original cause. (unit test, passing)
- Extended the existing `NearTipNotFoundFailsChunkAndRetries` view test to assert the returned error is now `dberror.IsTransient` and still unwraps to `rpc.ErrNotFound`; the old/pruned skip path is unchanged.
- New rollup logger test: `temporalLogger.Error` with `"Activity error."` carrying an EOF/connectivity error logs WARN, a non-transient activity error stays ERROR, context cancellation stays WARN, and a transient error under a *different* message stays ERROR.
- Note: the `rollup` and `permissionevents` test packages need a ClickHouse testcontainer (`TestMain`), which was unavailable in the authoring environment (no Docker); those tests compile and are expected to run in CI. `dberror` tests, `go vet`, and `golangci-lint` (0 issues) all pass locally.

### DIFF
--- a/indexer/pkg/rollup/rollup.go
+++ b/indexer/pkg/rollup/rollup.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/ClickHouse/clickhouse-go/v2"
 	"github.com/malbeclabs/lake/indexer/pkg/ingestionlog"
+	"github.com/malbeclabs/lake/utils/pkg/dberror"
 	temporalclient "go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/worker"
 )
@@ -175,6 +176,15 @@ func (l *temporalLogger) Error(msg string, keyvals ...any) {
 		l.log.Warn(msg, keyvals...)
 		return
 	}
+	// Temporal logs "Activity error." at ERROR on every failed attempt,
+	// including non-final ones the activity's retry policy will recover. A
+	// transient cause (ClickHouse connection blip, timeout) self-heals on
+	// retry, so demote to WARN; a sustained failure still pages via the
+	// workflow-side Escalator once retries are exhausted across iterations.
+	if msg == "Activity error." && isTransientActivityError(keyvals) {
+		l.log.Warn(msg, keyvals...)
+		return
+	}
 	l.log.Error(msg, keyvals...)
 }
 
@@ -194,6 +204,23 @@ func hasBenignTaskProcessingError(keyvals []any) bool {
 				strings.Contains(msg, "grpc: the client connection is closing") {
 				return true
 			}
+		}
+	}
+	return false
+}
+
+// isTransientActivityError reports whether Temporal's activity-error keyvals
+// carry an Error that dberror classifies as transient (a self-healing upstream
+// blip the activity's retry policy will recover). The Error value arrives
+// wrapped in the SDK's ApplicationError, but dberror.Classify matches on the
+// unwrapped message string, so a connectivity/timeout cause still classifies.
+func isTransientActivityError(keyvals []any) bool {
+	for i := 0; i+1 < len(keyvals); i += 2 {
+		if keyvals[i] != "Error" {
+			continue
+		}
+		if err, ok := keyvals[i+1].(error); ok {
+			return dberror.IsTransient(err)
 		}
 	}
 	return false
--- a/indexer/pkg/rollup/rollup.go
+++ b/indexer/pkg/rollup/rollup.go
@@ -211,9 +211,11 @@ func hasBenignTaskProcessingError(keyvals []any) bool {
 
 // isTransientActivityError reports whether Temporal's activity-error keyvals
 // carry an Error that dberror classifies as transient (a self-healing upstream
-// blip the activity's retry policy will recover). The Error value arrives
-// wrapped in the SDK's ApplicationError, but dberror.Classify matches on the
-// unwrapped message string, so a connectivity/timeout cause still classifies.
+// blip the activity's retry policy will recover). At the "Activity error." log
+// site the Error keyval is the raw error the activity returned (conversion to
+// the SDK's ApplicationError happens afterward). dberror.Classify matches on
+// the message string, so classification would also hold if a converted error
+// ever appeared here.
 func isTransientActivityError(keyvals []any) bool {
 	for i := 0; i+1 < len(keyvals); i += 2 {
 		if keyvals[i] != "Error" {
--- a/indexer/pkg/solingest/solingest.go
+++ b/indexer/pkg/solingest/solingest.go
@@ -19,6 +19,7 @@ import (
 	"github.com/malbeclabs/lake/indexer/pkg/ingestionlog"
 	"github.com/malbeclabs/lake/indexer/pkg/sol"
 	"github.com/malbeclabs/lake/indexer/pkg/validatorsapp"
+	"github.com/malbeclabs/lake/utils/pkg/dberror"
 )
 
 // Config configures the Solana ingest worker.
@@ -143,7 +144,34 @@ func (l *temporalLogger) Error(msg string, keyvals ...any) {
 		l.log.Warn(msg, keyvals...)
 		return
 	}
+	// Temporal logs "Activity error." at ERROR on every failed attempt,
+	// including non-final ones the activity's retry policy will recover. These
+	// activities write to ClickHouse and return their errors to Temporal, so a
+	// transient connection blip (a native-TLS read: EOF from ClickHouse Cloud)
+	// self-heals on retry but still trips the aggregate level="ERR" pager.
+	// Demote to WARN; sustained failure still surfaces via the ingestion log.
+	// Mirrors the rollup worker's temporalLogger.
+	if msg == "Activity error." && isTransientActivityError(keyvals) {
+		l.log.Warn(msg, keyvals...)
+		return
+	}
 	l.log.Error(msg, keyvals...)
+}
+
+// isTransientActivityError reports whether Temporal's activity-error keyvals
+// carry an Error that dberror classifies as transient (a self-healing upstream
+// blip the activity's retry policy will recover). At the "Activity error." log
+// site the Error keyval is the raw error the activity returned.
+func isTransientActivityError(keyvals []any) bool {
+	for i := 0; i+1 < len(keyvals); i += 2 {
+		if keyvals[i] != "Error" {
+			continue
+		}
+		if err, ok := keyvals[i+1].(error); ok {
+			return dberror.IsTransient(err)
+		}
+	}
+	return false
 }
 
 // hasBenignTaskProcessingError reports whether Temporal's "Task processing
--- a/indexer/pkg/solingest/solingest.go
+++ b/indexer/pkg/solingest/solingest.go
@@ -149,8 +149,11 @@ func (l *temporalLogger) Error(msg string, keyvals ...any) {
 	// activities write to ClickHouse and return their errors to Temporal, so a
 	// transient connection blip (a native-TLS read: EOF from ClickHouse Cloud)
 	// self-heals on retry but still trips the aggregate level="ERR" pager.
-	// Demote to WARN; sustained failure still surfaces via the ingestion log.
-	// Mirrors the rollup worker's temporalLogger.
+	// Demote to WARN; a sustained failure still pages via the workflow-side
+	// Escalator, which observes the frequent activities (solana/geoip/
+	// validatorsapp) enough times per run to reach its threshold, and observes
+	// the once-per-run block production activity at a dedicated threshold of 1
+	// (see workflow.go). Mirrors the rollup worker's temporalLogger.
 	if msg == "Activity error." && isTransientActivityError(keyvals) {
 		l.log.Warn(msg, keyvals...)
 		return
--- a/utils/pkg/dberror/dberror.go
+++ b/utils/pkg/dberror/dberror.go
@@ -18,6 +18,15 @@ var eofRe = regexp.MustCompile(`\beof\b`)
 // errors.Join or fmt.Errorf("...: %w", ErrTransient)) when the caller knows a
 // failure is self-healing but its message wouldn't be classified as transient
 // by Classify — for example an expected, retryable upstream miss.
+//
+// Two limits, both of which fail safe (page sooner rather than suppress):
+//   - IsTransient checks context.Canceled/DeadlineExceeded before this marker,
+//     so wrapping a context error with ErrTransient does not make it transient.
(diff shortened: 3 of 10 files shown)


indexer/pkg/rollup/rollup.go
```diff
@@ -199,6 +209,23 @@ func hasBenignTaskProcessingError(keyvals []any) bool {
 	return false
 }
 
+// isTransientActivityError reports whether Temporal's activity-error keyvals
+// carry an Error that dberror classifies as transient (a self-healing upstream
+// blip the activity's retry policy will recover). The Error value arrives
```
ben-dz: This comment misstates the mechanism: at the "Activity error." log site (SDK `internal_task_handlers.go:2492`) the `Error` keyval is the *raw* error the activity returned — conversion to `ApplicationError` happens afterward, in `convertActivityResultToRespondRequest`. The conclusion (classification works) is right, but a reader could wrongly infer `errors.Is`-based checks won't work here. Reword to "the raw error the activity returned; message-based classification would also hold if a converted error ever appeared here."

utils/pkg/dberror/dberror.go
```diff
@@ -13,6 +13,13 @@ import (
 // eofRe matches "eof" as a standalone word in a lowercased error message.
 var eofRe = regexp.MustCompile(`\beof\b`)
 
+// ErrTransient is a sentinel that explicitly marks an error as transient for
```
ben-dz, line 16: The doc invites wrapping any self-healing failure, but omits two real limits: (a) `IsTransient` checks `context.Canceled`/`DeadlineExceeded` before this sentinel (lines 48-56), so wrapping a context error with `ErrTransient` will not classify transient; (b) the marker does not survive Temporal failure serialization — `errors.Is(reconstructed, ErrTransient)` is false after the `ErrorToFailure`/`FailureToError` round-trip, so a future use across an activity boundary that propagates to Temporal silently loses the classification. Both fail safe (page sooner). Worth a sentence each so the next caller isn't surprised.

ben-dz (commented):
The change is correct and effective for both incidents; control flow is unchanged, the `%w` chain preserves `errors.Is(err, ErrTransient)` to the activity escalator, and sustained failure still pages via the workflow-side Escalator. Two non-blocking items:

**Medium** — The sibling `solingest` worker still pages on the same transient ClickHouse blip class this PR targets: `solingest.go:141`'s `temporalLogger.Error` has no transient demotion and its activities return errors to Temporal, so a `read: EOF` there trips the same `level="ERR"` pager. Likely a deliberate scope decision (solingest had no incident); mirror the demotion if the intent is to quiet the class broadly.

**Low** — (1) The `isTransientActivityError` comment (`rollup.go:214`) says the error "arrives wrapped in the SDK's ApplicationError"; at the "Activity error." log site it is the raw activity-returned error (conversion happens after logging). (2) The `ErrTransient` doc comment (`dberror.go:16`) omits that the sentinel is checked *after* the context-error guard, and that it does not survive Temporal failure serialization.

- `indexer/pkg/solingest/solingest.go:141` — medium: Sibling-parity gap: this `temporalLogger.Error` only demotes context cancellation, unlike the rollup copy patched in this PR. solingest activities return their ClickHouse-backed errors directly to Temporal (`activities.go:26`), so a transient `read: EOF` from ClickHouse Cloud produces the same per-attempt `Activity error.` at ERROR and trips the same aggregate `level="ERR"` pager this PR is meant to quiet. If the goal is to quiet the class rather than just the two incident components, mirror the `isTransientActivityError` demotion branch here.

ben-dz:
Thanks @ben-dz — addressed all three in 6663245:

- **Medium (solingest parity):** Mirrored the transient `Activity error.` demotion into `solingest`'s `temporalLogger.Error` (+ helper). Confirmed its activities return ClickHouse-backed errors to Temporal, so the same `read: EOF` class tripped the pager there. Added a standalone logger test (solingest has no ClickHouse-backed `TestMain`, so it runs without a container). Left `dzingest` alone since it swallows errors via `refresh()` and never emits the SDK line.
- **Low (1):** Reworded the `isTransientActivityError` comment — at the log site the `Error` keyval is the raw activity-returned error (conversion to `ApplicationError` happens after logging); classification would also hold if a converted error appeared.
- **Low (2):** Expanded the `ErrTransient` doc with both limits — checked after the `context.Canceled`/`DeadlineExceeded` guard, and does not survive Temporal's `ErrorToFailure`/`FailureToError` round-trip — noting both fail safe (page sooner).

indexer/pkg/rollup/rollup.go
ben-dz: Fixed in 6663245. Reworded to state the Error keyval here is the raw error the activity returned (conversion to ApplicationError happens after logging), and that message-based classification would also hold if a converted error ever appeared.

utils/pkg/dberror/dberror.go
ben-dz: Fixed in 6663245. Expanded the ErrTransient doc with both limits: (a) IsTransient checks context.Canceled/DeadlineExceeded before the sentinel, so wrapping a context error with ErrTransient will not classify transient; (b) the marker does not survive Temporal ErrorToFailure/FailureToError serialization, so classification is lost across an activity boundary. Noted both fail safe (page sooner) and that it should be inspected in-process.

indexer/pkg/solingest/solingest.go
```diff
@@ -143,9 +144,36 @@ func (l *temporalLogger) Error(msg string, keyvals ...any) {
...
 		return
 	}
+	// Temporal logs "Activity error." at ERROR on every failed attempt,
+	// including non-final ones the activity's retry policy will recover. These
+	// activities write to ClickHouse and return their errors to Temporal, so a
+	// transient connection blip (a native-TLS read: EOF from ClickHouse Cloud)
+	// self-heals on retry but still trips the aggregate level="ERR" pager.
+	// Demote to WARN; sustained failure still surfaces via the ingestion log.
+	// Mirrors the rollup worker's temporalLogger.
+	if msg == "Activity error." && isTransientActivityError(keyvals) {
```
nikw9944, line 157: This demotion removes the only paging signal for a sustained `RefreshBlockProduction` failure. It's safe for `solana`/`geoip`/`validatorsapp` because the workflow-side escalator observes them 12–60× per run and reaches its threshold of 3 — but `blockProductionEveryN` (60) equals `continueAsNewThreshold` (60), so block production runs only at iteration 0 of each ~hourly run (`workflow.go:74`), and the escalator is a workflow-local recreated on every ContinueAsNew (`workflow.go:50`, restart at iteration 0 at `workflow.go:112`), so its `block_production` count can never exceed 1. Pre-PR a sustained transient-class failure (per `dberror.Classify` that includes permanent misconfigs: `no such host`, `connection refused`, persistent timeouts) produced 3 ERROR lines per hour and paged; post-PR it logs WARN forever while block-production data goes stale. The comment's backstop claim on line 152 doesn't hold either: `log_ingestion_runs` has no alerting consumer (only a success-freshness query in `api/handlers/multicast_delivery.go`), and alerting fires solely on `level="ERR"`.

Cheapest fixes: (a) escalate `block_production` at threshold 1 workflow-side — a failure that survived Temporal's 3 in-activity retries on an hourly activity is already sustained, and this matches pre-PR paging semantics; (b) decouple `blockProductionEveryN` from `continueAsNewThreshold` (e.g. 20) so ≥3 observations land per run; (c) carry escalator counts through the ContinueAsNew argument. Either way, correct the line-152 comment.

nikw9944 (changes requested):
One high finding: the solingest demotion silently removes the only paging signal for a sustained `RefreshBlockProduction` failure — `blockProductionEveryN` (60) equals `continueAsNewThreshold` (60), so the workflow-side escalator (a per-run local recreated on ContinueAsNew) observes block_production at most once per ~hourly run and can never reach its threshold of 3; a sustained transient-class failure (including permanent misconfigs like `no such host`) that paged pre-PR now logs WARN forever, and the comment's claimed backstop (the ingestion log) has no alerting consumer. The rest of the PR is sound: the sentinel is classified in-process before the Temporal boundary, and the demotion is safe for the every-iteration activities whose workflow-side escalators still page.

- **Medium** — `indexer/pkg/dzingest/dzingest.go:175`: dzingest's temporalLogger lacks the transient demotion, and its backfill activities are the one place the new sentinel-wrapped error actually reaches Temporal. `BackfillPermissionEvents` returns `BackfillRefresh`'s error straight to Temporal (`backfill_permission_events.go:73-83`, `MaximumAttempts: 2`), and `BackfillRefresh` shares `decodeChunk` and the same near-tip gate (`view.go:356-361`, `view.go:377`), so a near-tip not-found during a backfill logs "Activity error." at ERROR and pages — exactly the transient-by-design condition this PR demotes. Bounded impact (manual admin-CLI trigger, at most 2 ERR lines per attended run), so non-blocking: mirror the same ~10-line demotion into dzingest, or explicitly accept the manual-backfill page.

indexer/pkg/solingest/solingest.go
ben-dz: Good catch — fixed in 2b86937. `block_production` now uses a dedicated threshold-1 Escalator (`bpEsc`), so a failure that survives Temporal's in-activity retries pages on the next observation, restoring pre-PR semantics while the per-attempt retry lines stay demoted. Went with your option (a) rather than decoupling the cadence (b) since it keeps the change to alerting semantics and does not touch the hourly data-refresh interval. Also corrected the line-152 comment that wrongly claimed the ingestion log was a backstop.

ben-dz:
Thanks @nikw9944 — both addressed in 2b86937.

**High (solingest block_production paging):** Fixed via your option (a). `block_production` now observes on a dedicated `logger.Escalator{ErrorAfter: 1, TransientErrorAfter: 1}`, so a failure that survived Temporal's in-activity retries pages immediately — restoring pre-PR paging semantics for the once-per-run activity while the noisy per-attempt retry lines stay demoted to WARN. Corrected the stale comment that claimed the ingestion log was a paging backstop. Left the frequent activities (solana/geoip/validatorsapp) on the shared threshold-3 escalator, which they still reach.

**Medium (dzingest backfill):** Mirrored the ~10-line transient demotion into dzingest's `temporalLogger`. Chose to mirror rather than accept the page, since the near-tip not-found sentinel is transient-by-design and backfill is the one path where it reaches Temporal — leaving it would page an attended backfill on exactly the condition this PR demotes elsewhere. The periodic dzingest activities are unaffected (they return nil via `refresh()`), so there's no block_production-style silent-failure gap here.

Test note: no new test files this turn. The demotion logic is identical to the already-tested rollup/solingest copies (the solingest one runs without a container), the `ErrTransient`→`IsTransient` path is covered by `TestIsTransientSentinel`, and the threshold-1 escalator by `TestEscalator_CustomThresholds`. Happy to add a dzingest logger test or a solingest workflow-level test if you'd prefer explicit coverage there.

nikw9944 (approved):
Re-check on 2026-07-24 after commit 2b86937: both prior findings are verified fixed in the current code. High (solingest block-production paging): block_production now observes on a dedicated `Escalator{ErrorAfter: 1, TransientErrorAfter: 1}` (workflow.go:57,103); confirmed against escalate.go that the first failure logs ERROR regardless of transient classification, and the workflow-side message bypasses the temporalLogger demotion — pre-PR paging semantics restored, and the stale ingestion-log-backstop comment is corrected. Medium (dzingest backfill parity): the transient "Activity error." demotion and helper are mirrored into dzingest.go:186-208 after the context-cancellation branch, so an attended backfill no longer pages on the transient-by-design near-tip gate while non-transient failures still log ERROR. Verified by building the affected packages and running the escalator, sentinel, and solingest logger tests on the new head (all pass; GitHub CI still pending at check time). No new issues: the third copy of isTransientActivityError follows the repo's existing per-package temporalLogger convention, and the untested dzingest copy is byte-identical to logic tested in both sibling packages.

- Both findings from the previous round are resolved in 2b86937 and verified in code: the threshold-1 `bpEsc` restores paging for the once-per-run block-production activity (first `Fail` hits `n >= threshold` per escalate.go, and `Observe`'s lowercase `error` key means the transient branch also resolves to threshold 1), and the dzingest temporalLogger now demotes transient "Activity error." lines identically to rollup/solingest. Non-blocking follow-up noted, not requested: `isTransientActivityError` now has three per-package copies, and the dzingest copy has no local test — extraction to a shared home would remove both concerns if the duplication grows again.

PR #714, opened 2026-07-24
indexer: skip telemetry-usage overlap re-read while catching up so ingest lag converges

### LINKED ISSUE #713: indexer: telemetry-usage catch-up rate is ~1.0x — ingest still pinned ~58 min stale after #711
## Summary

#711 (merged, deployed to prod 2026-07-24 13:32Z) fixed the telemetry-usage advancement bug — ingest no longer inserts nothing, and it recovered a **~47% silent data-density loss** (event-hours went from ~18.4M to ~27M rows, zero duplicates). But its convergence goal — *"`now() − max(event_ts)` should drop from ~3,500–3,800s to single-digit minutes"* — is **not happening**: prod lag is pinned flat at **~57.8 min** (measured over 2h post-deploy; sawtooth 57.7–61).

## Root cause: catch-up rate ≈ 1.0×

Prod logs (mainnet-beta, `capping catch-up window to bound memory` + `refresh completed`):

- Each refresh queries a **10-min window starting 5 min before the watermark** (the late-data re-read overlap) → **net +5 min of new data per cycle**.
- Each cycle takes **~5 min of wall clock** (main Flux query ~70s for 4.6M rows; full refresh 2m20s–4m20s; cadence = activity duration, not the ~1-min reschedule interval, because the refresh serializes).
- Watermark trace: `13:34 → 13:39 → 13:44 → 13:49 → …` (+5 min/cycle) while `target=now` stays ~55–60 min ahead.

**+5 min data per ~5 min wall = ratio ~1.0 — the 58-min debt never pays down.** #711's fake-clock tests verified advancement per refresh (correct) but couldn't falsify the wall-clock convergence claim.

## Proposed fixes (in preference order)

1. **Skip the 5-min overlap re-read while catching up** (e.g. when `target − queryEnd > 15 min`). The overlap exists to catch late-arriving points at the live edge; a span ~1h old has no late-arrival risk. Net gain doubles to +10 min/cycle (~2×) with **no change to the InfluxDB window size** → converges from ~58 min to single digits in ~1h, then reverts to overlap mode at the live edge.
2. **Profile the non-query cycle overhead.** Of the ~4-min cycle only ~70s is the main Flux query; with baselines now cached (#709), find where the other ~2 min go (delta-debug query? CH insert? second Flux query) and cut it — every minute saved raises the ratio.
3. **Do NOT raise the window beyond 10 min.** The cap bounds InfluxDB memory (#665) and backfill history confirms it: 20–30m chunks repeatedly blew InfluxDB's heap; 10m survived.

## Acceptance

`now() − max(event_ts)` on `fact_dz_device_interface_counters` reaches and holds single-digit minutes in prod (the original #708/#711 goal), with no InfluxDB heap incidents during catch-up.

## Context

- #708 (staleness issue), #711 (advancement fix — worked; convergence claim — didn't), #709 (baseline cache, verified working: query_log shows seeds-only), lake#702 / infra#1837 (tracking).
- Lesson for review: catch-up fixes are throughput inequalities — require the `net_gain / cycle_time > 1` math with *measured* cycle times, and attach a post-deploy metric check to any convergence claim.

nikw9944:
Fixes #713

## Summary of Changes
- Skips the `refreshOverlap` re-read during telemetry-usage catch-up: when the watermark (`maxTime`) is stale enough that everything the refresh span could touch is already past the late-arrival bound, the overlap re-read cannot surface new data, so its budget is spent on new data instead.
- This doubles net catch-up throughput from ~`QueryChunk` (5m) to overlap+chunk (10m) per refresh. After #711 the rate was ~1.0x — each ~5m refresh cycle gained ~5m of data — so a ~58m-stale watermark never converged; it now pays down ~5m of lag per cycle.
- The extended catch-up span is the same total size (overlap+chunk) as a steady-state refresh, so the Flux sub-query count, timeout budget, and peak memory are unchanged.
- Once the watermark gets within the overlap-retention bound (overlap + chunk + overlap = 15m) of now, the overlap re-read resumes and steady-state behavior is identical to before.

## Diff Breakdown
| Core logic   |     1 | +49 / -8    |  +41 |
| Tests        |     1 | +126 / -2   | +124 |
| **Total**    |     2 | +175 / -10  | +165 |

A small, focused change to the refresh windowing logic, backed by ~3x its size in regression tests.

<details>

- `indexer/pkg/dz/telemetry/usage/view.go` — in `Refresh`, drop the overlap re-read and extend the catch-up cap to overlap+chunk when `now − maxTime > ingestSpan + refreshOverlap`; log `overlapSkipped` on the window and cap log lines

</details>

## Testing Verification
- New regression test `CatchupSkipsOverlapWhenFarBehind` asserts a 30m-stale refresh reads `[maxTime, maxTime+overlap+chunk)` with the same 2 sub-queries as steady state
- New `CatchupOverlapSkipBoundary` pins the exact threshold: at `now − maxTime = overlap+chunk+overlap` the overlap is retained; one minute past it, skipped
- New `CatchupConvergesUnderModeledCycleTime` models the prod incident (58m-stale watermark, ~5m refresh cadence) over 12 cycles and asserts lag converges to the 15m overlap-retention bound and holds — on the old code this loop stayed pinned at ~58m
- Two existing steady-state tests re-seeded to watermarks inside the retention bound so they keep exercising the overlap re-read path

### DIFF
--- a/indexer/pkg/dz/telemetry/usage/view.go
+++ b/indexer/pkg/dz/telemetry/usage/view.go
@@ -224,15 +224,15 @@ const refreshOverlap = 5 * time.Minute
 // margin also covers capped catch-up steps.
 //
 // It must exceed refreshOverlap plus one capped catch-up step (QueryChunk ×
-// maxCatchupChunks, 5m); 15m leaves headroom. ViewConfig.Validate enforces this
-// against the operator-settable QueryChunk so a larger chunk can't silently
-// regress every refresh to a full re-scan; the runtime signal is
+// maxCatchupChunks, 10m); 20m leaves headroom. ViewConfig.Validate enforces
+// this against the operator-settable QueryChunk so a larger chunk can't
+// silently regress every refresh to a full re-scan; the runtime signal is
 // doublezero_data_indexer_clickhouse_baseline_query_total.
 //
 // This bounds only the backward direction (windowStart below the watermark).
 // The forward direction is bounded separately, by comparing ClickHouse's max
 // event_ts against the watermark (see queryBaselineCountersFromClickHouse).
-const baselineCacheMaxLag = 15 * time.Minute
+const baselineCacheMaxLag = 20 * time.Minute
 
 // maxCatchupChunks bounds how much NEW data (past maxTime) a single refresh
 // ingests: maxCatchupChunks × QueryChunk. After downtime maxTime can fall
@@ -240,30 +240,39 @@ const baselineCacheMaxLag = 15 * time.Minute
 // capping it spreads the catch-up over successive refreshes (maxTime advances
 // after each) so peak memory stays near steady state.
 //
-// The cap is anchored at maxTime, not queryStart (see Refresh). In steady state
-// queryStart = maxTime − refreshOverlap, so the ingested span is the overlap
-// plus one chunk (~10m) = 2 Flux queries per refresh: one re-reading the
-// overlap (dedups out) and one ingesting the new chunk. maxTime advances one
-// chunk per refresh.
+// The cap is anchored at maxTime, not queryStart (see Refresh). queryStart
+// always sits refreshOverlap behind maxTime — the overlap re-read is
+// load-bearing, not just late-arrival capture. Non-sparse counters (in-octets,
+// in-pkts, …) have no ClickHouse baseline; re-reading each key's
+// recently-written rows seeds lastKnownValues/firstRowSeen via the dedup path
+// in convertRowsToUsage, so the first NEW row per key emits a correct delta.
+// Without the re-read, every key whose latest written row sits behind the
+// global maxTime — the norm for an unsynchronized multi-device fleet — would
+// have its first in-window row consumed as a baseline and silently dropped,
+// undercounting its traffic. Catch-up throughput must therefore come from this
+// forward cap, never from skipping the overlap.
+//
+// Two chunks, not one, so catch-up converges (#713): the measured prod refresh
+// cadence is ~5m per cycle, and a 1-chunk (5m) cap gained ~1.0× wall clock per
+// cycle — a stale watermark never caught up. Two chunks pay lag down ~5m per
+// cycle. The cost, paid only while the cap binds (maxTime more than 2 chunks
+// behind now): a capped refresh spans overlap + 2 chunks (~15m, 3 Flux
+// queries) versus steady state's ≤ overlap + lag (~10m, ≤2 Flux queries) —
+// ~1.5× steady-state peak memory and one extra Flux query.
 //
 // The cap must keep each refresh inside the dzingest activity's
 // StartToCloseTimeout. Each chunk is a separate Flux query bounded only by the
 // client's per-request HTTP timeout (defaultFluxHTTPTimeout, 4m); the Flux
-// query does not abort on the activity context deadline, so the 2-query steady
-// state can spend up to ~8m in InfluxDB alone. RefreshTelemetryUsage runs under
-// a dedicated 10m StartToCloseTimeout (see dzingest/workflow.go) that bounds
-// this worst case plus the ClickHouse dedup/baseline/insert work, so ctx is not
-// already expired when InsertInterfaceUsage runs and maxTime advances every
-// cycle. (History: a 3-chunk/15m span under the old 5m deadline overran on both
-// mainnet-beta and testnet — ctx expired before the insert, every catch-up
-// refresh failed, and the window stayed pinned at the cap in an unrecoverable
-// loop.)
-//
-// Catch-up refreshes that skip the overlap re-read (see Refresh) cover the
-// same overlap+chunk span with the same 2 Flux queries — only the split
-// between re-read and new data changes — so the timeout and memory math above
-// is unchanged.
-const maxCatchupChunks = 1
+// query does not abort on the activity context deadline, so a capped 3-query
+// refresh can spend up to ~12m in InfluxDB alone. RefreshTelemetryUsage runs
+// under a dedicated 15m StartToCloseTimeout (see dzingest/workflow.go) that
+// bounds this worst case plus the ClickHouse dedup/baseline/insert work, so
+// ctx is not already expired when InsertInterfaceUsage runs and maxTime
+// advances every cycle. (History: a 15m span under the old 5m deadline overran
+// on both mainnet-beta and testnet — ctx expired before the insert, every
+// catch-up refresh failed, and the window stayed pinned at the cap in an
+// unrecoverable loop. The deadline, not the span, was the bug.)
+const maxCatchupChunks = 2
 
 type View struct {
 	log       *slog.Logger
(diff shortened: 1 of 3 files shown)


indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -373,10 +378,36 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error)
...
+			// further behind maxTime either), so the re-read cannot surface
+			// anything new. Spending its budget on new data instead doubles net
+			// catch-up throughput from ~QueryChunk to ~ingestSpan per refresh —
+			// at 1× (+5m of data per ~5m refresh cycle) a stale watermark never
+			// converged (#713) — with the same span size, Flux query count, and
+			// peak memory as steady state. The final catch-up span's tail is
+			// still re-read by the first steady-state refresh's overlap once
+			// maxTime gets close enough to now.
 			overlap := refreshOverlap
+			if now.Sub(*maxTime) > ingestSpan+refreshOverlap {
```
ben-dz: Skipping the overlap re-read silently corrupts the catch-up region this PR backfills. The premise that the re-read "cannot surface anything new" / "only the split changes" is wrong — the overlap has a second load-bearing job.

**Non-sparse baseline continuity.** Non-sparse counters (`in-octets`, `in-pkts`, `out-octets`, …) have no ClickHouse baseline (`CounterBaselines` carries only sparse error/discard counters, view.go:1000-1015). Their delta continuity depends entirely on rows in the read window. In overlap mode, `queryStart = maxTime − refreshOverlap` re-reads each active key's recent rows; the dedup path seeds `lastKnownValues` and sets `firstRowSeen[key]=true` (view.go:959-974), so the first new row emits a correct delta. In skip mode `queryStart = maxTime`, so `alreadyWritten` (filtered `event_ts >= maxTime`, store.go:167) contains only keys with a row at *exactly* the global max. Every other key — every key on a device whose latest scrape is even milliseconds behind the global max, the norm for an unsynchronized multi-device fleet — hits `isFirstRow && hasNonSparseValues` (view.go:1020-1044) and its first row in `[maxTime, maxTime+10m)` is consumed as a baseline and never inserted. That row is missing from `fact_dz_device_interface_counters` and the traffic over `[key's last written row, swallowed row]` appears in no delta row — a silent undercount, recurring once per lagging key per skip refresh across the whole recovered region.

**Late arrivals at a stall transition.** Same root cause: on a real-time→stall transition (last steady refresh read to ≈now, then a >15m stall), a point timestamped in `[maxTime−5m, maxTime)` arriving 1–4m late is read by no refresh — the first post-stall skip refresh starts at `maxTime` and the watermark only moves forward. Up to `refreshOverlap` of late telemetry lost per stall.

The new tests miss both: `captureIntfCounterWindows` returns zero rows, so only the query window is asserted, never emit/convert.

Fix: keep `queryStart = maxTime − refreshOverlap` during catch-up so overlap rows seed baselines and catch late arrivals; if extending the forward cap on top of that breaks the Flux/timeout budget, surface that tradeoff explicitly. Add a test seeding a key whose latest timestamp is behind the global `maxTime` and assert its first in-window row is emitted, not swallowed.

ben-dz (changes requested):
The convergence mechanism and its throughput/memory math are sound, but the change rests on a wrong premise: it treats the overlap re-read as pure waste during catch-up ("cannot surface anything new", "only the split changes"). The overlap actually does two things that don't dedup out, and skipping it silently corrupts exactly the catch-up region this PR backfills.

**H1 — first row per lagging key is dropped and traffic undercounted (view.go:407).** Non-sparse counters (`in-octets`, `in-pkts`, …) have no ClickHouse baseline; their delta continuity depends entirely on rows in the read window. In overlap mode the recently-written overlap rows seed `lastKnownValues`/`firstRowSeen` via the dedup path (view.go:959-974), so the first new row emits a correct delta. In skip mode `queryStart = maxTime`, so `alreadyWritten` (filtered `event_ts >= maxTime`, store.go:167) contains only keys with a row at exactly the global max. Every other key — i.e. every key on a device even milliseconds behind the newest scrape, the norm for a multi-device fleet — hits `isFirstRow && hasNonSparseValues` (view.go:1020-1044) and its first in-window row is consumed as a baseline and never inserted. Result: a missing row plus an undercounted traffic interval per key per skip refresh, systematically across the catch-up region. Pre-PR this common catch-up case used the overlap and was lossless.

Same root cause, secondary impact: the overlap also catches genuinely late-arriving data (timestamp < maxTime, within the 5m late bound). On a real-time→stall transition, the first post-stall skip refresh starts at `maxTime` and the watermark only moves forward, so up to `refreshOverlap` of late telemetry is lost per stall.

The new tests can't see either impact — `captureIntfCounterWindows` returns zero rows, so they assert the query window only, never the emit/convert path.

Fix: don't drop the overlap re-read during catch-up — it provides non-sparse baseline continuity and late-arrival capture. Keep `queryStart = maxTime − refreshOverlap`; if extending the forward cap on top of that breaks the Flux/timeout budget, surface and decide that tradeoff rather than silently taking it. Add a test that seeds a key whose latest timestamp is behind the global `maxTime` and asserts its first in-window row is emitted, not swallowed.

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: Fixed in 91a9254 by reverting the skip entirely — you're right that the premise was wrong. The dedup path's seeding of `lastKnownValues`/`firstRowSeen` doesn't happen for keys behind the global `maxTime` when the read starts at `maxTime`, so their first in-window row would be swallowed by the `isFirstRow && hasNonSparseValues` branch.

`queryStart = maxTime − refreshOverlap` is now unconditional; convergence comes from extending the forward cap instead (`maxCatchupChunks` 1 → 2), so a capped refresh covers `[maxTime−5m, maxTime+10m)` — net gain ~2× the measured ~5m cycle. The Flux/timeout tradeoff is taken explicitly: a capped refresh is 3 Flux queries (worst case ~12m), so `telemUsageStartToCloseTimeout` goes 10m → 15m, and `baselineCacheMaxLag` 15m → 20m to keep the cache guard above the larger capped step. Cost is ~1.5× steady-state peak memory, only while the cap binds.

Added the test you asked for (`CatchupEmitsFirstRowOfLaggingKey`): seeds `lag-device` with its latest row 2m behind the global max set by `fast-device`, runs a 30m-stale refresh with InfluxDB rows flowing through the real dedup/convert/insert path, and asserts the lagging key's first new row lands in ClickHouse with its delta computed against the re-read overlap row (not consumed as a baseline). The stall-transition late-arrival loss is covered by keeping the overlap unconditionally.

nikw9944:
Addressed the review in 91a9254 (branch also rebased on main).

**H1 (overlap skip corrupts the catch-up region):** accepted in full — the overlap re-read is load-bearing for non-sparse baseline continuity and late-arrival capture, so the skip is reverted rather than patched. `queryStart = maxTime − refreshOverlap` unconditionally. Convergence for #713 now comes from the forward cap: `maxCatchupChunks` 1 → 2, so a capped catch-up refresh covers `[maxTime−5m, maxTime+10m)` and pays lag down ~5m per ~5m cycle.

**The tradeoff, taken explicitly rather than silently:** a capped refresh is 3 Flux queries (~15m span) instead of 2, so the worst case is ~12m in InfluxDB. `telemUsageStartToCloseTimeout` goes 10m → 15m (worst-case ingest-loop stall grows accordingly; comments in `dzingest/workflow.go` updated with the math), and `baselineCacheMaxLag` 15m → 20m so the cache guard stays above the larger capped step. Peak memory while the cap binds is ~1.5× steady state — far below the whole-window spans that caused the #665 OOM. Steady state (within the cap) is unchanged: ≤2 Flux queries, same span.

**Tests:** the overlap-skip tests are replaced with `CatchupKeepsOverlapAndExtendsCap`, the convergence model test now converges to the refresh cadence (~5m) instead of holding at 15m, and the requested emit-path regression test `CatchupEmitsFirstRowOfLaggingKey` seeds a key whose latest row is 2m behind the global maxTime and asserts its first new row is inserted with the correct delta through the real dedup/convert/insert path (it fails on the previous approach).

Note: this environment has no Docker, so the ClickHouse testcontainer suite couldn't be run locally — relying on CI for the test run.

ben-dz (approved):
lgtm. The prior blocking finding (skipping the overlap re-read silently dropped the first row per lagging key and lost stall-transition late arrivals) is fixed at the source in commit 91a9254: the skip is removed and `queryStart = maxTime − refreshOverlap` is now unconditional, so the overlap re-read that seeds non-sparse counter baselines and catches late arrivals always runs. Convergence is achieved instead by extending the forward cap (`maxCatchupChunks` 1→2), with the dependent constants updated consistently (`baselineCacheMaxLag` 15m→20m so the default config still passes `Validate`; `telemUsageStartToCloseTimeout` 10m→15m to bound the 3-query worst case). A real end-to-end regression test (`CatchupEmitsFirstRowOfLaggingKey`) drives InfluxDB→dedup→convert→ClickHouse and asserts the lagging key's first row is emitted with a correct delta; it runs against a real ClickHouse in CI and passes. The 2-chunk cap costs ~1.5× steady-state peak memory only while the cap binds, which stays well within the window the #665 memory bound guards. No new issues.

PR #716, opened 2026-07-24
shreds: reflect per-escrow activation on subscribers dashboard (max, not sum)

### LINKED ISSUE #715: Lake: shred subscribers dashboard must reflect per-escrow activation (max, not sum)
# Lake: shred subscribers dashboard must reflect per-escrow activation (max, not sum)

## Context

A shred-subscription `ClientSeat` can have multiple `PaymentEscrow` accounts (one per funding
wallet). The oracle and onchain program evaluate activation and renewal **per escrow**: each
funded escrow enters the epoch auction independently, and a charge must be covered by a single
escrow's balance. A seat therefore activates only when its **largest single escrow** covers the
per-epoch price. Escrow balances are never aggregated across accounts by the program or oracle.

The lake dashboard aggregates with `sum(usdc_balance)` per seat. That misled operations this week:
seat `248q3UAAZ3rafTcVUZxBpsfQ7jMse1rBam4svNCZS7E7` (client IP `72.251.3.102`, device
`dz-tor1-sw01`, metro `yto`, price $30/epoch) holds two escrows of $5.83 and $25.65. The UI shows
"Pending, $31.48, 1 prepaid epoch" — implying it will activate at the next epoch — but no single
escrow covers $30, so the oracle will never activate it. Status, balance, and prepaid epochs must
be derived from the greatest single escrow balance so the dashboard matches what the oracle will
actually do. (Related context: malbeclabs/infra#2082 tracks this stuck subscriber; the decision is
to fix the dashboard, not the oracle.)

### API (lake repo)

1. `api/handlers/shreds.go`
   - Both `escrow_balances` CTEs (count query and data query) currently
     `sum(usdc_balance) as total_usdc_balance`. Change the aggregate to `max(usdc_balance)`.
   - The status-filter `prepaidExpr` (used for the active/expiring/pending/inactive buckets)
     consumes the same CTE value — it must use the max as well (follows automatically if the CTE
     changes).
   - Rename the field/JSON key so the name matches the semantics (e.g. `usdc_balance` or
     `spendable_usdc_balance` rather than `total_usdc_balance`), including the Go struct tag, the
     sort/filter field mappings (`"balance"` → column), and every consumer. If the rename ripples
     unreasonably far, keeping the JSON key with a corrected comment is acceptable — but prefer the
     rename; the misleading name is how this bug survived.
   - Optionally (should-have): expose the across-escrow total and escrow count as secondary fields
     (e.g. `escrow_count` already exists on the seat; add `all_escrows_usdc_balance`) so operators
     can still see stranded funds on multi-escrow seats.
2. `api/handlers/shred_subscribers.go`
   - Same change to its two `sum(usdc_balance)` CTEs.
3. Sweep the rest of `api/` for other `sum(usdc_balance)` aggregations over
   `dim_dz_shred_payment_escrows_current` and apply the same reasoning (stats/summary queries in
   `shreds.go` around the seat-stats block included).

### Web (lake repo)

4. `web/src/components/shreds-page.tsx`
   - `prepaidEpochs()` and the derived status chips (`pending` / `expired` / `expiring` / `active`)
     compute from the API balance field — they inherit the fix, but update the field name if
     renamed, and re-check the copy (e.g. the "Allocated for the current epoch with 1+ epochs of
     prepaid balance" tooltip) still reads correctly.
   - The Balance (USDC) column now shows the greatest escrow balance. Where the seat has multiple
     escrows and the secondary total is available, surface it unobtrusively (e.g. tooltip or muted
     subtext "2 escrows · $31.48 total") so residual funds aren't invisible.
5. Sweep other components referencing the balance/prepaid fields (`shred-fund-modal.tsx`,
   `shreds-subscribe-page.tsx`, `shreds-economics-page.tsx`, `access-pass-detail-page.tsx`, …) for
   the same field usage and update consistently. Customer-facing flows that estimate "epochs this
   payment buys" must also assume per-escrow spending.

### Tests

6. Update/extend `api/handlers/v1_shreds_client_seats_test.go` and
   `v1_shreds_subscribers_test.go`: add a fixture seat with two escrows (e.g. $5.83 + $25.65 at a
   $30 price) and assert balance = 25.65, prepaid = 0, and statu

ben-dz:
## Summary of Changes
* Shred subscriber status, balance, and prepaid epochs now derive from the greatest single escrow (`max`), not the sum across escrows — matching how the oracle evaluates activation/renewal per escrow (a charge must be covered by one escrow).
* Fixes the misleading dashboard case: a seat with escrows $5.83 + $25.65 at $30/epoch showed "Pending, $31.48, 1 prepaid epoch" but never activates; it now shows inactive/expired with spendable $25.65 and 0 prepaid.
* Internal handlers rename `total_usdc_balance` → `spendable_usdc_balance` and add `all_escrows_usdc_balance` (the across-escrow total, so operators still see stranded funds). Multi-escrow seats show the total as muted subtext.
* Public `/api/v1/edge/shreds/*` responses add both new fields and keep `total_usdc_balance` as a deprecated alias of the sum — backward compatible.
* Fixes #715

## Testing Verification
* New regression test `TestGetShredClientSeats_MultiEscrowUsesMaxNotSum`: two-escrow seat ($5.83 + $25.65, $30/epoch) asserts spendable = 25.65, all-escrows = 31.48, and status buckets inactive=1 / pending=0 / active=0 (reverting max→sum flips both the balance and pending↔inactive).
* New `TestV1EdgeShredsSubscribers_MultiEscrowMaxNotSum` asserts spendable ≠ all-escrows through the public endpoint, covering the shared `FetchShredSubscribers` CTE (existing v1 fixtures are single-escrow where max == sum).
* Web call-site split verified: survival/at-risk/runway use spendable (max); funds-held aggregates and fund/withdraw displays use the all-escrows sum. DB-backed tests run in CI (Docker/testcontainers unavailable locally); `go build`/`go vet`/`tsc -b` pass.

## Reviewer follow-ups
* **v1 contract (arch High):** kept `total_usdc_balance` as a deprecated alias of the across-escrow sum rather than removing it, so the public unauthed contract stays non-breaking at 1.0.0.
* **Prod data in test (security Low):** replaced the real client IP in the fixture with TEST-NET-3 `203.0.113.7`.
* **Multi-epoch runway (arch/steve Medium/question, disputed→proceeding):** the approved plan dictates deriving prepaid epochs from the greatest single escrow. Per-epoch activation is exact; for a seat with several individually-sufficient escrows, `floor(max/price)` is a conservative lower bound on multi-epoch runway. Comments softened to not overclaim. Whether the oracle round-robins escrows across epochs (which would make `sum_i floor(balance_i/price)` more accurate) is worth confirming, but is out of scope for this dashboard fix and not required by the acceptance criteria.
* **Fund/withdraw modals (Low/minor):** these components are not mounted anywhere in `web/src` (dead code) and their balance display is unchanged from before the fix; left as-is to keep the change minimal.
* **CTE duplication (Low):** the `escrow_balances` CTE is repeated across four queries; extracting a shared constant is deferred as an optional cleanup.


api/handlers/shreds.go
```diff
@@ -263,7 +269,9 @@ func (a *API) GetShredClientSeats(w http.ResponseWriter, r *http.Request) {
 	// Count query.
 	countQuery := `
 		WITH escrow_balances AS (
```
nikw9944: The `escrow_balances` CTE is duplicated verbatim four times (here in the count query, the list query at line 297, and both queries in shred_subscribers.go at lines 20 and 132). The repo already has the pattern for this: `facilitiesEnrichedCTE` / `facilitiesSimpleCTE` package-level constants in api/handlers/facilities.go:82/142, shared across the list and detail handlers. This PR itself had to hand-sync the max-vs-sum fix across all four copies, and the multi-epoch runway semantics are flagged as possibly changing again; a missed copy would silently desync the paginated `total` from the returned items, or the access-pass card from the seats table, with no compile-time or test signal. Extracting `const escrowBalancesCTE` following the facilities.go precedent is small and conventional — worth doing in this PR rather than deferring.

web/src/components/access-pass-detail-page.tsx
```diff
@@ -256,7 +257,14 @@ function ShredsSubscriptionCard({ seat }: { seat: AccessPassShredsSeat }) {
         </div>
         <div className="flex justify-between gap-4">
           <dt className="text-sm text-muted-foreground shrink-0">Escrow Balance</dt>
-          <dd className="text-sm tabular-nums">${balanceDollars.toFixed(2)}</dd>
+          <dd className="text-sm tabular-nums text-right">
+            ${balanceDollars.toFixed(2)}
+            {seat.escrow_count > 1 && (
+              <div className="text-xs text-muted-foreground">
+                {seat.escrow_count} escrows · ${allEscrowsDollars.toFixed(2)} total
```
nikw9944, line 267: This multi-escrow subtext lacks the explanatory tooltip the seats-table copy has (`title="Balances are evaluated per escrow; only the largest single escrow can cover a charge."` in shreds-page.tsx). Add the same `title` so the access-pass card explains the max-vs-sum distinction too.

api/handlers/shreds.go
```diff
@@ -146,7 +152,7 @@ var seatSortFields = map[string]string{
 	"tenure":        "s.tenure_epochs",
 	"active_epoch":  "s.active_epoch",
 	"funder":        "s.funding_authority_key",
-	"balance":       "total_usdc_balance",
+	"balance":       "spendable_usdc_balance",
 	"prepaid":       "price_per_epoch_dollars",
```
nikw9944: `seatSortFields` maps `"prepaid"` to `price_per_epoch_dollars`, so sorting the Prepaid column orders by price, not by the displayed prepaid epochs. Pre-existing, but this PR redefined prepaid semantics everywhere else (filter `prepaidExpr`, TS `prepaidEpochs`, displayed column), leaving this sort as the one stale proxy — and with multi-escrow seats the sort order now diverges from displayed values more often. Sort on the same `prepaidExpr`, or file a follow-up.

nikw9944 (commented):
Two small changes worth making, plus one CI note. The `escrow_balances` CTE is now duplicated verbatim in four queries; the repo already shares SQL fragments as package-level consts (`facilitiesEnrichedCTE` in api/handlers/facilities.go), and this PR itself had to hand-sync the max-vs-sum fix across all four copies — extract `const escrowBalancesCTE` now rather than deferring. Also: the access-pass multi-escrow subtext is missing the explanatory tooltip the seats table has, and the Prepaid column sort still orders by price rather than prepaid epochs (pre-existing, now the one stale proxy after this PR's semantics change). Separately, the `go-test` job failed on three unrelated `TestDeviceMulticastDelivery_*` tests (HTTP 500, looks like test-infra flakiness; this PR's new tests passed) — rerun before merging. The core fix is correct: SELECT/Scan alignment, count-vs-list filter resolution, the v1 deprecated alias mapping, and the spendable-vs-all-escrows split at every web call site all check out, and no consumer of the removed internal `total_usdc_balance` field remains.

- CI: the go-test job on this PR's head commit failed on three tests unrelated to this change — TestDeviceMulticastDelivery_{EmptyState,ObservedStateAndPagination,HealthFiltersUseHealthSemantics} (HTTP 500 after ~33s each; looks like ClickHouse-under-parallel-load flakiness, and test databases are per-test isolated so this PR's fixtures cannot leak into them). The PR's own new tests (TestGetShredClientSeats_MultiEscrowUsesMaxNotSum, TestV1EdgeShredsSubscribers_MultiEscrowMaxNotSum) passed in that same run. Re-run go-test before merging; if the multicast tests keep failing, that needs its own investigation.

api/handlers/shreds.go
ben-dz: Done — extracted `const escrowBalancesCTE` (the CTE body) following the `facilitiesEnrichedCTE` precedent, now shared by all four queries (count + list in shreds.go, and both queries in shred_subscribers.go). Commit 0c3e5d8.

web/src/components/access-pass-detail-page.tsx
ben-dz: Added the same `title` tooltip to the access-pass multi-escrow subtext. Commit 0c3e5d8.

api/handlers/shreds.go
ben-dz: Fixed in this PR. `"prepaid"` now sorts on the shared `seatPrepaidEpochsExpr` (spendable/price epochs) instead of `price_per_epoch_dollars`, so sort order matches the displayed value (commit 0c3e5d8). Added `TestGetShredClientSeats_PrepaidSortAndFilter` pinning it with two seats whose price order and prepaid-epoch order disagree, and while there I also closed the silently-dropped `prepaid:` filter by adding it to `seatFilterFields` (commit 1a53868).

ben-dz:
Thanks — addressed all three inline points plus the CI note.

**Fixed (commits 0c3e5d8, 1a53868):**
- Extracted `const escrowBalancesCTE` (the CTE body) following the `facilitiesEnrichedCTE` precedent; all four seat/subscriber queries now share it, so the max-vs-sum semantics can't desync across copies.
- Added the per-escrow explanatory tooltip to the access-pass multi-escrow subtext, matching the seats table.
- Changed the Prepaid column sort to order by `seatPrepaidEpochsExpr` (epochs) instead of the stale `price_per_epoch_dollars` proxy, so sort order matches the displayed value. Extracted `seatPriceExpr`/`seatPrepaidEpochsExpr` consts (also inlined the leftover `prepaidExpr` local). Added `TestGetShredClientSeats_PrepaidSortAndFilter` pinning both the prepaid sort and — since the const now exists — the previously silently-dropped `prepaid:` filter.

**CI:** re-running go-test; the earlier failures were the unrelated `TestDeviceMulticastDelivery_*` tests (per-test DB isolation rules out fixture leakage), and this PR's own tests passed in that run.

**Not changed (deliberate, per review):** the withdraw modal shows the across-escrow sum as the refundable amount while the tx closes only the connected wallet's escrow. That's pre-existing money-flow behavior (the old `total_usdc_balance` was also the sum) and touches the on-chain withdraw path, so it's out of scope here — worth a follow-up ticket to show the connected wallet's own escrow balance for multi-escrow seats.

nikw9944 (approved):
lgtm — all three review findings remain fixed and verified on the current head (shared `escrowBalancesCTE` const across all four queries, access-pass tooltip added, prepaid sort/filter now use the shared prepaid-epochs expression with a pinning test), and CI including go-test is green. No changes since the previous re-check.

PR #719, opened 2026-07-24
indexer: catch up telemetry-usage watermark gaps instead of silently skipping them

### LINKED ISSUE #718: indexer: telemetry-usage skip-ahead still silently drops data when watermark falls >1h behind (staging: every cycle)
## Summary

The telemetry-usage **skip-ahead still silently discards data** whenever the ingest watermark falls more than `QueryWindow` (1h) behind `now`: instead of catching up, `queryStart` jumps to `now−1h` and the intervening span is never ingested. #711 fixed the *catch-up path within* the window, and #714 made it converge — but the >1h jump remains, and it is actively losing data today.

## Evidence — caught live on staging (2026-07-24, post-#714)

Staging mainnet-beta telemetry-usage cycle takes **~14m14s** (staging CH write slowness) while each cycle ingests at most a 10-min chunk — cycle time > chunk, so the watermark loses ground every cycle, crosses the 1h line, and skip-ahead fires **every cycle**:

```
20:32:49  capping catch-up window  queryStart=19:32:03 queryEnd=19:42:03 target=20:32:03
20:46:18  telemetry/usage: refresh completed  duration=14m14.717s
20:47:21  capping catch-up window  queryStart=19:47:21 queryEnd=19:57:21 target=20:47:21
```

Watermark was 19:42:03 after the first refresh; the next window starts at 19:47:21 (= now−1h) — **the 19:42:03→19:47:21 span (5.3 min) was skipped permanently**. This repeats every cycle: staging is continuously dropping ~30% of its mainnet-beta interface-counter data. (This is the same lossy sawtooth prod exhibited before #711 — prod recovered ~47% density when it stopped.)

## Prod risk

Prod's cycle is ~4m10s, so post-#714 prod converges (measured basis for ~2.4× catch-up) and stays far from the 1h cliff in steady state. But **any ingest outage >1h (deploy freeze, InfluxDB incident, workflow stall) triggers the jump and the whole outage span is silently lost** — even though #714's catch-up could now genuinely recover it at ~2.4× real time (a 3h outage ≈ 75 min of catch-up).

## Proposed fix

1. Remove the `now−1h` jump in favor of bounded catch-up from `maxTime` (the #714 machinery already paces it one capped chunk per cycle, preserving the #665 InfluxDB memory bound). If a hard skip guard is still wanted, make it explicit and LOUD: skip only beyond a much larger horizon (e.g. >24h, aligned with backfill tooling) and log/alert at a level a human sees, since it is data loss.
2. Staging's structural condition (cycle 14m > chunk 10m) makes *any* policy lossy or ever-lagging there — that part is the **staging-CH-capacity** item (lake/infra), not this fix; but with the jump removed, staging would at least lag honestly instead of silently dropping ~30% of rows.

## Acceptance

- An induced >1h watermark gap (staging test or fake-clock regression) is *caught up*, not skipped; no span is silently dropped.
- Any remaining hard-skip path logs at a human-visible level with the exact skipped span.

## Context
#708 → #711 (advancement) → #713/#714 (convergence, merged 2026-07-24). Skip-ahead location per earlier investigation: `indexer/pkg/dz/telemetry/usage/view.go` (~335–343 pre-#711; the jump-to-`now−QueryWindow` branch). Related: staging-CH-capacity item in infra#1837.

nikw9944:
## Summary of Changes

* Telemetry-usage ingest now catches up from the watermark when it falls more than `QueryWindow` (1h) behind, instead of jumping to `now−1h` and silently dropping the intervening span — staging was losing ~30% of its mainnet-beta interface-counter rows this way every cycle, and any >1h prod ingest outage would have lost the whole outage span.
* Catch-up reuses the existing #714 capped-chunk pacing anchored at the watermark, so the #665 InfluxDB memory bound is unchanged; the only remaining skip-ahead is an explicit 24h horizon (`maxCatchupHorizon`) that logs the exact dropped span at ERROR so on-call sees it and can run the admin backfill.
* A new in-memory `sourceEmptyThrough` anchor advances past a capped span only after it has both returned zero ingestible rows and aged out of the 1h late-arrival window: genuine source data gaps traverse at the paced rate (trailing `now−1h`) instead of pinning ingest forever (the old jump was the only escape), while a source stall that later replays buffered data with past timestamps within the window still loses nothing.
* `ViewConfig.Validate` rejects `QueryWindow >= maxCatchupHorizon` (the skip target must stay ahead of the watermark), and the empty-window path now merges the start-of-window baselines into the cache (mirroring the backfill's zero-row branch) so a restart mid-gap can't regress every cycle to the 1-year InfluxDB baseline scan.
* Fixes #718

## Testing Verification

* Full `indexer/pkg/dz/telemetry/usage` suite run with `-race` against a real ClickHouse 26.5 server.
* New regression tests: a 2h watermark gap converges with contiguous read coverage (no skipped span) under the modeled prod cycle time; a >24h-stale watermark skips to the window start with the capped span intact; an aged-out empty capped span advances the catch-up anchor (both for zero rows and for overlap-only rows that dedup out) while a still-young empty span and uncapped steady-state empty windows keep re-reading; the pre-existing #708/#713/#714 catch-up regressions still pass, with the "maxTime older than query window" expectation updated from jump to catch-up.
* Horizon-skip ERROR line verified in test output to carry the exact skipped span, raw ClickHouse `maxTime`, and watermark age.


indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -274,6 +281,17 @@ const baselineCacheMaxLag = 20 * time.Minute
...
 const maxCatchupChunks = 2
 
+// maxCatchupHorizon is the only remaining hard skip-ahead: a watermark older
+// than this jumps to now−QueryWindow and the intervening span is PERMANENTLY
+// LOST, so the skip logs at ERROR (pages on-call) with the exact span — recover
+// it with the admin backfill tooling. Within the horizon the refresh always
+// catches up from the watermark instead (paced one capped span per cycle, see
+// maxCatchupChunks), so an ingest outage shorter than this is recovered rather
+// than dropped: at the measured prod cadence catch-up gains ~2× real time, so
+// even a near-24h gap clears in about half a day. The old jump at QueryWindow
```
ben-dz: **Catch-up only converges while a refresh cycle stays under the 10m cap, and nothing bounds or surfaces the other case.**

A capped cycle advances the watermark by at most `QueryChunk × maxCatchupChunks` = 10m (view.go:587-592) while wall clock advances by the cycle time `C`, so lag closes at `10m − C` per cycle — 1× real time at the measured 5m cadence, not 2×. A near-24h gap therefore takes ~24h to clear, not "about half a day". The PR's own test encodes 1×: `...GapBeyondQueryWindowIsCaughtUpNotSkipped` pays down 120m of lag in 24 cycles of 5m = 120m of wall clock.

The bigger issue is the other side of that inequality: **if `C > 10m` the lag grows every cycle, without bound, until the 24h horizon.** The repo explicitly budgets for `C > 10m` — a capped catch-up runs 3 Flux queries each bounded only by the 4m Flux HTTP timeout, "the worst case is ~12m of InfluxDB plus the ClickHouse dedup/baseline/insert work" under a 15m activity deadline (view.go:272-277, dzingest/workflow.go:26-37). Slow InfluxDB is also what puts the watermark behind in the first place (#665/#671), so the divergence condition is correlated with the trigger.

Concretely, with sustained `C ≈ 12m`, a healthy source and a watermark 2h behind: on `main` the jump pulled the window back to the head every cycle, keeping ingest ~1h fresh while silently dropping older spans; on this branch the watermark falls a further ~2m per cycle, `max(event_ts)` degrades toward 24h stale (the API and dashboards serve day-old interface counters), and then the horizon skip discards up to 24h in one step.

Nothing signals this before the cliff: there is no watermark-lag metric, and `result.SourceMaxEventTS` is set to wall-clock `now` (view.go:670-672), so ingestion-log freshness reads healthy at any lag. The only prod-visible hint is the Info "capping catch-up window" line.

Suggested fix: export the watermark lag as a gauge (or report the real max ingested `event_ts` as `SourceMaxEventTS` instead of `now`), and log WARN when lag fails to decrease across consecutive capped cycles, so on-call sees divergence hours before the horizon ERROR. Also correct the "~2× / half a day" figure here — an operator will use it to decide whether to wait or run the backfill.

indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -197,6 +197,13 @@ func (cfg *ViewConfig) Validate() error {
 		return fmt.Errorf("query chunk %s is too large for the baseline cache: refreshOverlap (%s) + QueryChunk×maxCatchupChunks (%s) must stay below baselineCacheMaxLag (%s)",
 			cfg.QueryChunk, refreshOverlap, cfg.QueryChunk*maxCatchupChunks, baselineCacheMaxLag)
 	}
+	// The beyond-horizon skip lands at now−QueryWindow (see maxCatchupHorizon);
+	// a window at or past the horizon would make the "skip" jump backwards
+	// behind the watermark. It would also disable the empty-span age-out that
+	// unpins genuine source gaps.
+	if cfg.QueryWindow >= maxCatchupHorizon {
```
ben-dz: This guard is one capped span short of enforcing what the comment above it claims. The age-out at view.go:626 needs watermark age > `QueryWindow + QueryChunk×maxCatchupChunks`, but age > `maxCatchupHorizon` is intercepted first at view.go:433. So `--device-usage-query-window 23h55m` is accepted while its age-out window is empty: a genuine source gap pins ingest and re-pages forever — exactly the case the comment says the guard prevents. Use `cfg.QueryWindow + cfg.QueryChunk*maxCatchupChunks < maxCatchupHorizon`.

indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -385,10 +419,41 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error)
...
+		if age := now.Sub(*watermark); age > maxCatchupHorizon {
+			queryStart = queryWindowStart
+			// chMaxTime is the raw ClickHouse watermark: the effective watermark
+			// may sit past it (sourceEmptyThrough), and this log line is the
+			// operator's backfill instruction, so report both ends honestly.
+			chMaxTime := "none"
+			if maxTime != nil {
+				chMaxTime = maxTime.UTC().String()
+			}
+			v.log.Error("telemetry/usage: watermark is beyond the catch-up horizon; skipping ahead — the intervening span will not be ingested once this refresh succeeds; recover it with the admin backfill",
```
ben-dz: This ERROR re-fires on every refresh iteration (60s base interval, `telemUsageEveryN = 1`) and cannot self-clear in this branch: `queryEnd = queryWindowStart + cap` never satisfies the age-out at view.go:626, and `maxTime` only moves on insert. A source that has been dead for >24h therefore pages roughly once a minute, claiming a permanently-lost span that in fact holds nothing to recover.

Repeating the report while the loss is ongoing is a deliberate choice, but CLAUDE.md reserves raw `.Error(...)` for terminal failures and routes periodic/background loops through `logger.Escalator`. Logging only when `skippedTo` advances past the last logged value (or escalator-style rate limiting) still reports every distinct skip without a page per minute.

indexer/pkg/dz/telemetry/usage/view_test.go
```diff
@@ -1725,6 +1730,225 @@ func TestLake_TelemetryUsage_View_Refresh_CatchupConvergesUnderModeledCycleTime(
 		"lag must converge to the refresh cadence and hold, got %s", lag)
 }
 
+// Regression for #718: a watermark gap larger than QueryWindow (but within
+// maxCatchupHorizon) must be caught up, not skipped. On the buggy code
+// queryStart jumped to now−QueryWindow and [maxTime, now−QueryWindow) was
+// permanently dropped — staging lost ~30% of its interface-counter rows this
+// way. Model a 2h ingest outage at the measured prod cycle time and assert the
+// reads cover the whole gap contiguously until lag converges.
+func TestLake_TelemetryUsage_View_Refresh_GapBeyondQueryWindowIsCaughtUpNotSkipped(t *testing.T) {
```
ben-dz, line 1765: This test passes on the new anchor alone: the mock returns zero rows, so every span is capped, empty and aged out, and `sourceEmptyThrough` advances 10m per cycle whether or not the `seedMaxTime` stand-in runs — remove that line and the test still passes, so it does not isolate watermark anchoring. Having the mock return rows inside the span (usage non-empty, so the anchor is never set) would make the contiguity assertions bind on the watermark path the test name describes.

ben-dz (changes requested):
The core fix (catch up from the watermark instead of jumping to `now−1h`) is the right call and the anchor/baseline-cache handling holds up on inspection. Three things to address before merge:

1. **Catch-up convergence is assumed, not bounded.** A capped cycle advances the watermark by at most 10m, so lag closes at `10m − cycle_time` per cycle — 1× real time at the measured 5m cadence, not the 2× the comment claims (a near-24h gap takes ~24h, not "about half a day"). If cycle time exceeds 10m — which the repo explicitly budgets for (3 Flux queries, ~12m worst case, 15m activity deadline) and which is correlated with the same slow InfluxDB that puts the watermark behind — lag grows without bound to the 24h horizon, then up to 24h is discarded at once. `main` kept ingest ~1h fresh in that regime. Nothing surfaces the divergence either: there is no watermark-lag metric and `SourceMaxEventTS` is wall-clock `now`, so freshness monitoring reads healthy at any lag.
2. The new `QueryWindow` guard is one capped span short of enforcing the age-out it claims to protect.
3. The horizon-skip ERROR fires every refresh iteration and cannot self-clear during a >24h source outage.

One test-isolation nit as well: the headline regression test passes on the new anchor alone.

indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -552,11 +609,34 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error)
...
+		// Nothing changed through queryEnd, so the start-of-window baselines
+		// are also the end-of-window state: merge them too (on a cache hit
+		// `baselines` aliases the cache — a self-merge, documented safe in the
+		// backfill's zero-row branch). Merging only `endBaselines` (empty when
+		// zero rows came back) would seed a restarted process's nil cache with
+		// empty maps, and every later refresh would "hit" a 0-key cache and
+		// fall back to the expensive 1-year InfluxDB baseline scan for the
+		// rest of the gap traversal.
+		v.updateBaselineCache(baselines, queryEnd)
 		v.updateBaselineCache(endBaselines, queryEnd)
```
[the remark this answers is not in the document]
nikw9944: Fixed in 52fedb1: the empty branch now skips the merge when `baselines` aliases the cache (`baselines != v.baselineCache`) — the `endBaselines` call still advances the watermark.

indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -552,11 +609,34 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error)
...
+		// ingestible rows AND has aged out of QueryWindow held nothing for the
+		// full late-arrival window — advance sourceEmptyThrough so the next
+		// refresh anchors past it, traversing the gap one capped span per cycle
+		// (trailing now−QueryWindow). A younger empty span is re-read every
+		// cycle instead: it may only be empty *yet* (source stall), and a
+		// writer replaying buffered data with past timestamps within
+		// QueryWindow must lose nothing. Uncapped (steady-state) empty windows
+		// keep the watermark where it is: the window isn't pinned (it grows
+		// with now), and the wide re-read maximizes late-arrival capture.
+		if capped && queryEnd.Before(queryWindowStart) && queryEnd.After(v.sourceEmptyThrough) {
```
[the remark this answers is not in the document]
nikw9944: Fixed in 52fedb1: the gate is now inclusive (`!queryEnd.After(queryWindowStart)`), so the anchor advances as soon as the span is exactly QueryWindow old.

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: All three points addressed in 52fedb1:

- The `maxCatchupHorizon` comment now states the real arithmetic: lag closes at `(10m − cycle_time)` per cycle, ~1× real time at the 5m cadence, so a near-24h gap takes on the order of a day — with a pointer to the backfill for clearing large gaps faster, and an explicit warning that a sustained cycle time ≥ 10m makes lag grow toward the horizon.
- `SourceMaxEventTS` now reports the real max ingested `event_ts` instead of wall-clock `now` (which also matches the field's documented meaning in `ingestionlog.RefreshResult`), so ingestion-log freshness shows the true watermark lag during catch-up.
- A capped refresh whose watermark lag failed to decrease since the previous capped cycle logs a WARN naming the divergence (`lastCatchupLag`, reset when the cap stops binding; tracked only when the cap is anchored at the watermark, so initial refreshes and horizon skips don't false-positive).

I kept the 10m cap itself unchanged — raising it trades the #665 memory bound and is a separate discussion.

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: Fixed in 52fedb1: the guard is now `QueryWindow + QueryChunk×maxCatchupChunks < maxCatchupHorizon`, with a Validate subtest at exactly your `23h55m` example.

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: Fixed in 52fedb1: the skip logs ERROR once per distinct stalled watermark (`lastHorizonSkipFrom`) and WARN on repeats while the watermark stays put; a watermark that moves and falls behind again is a new loss event and pages again.

I keyed the dedup on `skippedFrom` rather than `skippedTo`: `skippedTo` is `now−QueryWindow`, which advances every cycle during a dead-source stall, so it would still page per minute — the stalled watermark is the identity of the loss event.

indexer/pkg/dz/telemetry/usage/view_test.go
nikw9944: Fixed in 52fedb1: the mock now serves a row every minute across the whole timeline and the test seeds only the watermark row in ClickHouse — every span yields real inserts, `usage` is never empty so `sourceEmptyThrough` is never set, and the contiguity/convergence assertions bind on the watermark path alone (the `seedMaxTime` stand-in is gone).

nikw9944:
Addressed all review feedback in 52fedb1 (branch also rebased onto current main):

- **Convergence honesty and observability** (ben-dz): corrected the `maxCatchupHorizon` comment to the real rate (~1× real time at the 5m cadence; lag grows if a cycle sustains ≥ the 10m cap), `SourceMaxEventTS` now reports the real max ingested `event_ts` so ingestion-log freshness shows true watermark lag, and a WARN fires when lag fails to decrease across consecutive capped cycles — divergence is visible hours before the horizon.
- **Validate guard** (ben-dz): now `QueryWindow + one capped span < maxCatchupHorizon`, with a test at `23h55m`.
- **Horizon-skip paging** (ben-dz): ERROR once per distinct stalled watermark, WARN on repeats. Keyed on `skippedFrom` rather than `skippedTo`, since `skippedTo` advances with `now` every cycle and would still page per minute during a dead-source stall.
- **Test isolation** (ben-dz): the gap regression test now drives progress through real inserts (row-per-minute mock, no `seedMaxTime` stand-in), so its contiguity assertions bind on the watermark path.
- **Copilot**: skipped the baseline self-merge on cache hits; made the age-out boundary inclusive.

Not changed: the 10m catch-up cap itself (raising it trades against the #665 memory bound — separate discussion if wanted). Full package suite passed locally with `-race` against a real ClickHouse 26.5.

ben-dz (commented):
All four findings from the previous round are fixed in `52fedb1` — verified in code, not just claimed: the convergence comment now states the real 1× rate with an explicit divergence warning, `SourceMaxEventTS` reports the real max ingested `event_ts`, the `Validate` guard carries the capped-span headroom, the horizon skip pages once per distinct stalled watermark (keying on `skippedFrom` rather than `skippedTo` is the right call), and the #718 regression test now binds on the ClickHouse watermark alone.

One issue remains in the new divergence WARN: it also fires on cycles that ingest nothing, so an ordinary source outage triggers it on the majority of refreshes with a message that names the wrong cause — and lag in that state cannot actually reach the horizon, because the `sourceEmptyThrough` age-out pins it at ~70m. Skipping the comparison on zero-row cycles fixes it.

indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -513,26 +607,47 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error)
...
 	}
+	// Catch-up only converges while a cycle completes faster than the capped
+	// span (see maxCatchupHorizon); surface divergence before the horizon
+	// cliff. Lag is meaningful only when the cap is anchored at the watermark
+	// (newDataStart advanced past queryStart) — not on an initial refresh or a
+	// horizon skip, where newDataStart is the window start.
+	if capped && !newDataStart.Equal(queryStart) {
+		lag := now.Sub(newDataStart)
+		if v.lastCatchupLag > 0 && lag >= v.lastCatchupLag {
+			v.log.Warn("telemetry/usage: catch-up lag is not decreasing — refresh cycle time is at or above the capped span; ingest is diverging toward the catch-up horizon",
```
ben-dz: **This WARN fires for the whole of any source outage, with the wrong cause.**

The check runs before the InfluxDB query and records `lastCatchupLag` unconditionally, including on cycles that ingest nothing — the zero-row branch below returns without ever resetting it. Scenario: the telemetry source stops producing rows, the indexer is healthy and cycles are fast (empty cycles are the cheapest there are). From the first cycle where lag exceeds 10m, `capped` is true and `newDataStart == watermark != queryStart`, so the comparison runs and lag grows by exactly the cycle time each cycle. Every one of those cycles logs "refresh cycle time is at or above the capped span; ingest is diverging toward the catch-up horizon".

Both halves of that are wrong here. Cycle time is not the problem — the source is empty. And lag cannot diverge toward the horizon in this state: the `sourceEmptyThrough` age-out pins the effective watermark at roughly `QueryWindow` + one capped span (~70m) and holds it there for the outage, which is what it was built to do. The WARN goes quiet only on the cycles where the anchor advances 10m, so it fires on the large majority of refreshes for as long as the source is down, diluting the one signal that is supposed to mean "catch-up is genuinely diverging".

Fix: only compare lag across cycles that actually ingested rows — move the tracking below the `len(usage) == 0` early return and reset `lastCatchupLag` on that path (`capped`, `newDataStart` and `now` are all still in scope there). If you would rather keep it here, the message needs to name the other cause ("no new source data in the capped span") so on-call is not sent after a cycle-time problem that does not exist.

nikw9944: Fixed in cae2a27, as suggested: the tracking moved below the `len(usage) == 0` early return — the comparison now runs only after a capped, watermark-anchored cycle that actually inserted rows, and the empty branch resets `lastCatchupLag`. Added `TestLake_TelemetryUsage_View_Refresh_CatchupDivergenceWarn` with both sides: a 15m-cycle ingesting catch-up fires the WARN, a source outage with fast empty cycles stays quiet.

nikw9944:
Addressed the new finding on the divergence WARN in cae2a27: `lastCatchupLag` is now tracked only across capped, watermark-anchored cycles that actually ingested rows — the zero-row branch resets it, so a source outage (lag growing because there are no rows, bounded by the `sourceEmptyThrough` age-out) no longer fires the "cycle time" WARN. New regression test covers both the genuine-divergence and outage cases with a captured log.

Branch is up to date with main (no new base commits this round); full package suite passed with `-race` against a real ClickHouse locally.

ben-dz (approved):
All findings are addressed and verified in the current code. `cae2a27a` moves the catch-up divergence tracking below the zero-row early return and resets `lastCatchupLag` on the empty path, so the WARN no longer fires through source outages and now means what it says — traced both paths in the current code, and the new test covers each direction against a real log buffer. The earlier rounds' fixes still hold at this head: the convergence comment states the real 1× rate, `SourceMaxEventTS` reports the real max ingested `event_ts`, the `Validate` guard carries the capped-span headroom, the horizon skip pages once per distinct stalled watermark, and the #718 regression test binds on the ClickHouse watermark alone.

lgtm.

PR #721, opened 2026-07-25
indexer: bound and cache link-latency previous-RTT lookups

### LINKED ISSUE #720: indexer: link-latency previous-RTT lookup scans full history (argMax, no time bound) — now top lake CPU consumer (~9%)
## Summary

With the `argMaxIf` baseline fixed (#702 → #709, verified), the **link-latency previous-RTT lookup is now the top lake-owned CPU consumer on the shared ClickHouse cluster**: **~243K CPU-sec/day (~9% of cluster), reading 21.65 TiB/day across only 239 runs (~93 GiB and ~1,000 CPU-sec per run)** — 24h `system.query_log`, 2026-07-25.

## The queries

`indexer/pkg/dz/telemetry/latency/store.go`:

- `getPreviousDeviceLinkRTTBatch` (line ~45): latest RTT per device-link circuit —
  ```sql
  SELECT origin_device_pk, target_device_pk, link_pk,
         argMax(rtt_us, (epoch, sample_index, event_ts)) as rtt_us
  FROM (SELECT ... FROM fact_dz_device_link_latency
        WHERE rtt_us > 0 AND ((origin=? AND target=? AND link=?) OR ...))
  GROUP BY origin_device_pk, target_device_pk, link_pk
  ```
  **No time predicate at all** — every call scans the circuit's full history to answer "what was the last RTT." Same class as the fixed #702 baseline query, but with zero lookback bound (the fact table only grows, so this gets worse monotonically).
- `getPreviousInternetMetroRTTBatch` (line ~210): identical pattern over `fact_dz_internet_metro_latency`.
- Companion: the `max(sample_index)` query over `fact_dz_device_link_latency` (`GROUP BY origin, target, link, epoch`; qhash 7198007588618657322) — ~40K CPU-sec/day over 2,144 runs, another latest-state-by-scan.

## Fix options (compose; mirror the #709 playbook)

1. **Bound the lookback** — latency samples are written every refresh cycle, so the latest RTT is within hours; a `event_ts >= now() - INTERVAL 1 DAY` guard (with an unbounded fallback only when a circuit returns empty, e.g. new/quiet circuits) cuts the scan by orders of magnitude. One-line-ish, no schema.
2. **Watermark/in-memory carry-forward** — the refresh itself writes the newest RTTs; keep them in memory (à la #709's watermark cache) so steady state needs no CH read at all. The unbounded query then runs only at startup.
3. **Durable: materialize current state** — incremental MV into `AggregatingMergeTree` keyed by circuit with `argMaxState(rtt_us, (epoch, sample_index, event_ts))`, so "latest RTT per circuit" is an O(circuits) `argMaxMerge` read. Also answers the `max(sample_index)` companion. Kills the cost for good, including cold starts.

## Sizing

~243K + ~40K CPU-sec/day ≈ **~11% of current cluster CPU** (~2.6M/day total). Fixing it (plus co-tenant hyperliquid#201, ~10%) would take the cluster down another ~20%.

## Context

- Ranking + measurements: infra#1837 (2026-07-25 read).
- Playbook precedent: #702 → #709 (same shape: latest-value-by-full-scan → bounded + cached; verified in prod, ~20% of cluster reclaimed).
- Related backlog: vote-credits `ROW_NUMBER` (~159K/day) is the next lake item after this.

nikw9944:
Resolves: #720

## Summary of Changes

- The previous-RTT lookups in the latency store no longer scan each circuit's full history on every append — the top lake-owned CPU consumer on the shared ClickHouse cluster (~243K CPU-sec/day, ~9%, reading ~93 GiB per run). The last non-zero RTT per circuit is carried forward in memory, so steady-state appends compute IPDV with **no ClickHouse read at all**.
- Cache misses (indexer restart, first sight of a circuit) are seeded by a 2-day-bounded scan (matching #709's `baselineLookback`); circuits the bounded pass misses fall back to an unbounded query restricted to just those circuits, so quiet circuits keep their carried-forward IPDV baseline. Circuits with no non-zero RTT anywhere are negative-cached so the fallback runs at most once per circuit per process.
- The companion `max(sample_index)` queries (~40K CPU-sec/day) get a 4-day `event_ts` bound so ClickHouse can prune partitions, backstopped by a second in-memory carry-forward: `GetExisting*MaxSampleIndices` return the union of the bounded scan and a per-circuit-epoch cache fed by both scan results and successful writes. A circuit-epoch that outlives the bound while still being fetched (an agent quiet for days, or an epoch running longer than 4 days) costs one full-tail refetch per process start instead of one per refresh; cached epochs two or more behind the newest written epoch are pruned since callers only fetch the current and previous epoch.
- Adds `doublezero_data_indexer_clickhouse_prev_rtt_query_total{dz_env,table,kind}` and `..._prev_rtt_fallback_circuits_total{dz_env,table}` so the fix is verifiable in prod: both should be ~0 in steady state, and the circuits counter distinguishes one quiet circuit from new-circuit churn.
- Behavior notes: backfill IPDV baselines now carry forward across epochs within a run instead of re-reading latest-overall per epoch (slightly more correct — the old read could seed from a *future* value during historical backfills). A batch containing only samples older than the cached value anchors the next IPDV on that batch's last sample rather than the newest row overall; this rare, one-sample divergence is documented on the cache. The bounded prev-RTT seed assumes `event_ts` is monotone with `(epoch, sample_index)`, which holds for well-formed agent headers; the assumption and its blast radius (one stale `ipdv_us` per malformed circuit per restart) are documented on the query.

## Diff Breakdown
| Core logic   |     1 | +383 / -64  | +319 |
| Scaffolding  |     5 | +31 / -0    |  +31 |
| Tests        |     3 | +439 / -7   | +432 |
| Docs         |     1 | +1 / -0     |   +1 |
| **Total**    |    10 | +854 / -71  | +783 |

Test-heavy diff: the logic is concentrated in the latency store; everything else is metric definitions and `DZEnv` wiring.

<details>

- [`indexer/pkg/dz/telemetry/latency/store.go`](https://github.com/malbeclabs/lake/pull/721/files#diff-7c5bc193563b2245a44456cfa90bf2a406f1e7c0e94d1d45cd0f5f2febddde82) — both carry-forward caches (prev-RTT and max-sample-index), bounded seeding with targeted unbounded fallback, negative-cache sentinel, bounded max-index scans with cache union and epoch pruning
- [`indexer/pkg/dz/telemetry/latency/prev_rtt_cache_test.go`](https://github.com/malbeclabs/lake/pull/721/files#diff-f1393ec357a8c75da8382f28a3f70bf86d1da2ccb017614090ef2b3b5ddc236d) — new testcontainers suite pinning cache-hit, seeding, fallback, negative-cache, and max-index carry-forward behavior
- [`indexer/pkg/dz/telemetry/latency/view_test.go`](https://github.com/malbeclabs/lake/pull/721/files#diff-b5cd1843ab923c51c2ab1e7b970730036f34d6a45cc8619b5ffc129090333ff1) — incremental-append tests use recent mock timestamps so the second refresh exercises the bounded scan rather than the same-process cache
- [`indexer/pkg/metrics/metrics.go`](https://github.com/malbeclabs/lake/pull/721/files#diff-0657a67b972ffa8b61207d587788b27d467d2e0025024e562440ad48fc1b1462) — the two new prev-RTT counters

</details>

## Testing Verification

- New `prev_rtt_cache_test.go` suite runs against a real ClickHouse (testcontainers) and asserts on Prometheus counter deltas, so "no query happened" is verified directly, not inferred: a second append on the same store computes IPDV with zero previous-RTT reads; a fresh store (restart path) seeds via one bounded query without the fallback; a circuit quiet longer than the 2-day lookback resolves through the targeted fallback with the correct carried-forward IPDV; a brand-new circuit gets NULL IPDV; an all-loss circuit is queried once, not per append.
- Max-sample-index carry-forward tests pin all four sides of the new behavior: a store retains a circuit-epoch whose rows have aged past the 4-day bound (no recurring refetch), a fresh store drops it exactly once (the per-process-start refetch), a re-append restores it, and epochs two or more behind the newest written epoch are pruned. Internet-metro mirrors cover the second table's paths.
- Cache merges are gated on `WriteBatch` success, so a failed write's retry re-seeds from pre-batch state; the existing idempotency and incremental-append suites continue to cover the write path around the new caches.

## Post-merge verification plan

`system.query_log` runs/day for the prev-RTT and max-sample-index query shapes should drop to ~zero outside restarts; both new counters flat in steady state.


indexer/pkg/dz/telemetry/latency/store.go
```diff
@@ -7,14 +7,38 @@ import (
...
+// this bound).
+const prevRTTLookback = 2 * 24 * time.Hour
+
+// maxSampleIndexLookback bounds the max-sample-index scans to enable partition
+// pruning. Callers only fetch the current and previous epoch; Solana epochs
+// are ~2-3 days, so 4 days covers the previous epoch's boundary row. A key
+// dropped by the bound makes the caller pass existingMaxIdx=-1 and refetch the
+// epoch's full sample tail from the chain; ReplacingMergeTree dedups the
+// re-insert, so the worst case is a redundant RPC fetch, not data loss.
+const maxSampleIndexLookback = 4 * 24 * time.Hour
```
ben-dz, line 42: The claim above — that a dropped key causes one redundant refetch that ReplacingMergeTree dedups — does not hold. The refetch **recurs every refresh cycle**, and each one rewrites a stored `ipdv_us` row.

Traced: `GetExistingMaxSampleIndices` drops any circuit-epoch whose rows all have `event_ts < now-4d`. `view_device.go:117-121` then sets `existingMaxIdx = -1` and refetches the full tail from index 0. Those samples are re-inserted with the *same* recomputed `event_ts` (`view_device.go:155-160` derives it deterministically from `hdr.StartTimestampMicroseconds + i*interval`), so the key is dropped again next cycle. The loop runs every `RefreshInterval` until the epoch leaves `epochsToFetch` — days, not one cycle.

The re-insert is not inert either. `fact_dz_device_link_latency` is `ReplacingMergeTree(ingested_at)` (`migrations/20250117000002_fact_tables.sql:75-77`), so the re-inserted rows win. On each refetch the batch's first sample anchors its IPDV on the carry-forward cache, which after the previous refetch holds that same tail's *last* RTT — so sample 0's `ipdv_us` becomes `|rtt_0 − rtt_N|`, overwriting the correct value written when the sample first arrived.

Two reachable triggers:
1. A circuit whose agent stopped writing >4 days ago while its epoch is still in `epochsToFetch` (`view_device.go:36-38` = current + previous). The on-chain account persists so the tail still returns; down links are a modeled state here (`migrations/20250210000001_link_health_is_down.sql`).
2. Any Solana epoch running longer than ~4 days. `epochsToFetch` always includes the previous epoch, whose newest sample is up to one epoch-length old; at ~2-3 days the margin is under 2x, and exceeding it drops every circuit's previous-epoch key at once — a fleet-wide refetch storm against the cluster this PR is unloading.

The new comment at `view_test.go:770-774` already describes exactly this ("a stale timestamp would make the second refresh see no existing indices and re-insert instead of appending") — that is the production failure mode, fixed only in the mock data.

Fix: carry the max-sample-indices forward in memory the same way this PR carries `prevRTTs` — seed from the bounded query once, merge each circuit-epoch's written max index after a successful append — so a bound-dropped key costs at most one refetch per process. Swapping to an `epoch` bound does not work: `PARTITION BY toYYYYMM(event_ts)` and the `event_ts`-prefixed `ORDER BY` mean only an `event_ts` predicate prunes. Also correct the claim on lines 31-34 and in `CHANGELOG.md:11`.

Test: a circuit-epoch with all samples 5 days old, two consecutive `Refresh` calls against a mock RPC counting `GetDeviceLatencySamplesTail` calls with `existingMaxIdx=-1`; assert the second refresh does not refetch and sample 0's `ipdv_us` is unchanged.

indexer/pkg/dz/telemetry/latency/store.go
```diff
@@ -30,35 +54,138 @@ func (cfg *StoreConfig) Validate() error {
...
+		}
+	}
+	return result, nil
+}
+
+// queryPreviousDeviceLinkRTTBatch gets the most recent RTT for multiple device
+// link circuits in one query. A non-nil since adds an event_ts lower bound so
+// ClickHouse can prune partitions and granules; within a circuit the argMax
+// ordering key (epoch, sample_index, event_ts) is monotone with event_ts, so a
+// bounded result equals the unbounded one whenever the circuit has any row
```
ben-dz, line 226: The "bounded result equals the unbounded one" argument holds only if `event_ts` is monotone with `(epoch, sample_index)` per circuit, but `event_ts` derives from the device-agent-supplied `hdr.StartTimestampMicroseconds`. A circuit whose newest epoch carries a bogus old start timestamp has its true `argMax` row outside the 2-day window while an older-epoch row sits inside it: the bounded pass returns the stale RTT, the fallback never runs, and the circuit gets a silently wrong IPDV baseline after every restart. Sanity-clamp `event_ts` at ingest (reject samples implausibly far outside their epoch) — which also removes trigger 1 of the max-sample-index finding — or state the assumption as conditional on well-formed headers.

ben-dz (changes requested):
The prev-RTT carry-forward cache itself is sound — SQL is fully parameterized, error paths fail closed, the merge-after-successful-write ordering is correct, and the new testcontainers suites do execute in CI.

The blocker is the other half of the PR: the 4-day `event_ts` bound on `GetExistingMaxSampleIndices` produces a *recurring* per-refresh full-tail refetch and re-insert for any circuit-epoch whose samples are all older than 4 days, not the one-time redundant fetch the constant's doc and the CHANGELOG claim. Because the fact table is `ReplacingMergeTree(ingested_at)`, each re-insert also rewrites one row's `ipdv_us` to a wraparound value. Two reachable triggers: a circuit whose agent stopped writing >4d ago while its epoch is still fetched, and any Solana epoch exceeding ~4 days (which drops every circuit's previous-epoch key at once). Carry the max-sample-indices forward in memory the way this PR already carries `prevRTTs`, and correct the claim in the comment and CHANGELOG.

One Medium: the bounded prev-RTT scan's correctness argument assumes `event_ts` monotonicity, but `event_ts` comes from a device-agent-supplied header timestamp.

indexer/pkg/dz/telemetry/latency/store.go
nikw9944: Fixed in 5272bde. The max sample indices are now carried forward in memory the same way `prevRTTs` are: `GetExisting*MaxSampleIndices` merges the bounded scan into a per-circuit-epoch cache (populated from both reads and post-`WriteBatch` writes) and answers from the union, so a key whose rows age past the 4-day bound is never dropped while this process has seen it once. Both triggers (quiet agent, >4-day epoch) now cost at most one full-tail refetch per process start instead of one per refresh. Epochs two or more behind the newest written epoch are pruned since callers only fetch current and previous. The constant's doc and the CHANGELOG now state the residual honestly, including the first re-inserted sample's `ipdv_us` rewrite on that once-per-restart path. The old time-bound tests became carry-forward tests: same-store retention, fresh-store drop (restart path), re-append restore, and pruning.

indexer/pkg/dz/telemetry/latency/store.go
nikw9944: Documented in 5272bde: the comment on `queryPreviousDeviceLinkRTTBatch` now states the monotonicity claim as an assumption about well-formed agent headers rather than an invariant, and names the failure mode (a bogus old start timestamp can leave the true argMax row outside the window, seeding one stale `ipdv_us` per malformed circuit per restart). I chose not to add ingest-time clamping in this PR: the agent owns the header, the blast radius is one sample's IPDV, and trigger 1 of the max-index finding is now removed by the carry-forward cache independently of timestamps. Happy to do clamping as a follow-up if you want it.

nikw9944:
Addressed the review in 5272bde.

**Blocker (recurring stale-epoch refetch):** `GetExistingMaxSampleIndices` / `GetExistingInternetMaxSampleIndices` now back the 4-day bounded scan with an in-memory carry-forward cache of the max sample index per circuit-epoch, populated from both scan results and successful writes (merged post-`WriteBatch`, mirroring the prev-RTT cache) and pruned to the newest two epochs. A circuit-epoch that outlives the bound while still fetched — quiet agent or a >4-day epoch — is now refetched at most once per process start, not every refresh, which also bounds the `ipdv_us` rewrite to that once-per-restart path. The `maxSampleIndexLookback` doc and the CHANGELOG entry now describe this accurately instead of claiming a one-time deduped refetch. The two time-bound tests were replaced with carry-forward tests covering same-store retention, the fresh-store (restart) drop, re-append restore, and epoch pruning.

**Medium (`event_ts` monotonicity):** stated as a conditional assumption on well-formed agent headers in the `queryPreviousDeviceLinkRTTBatch` comment, with the failure mode and blast radius spelled out. I did not add ingest-time clamping here — the carry-forward fix removes trigger 1 independently, and clamping (rejecting samples) is a behavior change better done as a follow-up if wanted.

ben-dz (approved):
Both previously-raised findings are verified fixed in `5272bde`, and the new carry-forward cache does not introduce defects.

The recurring max-sample-index refetch is genuinely resolved: `GetExisting*MaxSampleIndices` now return the union of the bounded scan and an in-memory per-circuit-epoch cache, and both append paths merge written indices after a successful `WriteBatch`, so a bound-dropped key costs one refetch per process start instead of one per refresh. Confirmed independently of the description that the merge is gated on write success (failing safe toward a redundant refetch), that the admin backfill never consults these getters, and that no TTL or delete path exists on the latency fact tables that could desync the cache from the database. The new tests pin same-store retention, fresh-store drop, re-append restore, and pruning. The constant doc and CHANGELOG now describe the residual accurately, including the once-per-restart `ipdv_us` rewrite.

The prev-RTT monotonicity assumption is now documented as an assumption about well-formed agent headers with the blast radius named, which was one of the two remedies offered; deferring ingest-time clamping is a reasonable scope call given the max-index trigger is removed independently.

Checked the new code for introduced problems and found none: the prune predicate only ever evicts strictly-older epochs so a batch-local newest epoch can under-prune but never drop a needed key; the two mutexes are sequential rather than nested and neither is held across ClickHouse I/O; memory is bounded to current plus previous epoch and trimmed on append.

PR #733, opened 2026-07-28
indexer: classify transient S3 failures as transient (#731)

### LINKED ISSUE #731: indexer: SyncMSDP logs ERROR for transient S3 failures (200-with-embedded-error) — classify transient + retry
## Summary

`SyncMSDP` logs ERROR for **transient S3 failures**. Observed on staging (2026-07-27/28): ~26 ERRs in 16.5h, all the same shape:

```
activity refresh failed ... activity=SyncMSDP error="msdp fetch: msdp: list snapshots/ip-msdp-sa-cache-rejected/device=.../date=2026-07-26/:
operation error S3: ListObjectsV2, https response error StatusCode: 200, RequestID: ..."
```

`StatusCode: 200` with an operation error is the classic S3 "200 OK with an error mid-body" — a documented S3 transient that the AWS SDK treats as retryable. These are self-healing and non-actionable, but they page `#alerts-l2` per occurrence (and would page prod `#alerts` if it happens there).

## Fix options (compose)

1. **Classify as transient**: the ingest error logger already downgrades `dberror.IsTransient` causes (per #683/#697); S3 SDK error shapes (`operation error S3`, embedded-error 200s, 5xx, RequestTimeout) aren't in the pattern set. Add them (in `utils/pkg/dberror` or an S3-specific check at the call site, `indexer/pkg/dzingest/activities.go:262-271`).
2. **Bounded retry inside the fetch** (the SDK's retryer may already handle most of these if configured; the ListObjectsV2 paginator context matters).
3. Ensure the activity's failure path goes through the consecutive-failure escalation so only *sustained* S3 unavailability pages.

Same class as the existing "shreds leaf-fetch transient noise" backlog item (S3 transport errors → WARN + optional bounded retry) — worth fixing with the same shape/helper.

nikw9944:
## Summary of Changes
* `dberror.Classify` now treats AWS SDK v2 transient S3 responses — the "200 OK with error mid-body" blip and 5xx server errors — as transient connectivity.
* `SyncMSDP` (and the identically-shaped mroute/ISIS S3 sources) were paging `#alerts-l2` per occurrence on these self-healing blips; the existing ingest escalator now holds them at WARN until sustained (≥10 consecutive iterations, ~10 min), while actionable 4xx (403/404) keep paging.
* Fixes #731

## Testing Verification
* Extended `TestClassifyAndIsTransient` with the exact observed staging string (`operation error S3: ListObjectsV2, https response error StatusCode: 200, ...`), 500/503 variants → transient, and 403/404 plus a non-AWS `statuscode: 200` message → non-transient (still page). Confirmed no other `IsTransient` consumer (ClickHouse/Neo4j/Influx/Anthropic) emits the prefix-scoped AWS SDK message, so classification is unchanged for them.


ben-dz (commented):
Two non-blocking items, both in the new regex's fidelity to the SDK's actual behavior.

`5\d\d` is broader than the transient set the comment above it claims: the SDK's `DefaultRetryableHTTPStatusCodes` is exactly `{500, 502, 503, 504}`, so the pattern also swallows 501 NotImplemented and 505 — permanent endpoint/config failures that would now be held at WARN for ~10 min instead of ~3. Tighten to `(200|500|502|503|504)`.

Separately, the bare `StatusCode: 500/503` test strings are shapes that cannot occur in production: the SDK retries those statuses internally, so a sustained 5xx arrives wrapped in `exceeded maximum number of attempts, 3, ...`. The unanchored regex matches that today, but nothing pins it.

The wiring the PR depends on checks out end-to-end: the `%w` chain from `source_s3.go` through `refresh` reaches the zero-value `Escalator` (transient threshold 10) with the SDK message intact, and 403/404 stay non-transient.

utils/pkg/dberror/dberror.go
```diff
@@ -13,6 +13,17 @@ import (
...
+// awsRespErrRe matches the AWS SDK v2 "https response error StatusCode: ..."
+// shape (lowercased by Classify before matching) for a transient S3 status: a
+// 200 (S3's documented "200 OK with an error mid-body" blip, which the SDK
+// surfaces as a response error) or any 5xx server error (InternalError, 502,
+// 503 SlowDown, 504). The "https response error statuscode:" prefix scopes the
+// match to AWS SDK v2 messages, so ClickHouse, Neo4j, and Influx errors never
+// hit it. Actionable 4xx (403 AccessDenied, 404 NoSuchBucket/NoSuchKey) are
+// deliberately excluded so they keep paging. The trailing \b prevents matching
+// a status inside a longer digit run (e.g. "statuscode: 2001").
+var awsRespErrRe = regexp.MustCompile(`https response error statuscode: (200|5\d\d)\b`)
```
ben-dz: `5\d\d` is broader than the transient set the comment above it claims. The comment says "any 5xx server error (InternalError, 502, 503 SlowDown, 504) — self-healing blips the SDK marks retryable", but the SDK's `DefaultRetryableHTTPStatusCodes` is exactly `{500, 502, 503, 504}` (`aws-sdk-go-v2@v1.42.0/aws/retry/standard.go:53`). `5\d\d` also swallows 501 NotImplemented and 505, which are permanent endpoint/configuration failures rather than blips — they'd be held at WARN for ~10 min (10 consecutive iterations) instead of ~3.

Fix: use `(200|500|502|503|504)` and add a `s3 501 not implemented → ErrorTypeUnknown, non-transient` table case.

While there, drop "the SDK marks retryable" from the 200 rationale: a 200-status `InternalError` is in neither `DefaultRetryableHTTPStatusCodes` nor `DefaultRetryableErrorCodes` (`standard.go:61-64`), so the SDK does not retry it internally. Classifying it transient is still right per S3's own guidance — the comment just shouldn't attribute it to the SDK.

utils/pkg/dberror/dberror_test.go
```diff
@@ -38,6 +38,20 @@ func TestClassifyAndIsTransient(t *testing.T) {
 		// "eof" must match as a word, not an embedded trigram.
 		{"geofence not transient", errors.New("failed to update geofence for device"), dberror.ErrorTypeUnknown, false},
 
+		// AWS SDK v2 transient S3 responses (SyncMSDP staging noise, #731).
+		// The observed shape: a 200 with an error embedded mid-body.
+		{"s3 200 embedded error", errors.New("msdp fetch: msdp: list snapshots/ip-msdp-sa-cache-rejected/device=x/date=2026-07-26/: operation error S3: ListObjectsV2, https response error StatusCode: 200, RequestID: abc123, HostID: def, api error InternalError"), dberror.ErrorTypeConnectivity, true},
+		{"s3 500 internal error", errors.New("operation error S3: ListObjectsV2, https response error StatusCode: 500, RequestID: x, InternalError"), dberror.ErrorTypeConnectivity, true},
```
ben-dz, line 44: These bare `https response error StatusCode: 500/503` strings cannot reach `Classify` in production. The SDK retries 500/502/503/504 internally, so a sustained 5xx surfaces wrapped as `operation error S3: ListObjectsV2, exceeded maximum number of attempts, 3, https response error StatusCode: 500, ...` (`aws/retry/errors.go:12-14`). The unanchored regex does match that today, but nothing pins it, so a future tightening of the prefix would silently break the real 5xx case with a green table. Add one case carrying the `exceeded maximum number of attempts, 3,` wrapper.

utils/pkg/dberror/dberror.go
nikw9944: Fixed in 9ea9cd5: the pattern is now `(200|500|502|503|504)`, matching the SDK's `DefaultRetryableHTTPStatusCodes` exactly, with a `s3 501 not implemented → ErrorTypeUnknown, non-transient` table case. The comment no longer attributes the 200 case to SDK retry behavior — it now notes the SDK surfaces but does not retry it, and that classifying it transient follows S3's own guidance.

utils/pkg/dberror/dberror_test.go
nikw9944: Added in 9ea9cd5: a table case carrying the `exceeded maximum number of attempts, 3,` wrapper (`s3 500 after sdk retries`), with a comment noting it pins the regex staying unanchored so a future prefix tightening can't silently break the real post-retry 5xx shape.

nikw9944:
Addressed both review items in 9ea9cd5:

- Tightened the transient 5xx match from `5\d\d` to the SDK's `DefaultRetryableHTTPStatusCodes` set `(200|500|502|503|504)`, so 501/505 stay non-transient and page on the normal threshold. Added a 501 NotImplemented test case.
- Added a test case for the real production 5xx shape (wrapped in `exceeded maximum number of attempts, 3,`), pinning the unanchored match.
- Corrected the comment's 200 rationale (the SDK surfaces but does not retry a 200-embedded error; transient classification follows S3's guidance) and updated the CHANGELOG entry to name the exact status set.

No behavior beyond the classification set changed. `go test ./utils/pkg/dberror/` passes.

ben-dz (approved):
Both round-1 findings are fixed in `9ea9cd5`, verified in the code rather than from the replies.

The pattern is now `(200|500|502|503|504)\b`, matching the SDK's `DefaultRetryableHTTPStatusCodes` exactly; 501, 505, and 507 no longer classify as transient, and 403/404 plus longer digit runs (`2001`, `5001`, `5040`) still don't match. The comment no longer attributes the 200 case to SDK retry behavior, the 501/505 exclusion is documented, and the CHANGELOG entry was corrected to match. The test table gained the 501 case and a `s3 500 after sdk retries` case carrying the real post-retry `exceeded maximum number of attempts, 3, ...` shape.

No new issues in the follow-up commit. lgtm.

PR #734, opened 2026-07-28
indexer: stop dzingest workflow from double-paging activity failures (#730)

### LINKED ISSUE #730: indexer: workflow-level 'telemetry usage refresh failed' ERROR duplicates the escalation-gated activity error (#697 regression via #711)
## Summary

Every failed `RefreshTelemetryUsage` activity now produces **two ERROR log lines** — the activity-level one (correct: gated by the `consecutive_failures` escalation from #697) and a workflow-level duplicate at `indexer/pkg/dzingest/workflow.go:139`:

```go
logger.Error("telemetry usage refresh failed", "error", err)
```

This is a **regression of the #697 error-noise sweep**, which eliminated exactly this duplicate class (verified on staging 2026-07-17: the `component=temporal` twin shape went to zero). It resurfaced after **#711** set `MaximumAttempts: 1` on the activity — with no retries, every activity failure immediately surfaces to the workflow future and hits this unconditional `logger.Error`.

## Impact

Staging (2026-07-27/28, 16.5h): 56 activity-level ERRs (legitimate, escalation-gated) **+ 50 workflow-level duplicates** — every real page doubled on `#alerts-l2`. On prod the same doubling would apply to any sustained telemetry-usage failure paging `#alerts`.

## Fix

Demote the workflow-level log to WARN (or drop it): the activity-level escalation (`utils/pkg/logger` consecutive-failure threshold) already owns the paging decision for this failure; the workflow layer re-reporting the same error at ERROR adds a page without adding information. Check the sibling futures in the same function (permission-events etc., `workflow.go:142+`) for the same pattern while there.

Per the standing principle (see #696/#697): one failure → at most one alert-bearing line, owned by the layer with the escalation context.

nikw9944:
## Summary of Changes
* DZ ingest workflow logs activity-failure lines at WARN instead of ERROR; removes the workflow-level `logger.Escalator`.
* Every activity failure that reached the workflow future — StartToClose timeouts and panics; ordinary failures never do, since `activities.refresh` swallows them and self-escalates — produced both the activity-level escalated ERROR and a workflow-level duplicate, doubling every timeout-driven page on `#alerts`/`#alerts-l2`. #711's `MaximumAttempts: 1` made every `RefreshTelemetryUsage` deadline miss surface immediately (a #697 regression); the activity layer owns the paging decision, so the workflow re-reporting the same failure added a page without adding information.
* Fixes #730

## Testing Verification
* New `workflow_test.go` regression test drives `DZIngestWorkflow` under the Temporal test suite with every activity mocked to fail, running a span derived from the activity cadence constants (`3*max(EveryN...)`, currently 15 iterations) so every asserted message — including the every-5th-iteration permission events — accumulates >=3 consecutive failures, past the old workflow escalator's ERROR threshold. It asserts every workflow failure message logs at WARN and none at ERROR — the exact class that has now regressed twice (#697 → #711).
* Verified out-of-band (Docker-less env) that under sustained failure the workflow emits all 11 failure messages at WARN (permission events 3×, the rest 15×) and zero at ERROR.

> Note for dashboards/alerts: the workflow `"<name> refresh failed"` / `"<name> sync failed"` lines no longer appear at `level="ERR"`. Anything keyed on them must move to the activity-side `"activity refresh failed"` line, which carries `activity=<name>` and `consecutive_failures`.


indexer/pkg/dzingest/workflow_test.go
```diff
@@ -0,0 +1,123 @@
...
+		a.RefreshShredEscrowEvents,
+		a.SyncISIS,
+		a.SyncIPMroute,
+		a.SyncMSDP,
+		a.SyncGraph,
+		a.RefreshTelemetryUsage,
+		a.RefreshPermissionEvents,
+	} {
+		env.OnActivity(fn, mock.Anything).Return(boom)
+	}
```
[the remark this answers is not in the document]
nikw9944: No change needed: Temporal's test environment matches mocked activities by registered activity name, and both the method expression `(*Activities).RefreshServiceability` and the bound method value `a.RefreshServiceability` resolve to the same name (`RefreshServiceability`). The mocks demonstrably apply — the test passes in CI's `go-test`, and with `Activities{}` holding all-nil dependencies, any real activity execution would panic on the nil `IngestionLog` rather than return the mocked error the assertions observe.

indexer/pkg/dzingest/workflow_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// Start a few iterations short of continue-as-new so the run exercises >3
+	// consecutive failures (past the old escalation threshold of 3) for every
+	// per-iteration activity. Permission events fires once in this range
+	// (iteration%permissionEventsEveryN == 0 at iteration 55); the rest fire
+	// every iteration.
+	env.ExecuteWorkflow(DZIngestWorkflow, continueAsNewThreshold-5)
```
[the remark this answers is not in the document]
nikw9944: Fixed in 63aaaa1: the start iteration is now derived as `continueAsNewThreshold - max(5, telemUsageEveryN, permissionEventsEveryN)`, so the executed range always contains at least one tick of every modulo-gated activity (any window of N consecutive iterations contains a multiple of N) while still exceeding the old escalation threshold of 3.

nikw9944:
Addressed the Copilot review in 63aaaa1:

- **Iteration-span brittleness** (workflow_test.go): fixed — the start iteration is now derived from the activity cadence constants (`max(5, telemUsageEveryN, permissionEventsEveryN)`), so the test keeps scheduling every modulo-gated activity if those constants change.
- **Mocking bound methods vs. method expressions**: no change — Temporal's test env matches activities by registered name, which is identical for both forms. The mocks verifiably apply: the test passes in CI, and the all-nil `Activities{}` would panic (not return the mocked error) if a real implementation ever ran.

Branch is already based on the latest `main`; CI was green before this push and the change is test-only.

indexer/pkg/dzingest/workflow_test.go
```diff
@@ -0,0 +1,124 @@
...
+	} {
+		env.OnActivity(fn, mock.Anything).Return(boom)
+	}
+
+	// Run enough iterations to exercise >3 consecutive failures (past the old
+	// escalation threshold of 3) and to guarantee at least one tick of every
+	// modulo-gated activity: any window of N consecutive iterations contains a
+	// multiple of N, so a span of max(EveryN...) covers them all regardless of
+	// how the constants change.
+	span := max(5, telemUsageEveryN, permissionEventsEveryN)
```
ben-dz: This start iteration makes the `permission events refresh failed` assertion inert. Iterations 55–59 run, and with `permissionEventsEveryN = 5` the activity fires exactly once (55) — so under the old escalator its count would be 1, below the threshold of 3, and it logged WARN there too. Both assertions for that message pass identically before and after the change, contrary to the claim that the run exercises >3 consecutive failures for every asserted line.

The other ten messages do fail on revert, so the test still catches a straight revert. One-line fix: `continueAsNewThreshold-15` (45), so permission events fails at 45/50/55 and all 11 messages cross the old threshold.

ben-dz (commented):
Two things worth fixing, neither blocking.

1. The new regression test's `permission events refresh failed` assertion is inert. Starting at `continueAsNewThreshold-5` runs iterations 55–59, so `RefreshPermissionEvents` fires exactly once — one failure logged WARN under the old escalator too (1 < 3), meaning both assertions for that message pass before and after the change. The other ten messages do fail on revert, so the test still catches a straight revert, but it doesn't cover all 11 lines it asserts. Starting at `continueAsNewThreshold-15` (45) makes permission events fail at 45/50/55 and crosses the old threshold for every message.

2. The CHANGELOG entry says "every `RefreshTelemetryUsage` (and sibling activity) failure had produced both" lines. `activities.refresh` always returns nil to Temporal, so ordinary activity failures never reached the workflow future at all — only StartToClose timeouts and panics double-paged, which is precisely why #711's `MaximumAttempts: 1` mattered. The new comment in `workflow.go` states this correctly, so the CHANGELOG contradicts it. Worth scoping the claim, since this is the durable record for a bug that has regressed twice on exactly this which-layer-sees-what confusion (the PR description has the same overstatement).

The core alerting analysis checks out: all 11 activities route through `activities.refresh` and self-escalate; the sustained-timeout path keeps its threshold-3 page because `dberror.IsTransient` classifies `context.DeadlineExceeded` as non-transient; panics still page via the SDK's non-demoted "Activity error." line; and the `solingest`/`rollup` workflows are not sibling instances of this bug (their views' escalators sit on the background `safeRefresh` path the activities don't use).

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Indexer DZ ingest workflow no longer emits a second ERROR line duplicating the activity-side escalation-gated page: every `RefreshTelemetryUsage` (and sibling activity) failure had produced both the activity-level escalated ERROR and a workflow-level duplicate, doubling every page on `#alerts`/`#alerts-l2` (a #697 regression via #711's `MaximumAttempts: 1`). Workflow-level activity-failure lines are now WARN-only; the activity layer's consecutive-failure escalation owns the paging decision. Any dashboard or alert keyed on the workflow `"<name> refresh failed"` / `"<name> sync failed"` lines at `level="ERR"` must instead key on the activity-side `"activity refresh failed"` line, which carries `activity=<name>` and `consecutive_failures` (#730)
```
ben-dz: "every `RefreshTelemetryUsage` (and sibling activity) failure had produced both" isn't what the code did. `activities.refresh` always returns nil to Temporal (`indexer/pkg/dzingest/activities.go:65-78`), so an ordinary activity failure never reached the workflow future or its escalator — only Temporal-level failures (StartToClose timeouts, panics) produced both lines, which is why #711's `MaximumAttempts: 1` plus the 15m timeout mattered. The new comment at `workflow.go:59-76` states this correctly, so the entry contradicts it.

Worth scoping the claim (e.g. "every activity failure that reached the workflow future — StartToClose timeouts and panics — produced both…"), since this is the durable record for a bug that has now regressed twice on which-layer-sees-which-error. Same overstatement in the PR description.

indexer/pkg/dzingest/workflow_test.go
nikw9944: Fixed in 0d9b23d. The span is now `3 * max(1, telemUsageEveryN, permissionEventsEveryN)` (currently 15, start iteration 45), so permission events fails at 45/50/55 and every asserted message crosses the old threshold of 3 — and stays that way if the cadence constants change. Verified the run: permission events 3× WARN, the other ten 15× WARN, zero of the workflow messages at ERROR.

CHANGELOG.md
nikw9944: Fixed in 0d9b23d. The entry now scopes the claim to failures that reach the workflow future — StartToClose timeouts and panics — and notes that ordinary failures never do because `activities.refresh` swallows them, with #711's `MaximumAttempts: 1` called out as what made every deadline miss surface immediately. Updated the PR description the same way.

nikw9944:
Both findings from @ben-dz's review addressed in 0d9b23d:

- **Inert permission-events assertion**: the test span is now derived as `3 * max(1, telemUsageEveryN, permissionEventsEveryN)` (currently 15 iterations, start 45), so permission events fails 3× and every asserted message crosses the old escalator's threshold — the derivation survives cadence-constant changes, per the earlier Copilot thread. Verified: 3× WARN for permission events, 15× for the rest, zero workflow messages at ERROR.
- **CHANGELOG overstatement**: scoped to "every activity failure that reached the workflow future — StartToClose timeouts and panics" with a note that ordinary failures never do (`activities.refresh` swallows them). The PR description carried the same overstatement and is updated to match.

Nothing declined. Thanks for confirming the alerting analysis — the threshold-3 timeout path, the SDK-owned panic line, and solingest/rollup being out of scope were exactly the load-bearing assumptions.

ben-dz (approved):
Both findings from the previous round are fixed and verified in the current head.

The test's iteration span is now derived as `3 * max(1, telemUsageEveryN, permissionEventsEveryN)` (= 15, starting at iteration 45), so `RefreshPermissionEvents` ticks at 45/50/55 and reaches the old escalator's threshold of 3 — that assertion now discriminates against the regressed behavior along with the other ten, and stays correct if the cadence constants change. The CHANGELOG entry now scopes the double-page to failures that reach the workflow future (StartToClose timeouts and panics) and notes that ordinary failures never do because `activities.refresh` swallows them, matching the code and the comment in `workflow.go`.

The method-expression-vs-bound-method-value question raised on the mock setup is unfounded: the Temporal SDK trims the `-fm` suffix from bound method values (`internal_worker.go:2654`), so both forms resolve to the same registered activity name and the mocks apply. Nothing in the new commits introduces a new issue.

PR #736, opened 2026-07-28
api: cache vote/stake/gossip dashboard scalars and validators listing

### LINKED ISSUE #732: api: vote/stake/gossip dashboard family recomputes cheap answers at high frequency — largest remaining CPU block (~15% of cluster)
## Summary

The lake_api "vote/stake/gossip" dashboard-query family is now the **largest remaining CPU block on the shared ClickHouse cluster: ~245K CPU-sec per 21h ≈ ~15% of total load** (07-28 measurement; cluster total now ~2.2M CPU-sec/day after the recent fixes). No single member cracks the top-3 shapes, but as one *fix-sized unit* it's #1 by ~2×.

The shared diagnosis: **cheap-to-cache answers recomputed at absurd frequencies** — not expensive queries, wrong cadences.

## Members (24h diagnostics from 07-25, cost re-measured 07-28)

| # | Shape (qhash) | Runs/day | Param variance | CPU %/cluster | Diagnosis |
|---|---|---:|---|---:|---|
| 1 | `total_stake` CTE (`12413159124723682455`) | ~9,000 (every ~10s) | 9 variants | ~5.0% | `sum(activated_stake_lamports)` — an **epoch-timescale scalar** — copy-pasted in **five places in `api/handlers/timeline.go`** (~2162, 2364, 2508, 2678, 2842) and recomputed per request |
| 2 | Validators listing (`2239025009229925591`) | ~8,600 | **1 variant** (byte-identical) | ~2.8% | The *same* query every ~10s at ~5.6s avg duration — a >50% duty cycle for one fixed answer; either a fast poller or missing page-cache coverage |
| 3 | Per-validator gossip join (`9329887559298878005`) | ~12,000 | **every run unique** (~490/hr, flat 24/7) | ~3.5% | A per-entity loop: something iterates ~all validators issuing one ~5s query each (grep points at the v1 validators/metadata path — possibly an **external API consumer** polling per-validator) |
| 4 | `current_slot_info` ×2 (`8822595497158257077`, `15845906773532294718`) | ~6,000 each | 19 / 2,036 variants | ~3.6% | `max(cluster_slot)` re-derived per request; slow-changing |

## Proposed structure: one issue, two PRs (different risk profiles)

**PR A — low-risk frequency fixes (members 1, 2, 4):** behavior-preserving handler changes.
- Cache the `total_stake` scalar (~5-min TTL or compute in the page-cache worker); deduplicate the five copies into one helper.
- Find the caller of #2 (its single param variant makes this easy) and give it page-cache coverage or a ≥60s cadence — running it 6× less erases ~90% of its cost.
- Same short-TTL treatment for the `current_slot_info` lookups.
- Expected: most of ~11% of cluster for roughly a day of work.

**PR B — the per-validator loop (member 3):** needs a 15-minute investigation first — check `system.query_log` `initial_address`/HTTP paths to determine whether the driver is internal iteration or an external v1-API consumer. Then either batch the loop into one query (internal) or cache the per-validator endpoint (external). Separate PR because it may change an API path's behavior under load, so its blast radius and rollback story differ from PR A.

## Verification

Each member keeps its own `normalized_query_hash`, so per-shape before/after attribution in `system.query_log` survives combined PRs (proven pattern: #721 fixed two shapes in one PR with clean attribution). Expect the family to drop from ~15% to low single digits. Snapshot the before numbers at PR time — replica scaling events wipe `query_log`.

## Context

Cluster tracking: malbeclabs/infra#1837. Playbook precedents: #702→#709 and #720→#721 (bound/cache the recomputation), #699 (cut refresh frequency), #688 (page-cache tunables). Measurements in this issue: 07-25 param-variance diagnostics + 07-28 21h cost re-measurement.

nikw9944:
## Summary of Changes
* Cache the slow-moving scalars the vote/stake/gossip dashboards recompute on every request — network total stake (5-min TTL) and current cluster slot (15s TTL) — in a per-env in-memory cache instead of re-scanning window-function views; the seven copy-pasted `total_stake` CTEs in `timeline.go` and the two `max(cluster_slot)` subqueries in the multicast members handler now read from it.
* Add page-cache coverage (~60s) for the default validators listing (first page, stake desc, no filters), the shape polled continuously by the UI and an external consumer.
* Attacks the largest remaining CPU block on the shared ClickHouse cluster (~15%) by fixing cadence, not query cost. Behavior-preserving: no response shape changes; the `X-Cache` header on the validators listing is additive.
* Fixes #732

This PR covers the low-risk frequency fixes (issue members 1, 2, 4). The second (edge-scoreboard leader map + v1 validators-metadata) is proposed as follow-up issue #738; its reviewed implementation is preserved on the `nikw9944/lake-732-b` branch.

## Testing Verification
* Scalar-cache unit tests: value cached within TTL, recomputed after expiry, zero never pinned (cold/empty DB serves live), per-env key isolation.
* Default-request detection: both the bare request and the exact polled string `?limit=100&offset=0&sort_by=stake&sort_dir=desc` are cache-eligible; filtered/paginated/ascending shapes bypass. A sort-key parity test pins the query-column map against `validatorSortFields` so a future drift can't silently drop the ORDER BY.
* Worker test asserts the new `validators` entry refreshes every other slow cycle (~60s).
* The container-backed timeline/multicast handler suites (run in CI) exercise the rewritten SQL end-to-end — `stake_share`/`dz_total_stake_share_pct` and multicast member/leader-slot output are unchanged.


ben-dz (commented):
Two small follow-ups; nothing blocking.

**Medium** — the page-cache gate's `DefaultLimit` comparison isn't pinned to the handler's hardcoded `ParsePagination(r, 100)`. They agree today, but the new test re-hardcodes `100` itself, so a future change to the handler default would leave the test green while the cache silently stops hitting and the worker keeps refreshing an entry nobody reads.

**Low** — extracting `fetchValidatorsPage` collapsed three distinct failure logs into one, so a row-scan error and a ClickHouse error are no longer distinguishable in prod logs (the worker refresh path surfaces the same errors).

The risky parts of the change check out: every `%` in the newly-Sprintf'd query literals is an intended verb with a matching arg (including the `>5%` → `>5 pct` fix); the removed `CROSS JOIN total_stake ts` is a row-count no-op at all nine sites with no surviving `ts.` reference; `slot` is wrapped in `toUInt64()` before any arithmetic so the narrower literal type can't overflow; `cluster_slot`/`activated_stake_lamports` are non-nullable `Int64` and `sum()`/`max()` over an empty set return 0, so the `int64` scan can't hit a NULL; and per-env key/connection isolation holds through `context.WithoutCancel`.

api/handlers/validators.go
```diff
@@ -72,47 +72,100 @@ var validatorFilterFields = map[string]FilterFieldConfig{
...
+	"client":     "software_client",
+}
+
+// validatorsPageCacheKey is the page-cache key for the default validators listing
+// (no filters, sorted by stake desc, first page). The worker refreshes it; the
+// handler serves it for the matching request shape.
+const validatorsPageCacheKey = "validators"
 
+func (a *API) GetValidators(w http.ResponseWriter, r *http.Request) {
 	pagination := ParsePagination(r, 100)
```
ben-dz: The page-cache gate at line 145 compares `p.Limit == DefaultLimit`, but this call hardcodes `100`. They agree today (`DefaultLimit = 100`, `api/handlers/pagination.go:11`), and nothing pins that: `TestIsDefaultValidatorsRequest` re-hardcodes `ParsePagination(r, 100)` itself (`api/handlers/validators_default_request_test.go:29`), so changing the handler's default to e.g. 50 would leave the test green while the cache silently stops hitting and the worker keeps refreshing an entry no request ever reads — a silent regression of exactly what this PR is for.

Fix: `ParsePagination(r, DefaultLimit)` here, and use `DefaultLimit` in the test's parse helper.

api/handlers/validators.go
```diff
@@ -72,47 +72,100 @@ var validatorFilterFields = map[string]FilterFieldConfig{
...
 	filterClause, filterArgs := filters.BuildFilterClause(validatorFilterFields)
 	whereFilter := ""
 	if filterClause != "" {
 		whereFilter = " AND " + filterClause
 	}
+	orderBy := sort.OrderByClause(validatorsQuerySortFields)
+
+	resp, err := a.fetchValidatorsPage(ctx, whereFilter, filterArgs, orderBy, pagination.Limit, pagination.Offset)
+	if err != nil {
+		logError("validators query failed", "error", err)
```
ben-dz, line 133: Extracting `fetchValidatorsPage` collapsed three distinct log messages ("validators query failed" / "validators row scan failed" / "validators rows iteration failed") into this one, since lines 283, 310, and 316 all `return nil, err` bare. A scan-type mismatch and a ClickHouse failure are now indistinguishable in prod logs, and the worker's refresh path surfaces the same errors.

Fix: wrap each return with stage context — `fmt.Errorf("validators row scan: %w", err)` and `fmt.Errorf("validators rows iteration: %w", err)`. `%w` keeps `dberror.IsTransient` classification working through `logError`.

api/handlers/validators.go
```diff
@@ -72,47 +72,100 @@ var validatorFilterFields = map[string]FilterFieldConfig{
...
+	"out":        "out_bps",
+	"skip":       "skip_rate",
+	"version":    "version",
+	"client":     "software_client",
+}
+
+// validatorsPageCacheKey is the page-cache key for the default validators listing
+// (no filters, sorted by stake desc, first page). The worker refreshes it; the
+// handler serves it for the matching request shape.
+const validatorsPageCacheKey = "validators"
```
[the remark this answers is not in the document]
nikw9944: Done in 6d8e6e7 — `validatorsPageCacheKey` is now exported as `ValidatorsPageCacheKey` and the worker entry references it instead of the raw string, matching the `MulticastHealthSummariesCacheKey` pattern.

api/handlers/validators.go
nikw9944: Done in 6d8e6e7 — the handler now calls `ParsePagination(r, DefaultLimit)` and the test's parse helper uses `DefaultLimit` too, so the parse default and the cache gate's comparison move together.

api/handlers/validators.go
nikw9944: Done in 6d8e6e7 — scan and iteration errors are wrapped with stage context (`validators row scan: %w`, `validators rows iteration: %w`); `%w` preserves `dberror.IsTransient` classification. A bare error now unambiguously means the query itself failed.

nikw9944:
Addressed all review feedback in 6d8e6e7:

- **Shared cache key** (Copilot): exported `ValidatorsPageCacheKey`; the worker entry now uses it instead of a duplicate string literal.
- **`DefaultLimit` pinning** (@ben-dz, medium): `GetValidators` parses with `DefaultLimit` instead of a hardcoded `100`, and the default-shape test does the same, so the cache gate can't silently drift from the handler default.
- **Error stage context** (@ben-dz, low): row-scan and rows-iteration errors in `fetchValidatorsPage` are wrapped (`%w`, keeping `dberror.IsTransient` intact), so the three failure modes are distinguishable in prod logs again.

Nothing deliberately skipped. Build, vet, lint, and worker tests pass locally; base is current with `main`.

ben-dz (approved):
Both findings from the previous round are fixed — verified in code at `6d8e6e7`, not just claimed.

- `GetValidators` now parses with `ParsePagination(r, DefaultLimit)` (`api/handlers/validators.go:105`) against the gate's `p.Limit == DefaultLimit` (`:149`), so the two sites share one constant and the silent-drift path is closed structurally rather than by test convention.
- Row-scan and rows-iteration failures are wrapped with stage context (`:314`, `:320`); the bare error at `:287` now unambiguously means the query itself failed. `%w` preserves classification — `dberror.IsTransient` unwraps via `errors.Is`.
- The duplicated page-cache key is gone: exported as `ValidatorsPageCacheKey` (`:100`) and referenced by the worker entry (`api/worker/workflow.go:252`), with no stale references to the old unexported name anywhere in `api/`.

The new commit introduces nothing new: it touches only the constant rename, two `fmt.Errorf` wraps, and the test helper — none of `scalar_cache.go`, `timeline.go`, or `multicast.go`, so the earlier verification of the SQL rewrites, ClickHouse literal typing, NULL-scan safety, and per-env cache isolation still stands.

One thing to glance at before merging: `go-test` was still running on this head. It was green on the previous head and the delta is a rename plus two error wraps, so nothing here should break it.

PR #737, opened 2026-07-28
api: cache edge scoreboard leader info and v1 validators-metadata

nikw9944:
## Summary of Changes
* Replace the edge scoreboard's per-run `validatorsapp_validators_current` + geoip leader-enrichment join with a per-env 5-minute TTL map (one unfiltered query rebuilds it; leader lookups become Go map reads). This is the query shape endpoint caching could not fix — most of its runs are the page-cache worker refreshing already-cached endpoints — so the fix sits at the call site, cutting it from ~9,000 runs/day to ~288/day/env.
* Add page-cache coverage (~60s) for the externally-polled, parameterless v1 `/solana/validators-metadata` endpoint.
* Enrichment/behavior preserved: a pubkey absent from the map yields empty name/ip/asn/geo, identical to the prior LEFT-JOIN no-match; the v1 cache read is gated on mainnet so non-mainnet requests still get their live per-env answer.
* Relates to #732

Second of two PRs for #732, stacked on #736 (base is `nikw9944/lake-732`; GitHub will retarget to `main` once #736 merges). Separated because it changes an API path's behavior under load, so its blast radius and rollback story differ from the frequency fixes in #736.

Operational note: the ClickHouse query metric label for this shape changes from `edge_scoreboard:q6b` to `edge_scoreboard:leader_info` — retarget any dashboard/alert keyed on the old label.

## Testing Verification
* Leader-info cache unit tests: map cached within TTL and dropped after expiry, empty map never pinned (cold/empty tables serve live), per-env isolation.
* Worker test asserts the new `validators_metadata` entry refreshes every other slow cycle (~60s).
* v1 metadata cache read falls back to the live per-env query on non-mainnet requests and on a corrupt cached payload (logged at WARN).


nikw9944:
Closing per operator guidance: this work ships as one PR per repo. The edge-scoreboard leader-info cache and v1 validators-metadata coverage will be proposed as a follow-up issue instead; the reviewed implementation is preserved on the `nikw9944/lake-732-b` branch (this PR's head) for the follow-up to reuse.

PR #742, opened 2026-07-30
indexer: track feed_pks and feed entities in entity_changes_v

nikw9944:
## Summary of Changes

* `entity_changes_v` now enumerates `feed_pks` in the users section's `changed_fields`, so a user gaining or losing an EdgeSeat feed seat renders as a real timeline event instead of a contentless "updated (0 fields)". `feed_pks` was already part of the users `attrs_hash` (#703), so the SCD delta fired but named no field — the same gap the `bgp_status` follow-up closed. The user details fetcher now also diffs `bgp_status` and `feed_pks`, so neither vanishes from an event when it changes together with an already-diffed field.
* Adds a sixth `feed` section over `dim_dz_feeds_history`, so Feed accounts appear in the timeline as first-class entities. Feed events get a details fetcher like the other five types: old→new field chips, an entity card with a metro link, and `metro:` search matching. `changed_fields` covers every column in the feed `attrs_hash` (`code`, `name`, `owner_pubkey`, `metro`, `groups`), so no feed update can render as "0 fields".
* Adds `feed` to the timeline UI's entity-type constants and filter chips. This is load-bearing, not cosmetic: the UI sends an explicit `entity_type=` allowlist on every request and the handler filters on it, so the new section's rows would be discarded server-side otherwise.
* Follow-up to malbeclabs/lake#703, requested in review there. Fixes malbeclabs/infra#2092.

Both view changes are retroactive — the view is computed at query time, so past feed-seat transitions gain `['feed_pks']` and the full Feed account history since #703 started ingesting appears in the timeline at once. Expect the timeline to change on deploy.

The migration restates the whole view because ClickHouse has no ALTER VIEW delta. Its `Down` body is byte-identical to the current deployed definition (the `Up` of `20260709000001`), and its `Up` differs from that by exactly the three users-section lines plus the appended feed section — verified by diff rather than by eye.

## Testing Verification

* `TestTimeline_FeedRoundTrip` — full handler path: `GET /api/timeline?entity_type=feed` returns the feed update with its title, field diff (`groups` old→new), and the entity's join-resolved `metro_code`.
* `TestEntityChanges_PayloadColumnsTracked` — pins the `PayloadColumns()` ↔ `changed_fields` contract for all six entity types: per schema, every payload column is changed in isolation and must produce a non-empty `changed_fields`, with an explicit skip list pinning the known gaps as empty. The next payload column added to any schema fails this test instead of shipping another "updated (0 fields)" regression (how #703 shipped this bug).
* `TestEntityChanges_FeedPksTransition` — two user snapshots differing only in `feed_pks` produce exactly one `updated` event with `changed_fields = ['feed_pks']`.
* `TestEntityChanges_FeedEntity` — exercises all three feed change types (`updated`, `deleted`, `created` after the `min_ts` baseline), asserts `metro_code` resolves through the join and that an unresolvable `metro_pk` renders as `''`.
* Copy fidelity checked mechanically: `Down` diffs empty against the deployed definition; the `Up` diff contains only the intended additions.
* The ClickHouse-backed tests were not run locally (no Docker in the authoring sandbox) — CI runs them against the full goose migration set.

## Follow-up

The users section's `changed_fields` still omits `owner_pubkey`, `publishers`, `subscribers`, and `tenant_pk`, which are also in the users `attrs_hash` — e.g. a multicast group join/leave remains a "0 fields" event. Pre-existing and a separate call about which user fields deserve timeline labels, so it is not folded in here; the skip list in `TestEntityChanges_PayloadColumnsTracked` names each gap, and closing one in the view now forces removing it there. Worth a follow-up issue so this is not patched one field at a time a third time.


indexer/db/clickhouse/migrations/20260730000001_entity_changes_feeds.sql
```diff
@@ -0,0 +1,681 @@
...
+            if(h.row_num > 1 AND h.prev_owner_pubkey != h.owner_pubkey, 'owner_pubkey', ''),
+            if(h.row_num > 1 AND h.prev_metro_pk != h.metro_pk, 'metro', ''),
+            if(h.row_num > 1 AND h.prev_groups != h.groups, 'groups', '')
+        ]) AS changed_fields,
+        '' AS new_status,
+        '' AS contributor_code,
+        COALESCE(m.code, '') AS metro_code
+    FROM all_history h
+    CROSS JOIN min_ts
+    LEFT JOIN dz_metros_current m ON h.metro_pk = m.pk
```
ben-dz, line 375: This join's output is never read. `batchFetchEntityDetails` (`api/handlers/timeline.go:1766-1779`) has cases for `device`, `link`, `metro`, `contributor`, `user` — not `feed` — so feed rows fall through to the minimal fallback at `timeline.go:1968-1971`. `entityChangeRow.MetroCode` is written by the scan at `timeline.go:1048` and read nowhere; the metro/contributor search filters at `timeline.go:413-447` read `entity.MetroCode`/`details.MetroCode`, which come from the details fetchers. So `metro:ams` will not match a feed event, and the CHANGELOG's "with `metro_code` resolved through `metro_pk`" is not observable in the product even though `TestEntityChanges_FeedEntity` asserts it.

Separately, feed cards carry no old→new values: `ChangeSummary` (`web/src/components/timeline-event-card.tsx:115`) returns `null` on an empty `changes` array, so a feed update renders as a bare title where the other five DZ types render the field diff. `EntityDetailsView` returns `null` for an unknown type (`timeline-event-card.tsx:236`), so nothing crashes — it just degrades.

Pick one direction: add a `fetchFeedChangeDetails` plus a `grouped["feed"]` branch at `timeline.go:1778` (`fetchMetroChangeDetails` is the closest template — feeds have no status and no contributor), or drop this join and the `metro_code` claim from the CHANGELOG. Shipping the join and a test for a value nothing reads is the worst of the three.

indexer/db/clickhouse/migrations/20260730000001_entity_changes_feeds.sql
```diff
@@ -0,0 +1,681 @@
...
+        END AS change_type,
+        arrayFilter(x -> x != '', [
+            if(h.row_num > 1 AND h.prev_status != h.status, 'status', ''),
+            if(h.row_num > 1 AND h.prev_kind != h.kind, 'kind', ''),
+            if(h.row_num > 1 AND h.prev_client_ip != h.client_ip, 'client_ip', ''),
+            if(h.row_num > 1 AND h.prev_dz_ip != h.dz_ip, 'dz_ip', ''),
+            if(h.row_num > 1 AND h.prev_device_pk != h.device_pk, 'device', ''),
+            if(h.row_num > 1 AND toString(h.prev_tunnel_id) != toString(h.tunnel_id), 'tunnel_id', ''),
+            if(h.row_num > 1 AND h.prev_bgp_status != h.bgp_status, 'bgp_status', ''),
+            if(h.row_num > 1 AND h.prev_feed_pks != h.feed_pks, 'feed_pks', '')
```
ben-dz, line 310: This line is correct, but it doesn't reach the UI in the co-occurring case. `buildEntityChangeTitle` overrides the view's `changed_fields` with the details fetcher's list whenever the latter is non-empty (`api/handlers/timeline.go:1832-1839`), and `fetchUserChangeDetails` diffs only `status`, `kind`, `client_ip`, `dz_ip`, `device`, `tunnel_id` (`timeline.go:1696-1712`) — not `feed_pks`.

Reproduce: one user snapshot changing both `device_pk` and `feed_pks`. The view emits `changed_fields = ['device','feed_pks']`, but `details.Changes = [device]` wins, `len(changes) == 1`, and the title becomes `User <code> moved to <DEV>` — the feed-seat change is gone from both the title and the card body. The isolated case works (`details.Changes` is empty, so the view's list survives and the title reads `User <code> feed_pks changed`), which is why `TestEntityChanges_FeedPksTransition` passes.

Fix: add `feed_pks` to the `fetchUserChangeDetails` SELECT/`lag()` list and to its diff block. `bgp_status` has the identical gap and the identical one-line fix; folding it in here is cheaper than a third pass over this function.

indexer/db/clickhouse/migrations/20260730000001_entity_changes_feeds.sql
```diff
@@ -0,0 +1,681 @@
...
+        h.pk AS entity_pk,
+        h.owner_pubkey AS entity_code,
+        h.snapshot_ts AS snapshot_ts,
+        CASE
+            WHEN h.row_num = 1 THEN 'created'
+            WHEN h.is_deleted = 1 AND h.prev_is_deleted = 0 THEN 'deleted'
+            ELSE 'updated'
+        END AS change_type,
+        arrayFilter(x -> x != '', [
+            if(h.row_num > 1 AND h.prev_status != h.status, 'status', ''),
```
ben-dz, line 303: This block enumerates 8 fields; `UserSchema.PayloadColumns()` (`indexer/pkg/dz/serviceability/schema.go:153-167`) has 12. The PR body flags `publishers`, `subscribers`, and `tenant_pk` as deferred, but not `owner_pubkey`, which is also in the users `attrs_hash` and also renders as "updated (0 fields)". Worth naming in the follow-up issue so the deferral list is complete.

ben-dz (commented):
The view SQL itself is correct — feed `changed_fields` covers all five `FeedSchema.PayloadColumns()` entries, no Nullable comparison drops a field, no join fan-out, UNION column order matches across all six branches, the `Down` body is byte-identical to the deployed `20260709000001` definition, and the per-test-database isolation the `min_ts` comment relies on does hold. Three things to resolve before this lands:

1. The new `feed` section is wired into the view and the UI allowlist but not into `batchFetchEntityDetails`, so the `metro_code` the PR adds and tests is read by nothing, and feed cards render without field values.
2. `feed_pks` still vanishes from the timeline when it changes alongside a field the user details fetcher diffs — the same invisibility the PR exists to fix, in the co-occurring case.
3. Neither new test covers the `entity_type=feed` handler round-trip the PR calls load-bearing, nor the `PayloadColumns()` ↔ `changed_fields` invariant whose violation caused this bug in the first place.

- **Test coverage sits below the layer where the bug lives.** Both new tests query `entity_changes_v` directly via `api.DB.Query`, leaving two gaps.

First, the `entity_type=feed` round-trip. The PR body calls the `timeline-constants.ts` change "load-bearing" because the handler filters on the client-sent allowlist (`api/handlers/timeline.go:788-799`), but nothing asserts a feed event survives `GET /api/timeline?entity_type=feed`. `timeline_test.go` already has the `httptest.NewRequest` pattern, so this is a few lines.

Second, and more valuable: nothing pins the invariant. `feed_pks` was added to `UserSchema.PayloadColumns()` in #703 and no test failed — that is the answer to "how wasn't this caught by a test". The two new tests pin two specific columns, so the next payload column added to any schema reintroduces "updated (0 fields)" just as silently. A table-driven test that, per schema, inserts two snapshots differing in exactly one payload column and asserts `changed_fields` is non-empty — with an explicit skip list naming the deliberate omissions — turns the next occurrence into a build failure, instead of a third round of this same fix.

indexer/db/clickhouse/migrations/20260730000001_entity_changes_feeds.sql
nikw9944: Went with the first direction: 1545abd adds `fetchFeedChangeDetails` (modeled on the metro fetcher, plus the metro join) and the `grouped["feed"]` dispatch. Feed cards now carry old→new field diffs and the entity with its resolved `metro_code`, `metro:ams` search matches feed events via the new `FeedEntity` case in the metro search arm, and the web side gets the `FeedEntity` type plus an `EntityDetailsView` feed case with a metro link. `TestTimeline_FeedRoundTrip` asserts the `metro_code` through the full handler path, so the join and the CHANGELOG claim are now observable.

indexer/db/clickhouse/migrations/20260730000001_entity_changes_feeds.sql
nikw9944: Fixed in 1545abd: `fetchUserChangeDetails` now selects, lags, and diffs both `feed_pks` and `bgp_status`, so the view's list is no longer overridden by a details list missing them in the co-occurring case. Folded `bgp_status` in as suggested — same gap, same one-line shape.

indexer/db/clickhouse/migrations/20260730000001_entity_changes_feeds.sql
nikw9944: Done — the PR body's follow-up list now names all four (`owner_pubkey`, `publishers`, `subscribers`, `tenant_pk`), and the new `TestEntityChanges_PayloadColumnsTracked` (1545abd) pins each of them in an explicit skip list, so the deferral list can no longer drift from reality: closing a gap in the view forces removing it from the skip list.

nikw9944:
All three threads plus the review-body test asks are addressed in 1545abd:

- **Feed details fetcher**: added `fetchFeedChangeDetails` + the missing `grouped["feed"]` dispatch, `FeedEntity` on both the Go and TS sides, an `EntityDetailsView` feed card with a metro link, and `metro:` search matching for feed events. The view's metro join is now read, and its output is asserted through the full handler path.
- **Co-occurring `feed_pks` invisibility**: `fetchUserChangeDetails` now diffs `feed_pks` and `bgp_status`, so neither is dropped when the details list overrides the view's `changed_fields`.
- **Test layer**: `TestTimeline_FeedRoundTrip` covers `GET /api/timeline?entity_type=feed` end to end (the allowlist path the PR body calls load-bearing). `TestEntityChanges_PayloadColumnsTracked` pins the `PayloadColumns()` ↔ `changed_fields` contract for all six entity types — every payload column changed in isolation must produce non-empty `changed_fields`, with an explicit skip list pinning the known gaps (including `owner_pubkey`) as empty. A new payload column on any schema now fails this test instead of shipping another "updated (0 fields)" regression.
- **Deferral list**: the PR body's follow-up section now names `owner_pubkey` alongside `publishers`/`subscribers`/`tenant_pk`.

Not changed: no `feed:` search prefix and no `/dz/feeds/:pk` detail page — there is no feeds page to link to yet, and none of the threads asked for one.

ben-dz (approved):
All three previously-raised issues plus the deferral-list nit are addressed in `1545abd`, verified in code rather than taken from the replies.

- The feed details fetcher (`fetchFeedChangeDetails`, `api/handlers/timeline.go:1766-1886`) plus the `grouped["feed"]` dispatch at `:1931` are in, modeled on the metro fetcher with the metro join added the way the device and user fetchers do it. `FeedEntity` exists on both the Go and TS sides, `EntityDetailsView` has a feed case, and the metro search arm matches it. `TestTimeline_FeedRoundTrip` asserts the resolved `metro_code` and the `groups` old→new diff through `GetTimeline`, so the join's output is now observed.
- `fetchUserChangeDetails` diffs `feed_pks` and `bgp_status`, so neither is swallowed when the details list overrides the view's `changed_fields`. One consequence worth knowing rather than fixing: a device move with a concurrent BGP flap now titles as `updated (2 fields)` instead of `moved to <DEV>`, with both field chips present — the CHANGELOG describes the change.
- `TestEntityChanges_PayloadColumnsTracked` is a tighter contract than asked for: it reflects `PayloadColumns()` for all six schemas and pins skip-listed columns as *empty*, so closing a gap in the view forces updating the list. Confirmed it can't flake — `SetupClickHouseWithMigrationsForTest` (`api/testing/clickhouse.go:257-270`) gives each test its own database, so the `2025-06-01` baselines can't collide with the other tests' timestamps under `t.Parallel()`, and `unicast_drained:Boolean` matches the `valueFor` `Boolean` arm so no arm inserts a string into a `UInt8`.

The remaining user columns (`owner_pubkey`, `tenant_pk`, `publishers`, `subscribers`) and the pre-existing device/link/metro gaps are now enumerated in the skip lists and named in the PR body's follow-up — documented deferral, not a blocker.

PR #744, opened 2026-07-30
indexer,api: parallelize msdp/mroute S3 scans and multicast-delivery queries

ben-dz:
## Summary of Changes

- **The msdp/mroute S3 walks now fan out.** They read devices serially — ~1,200 sequential round trips per MSDP cycle, ~400 for mroute — inside one 5m activity timeout. Prod mainnet-beta `SyncMSDP` averaged 155s (max 190s, 63% of the deadline) and staging had started aborting whole cycles on single transient S3 blips. Reads now run under `errgroup` + `SetLimit(10)`, collected by device index so ordering still follows the S3 listing.
- **Every S3 call in the walk retries**, the device listing included, bounded at 3 attempts and gated on `dberror.IsTransient` so a cancelled activity aborts rather than retrying into its deadline. Nothing above covers this: `Activities.refresh` returns nil to Temporal, so the workflow's `MaximumAttempts` never applies to these activities.
- **The all-or-nothing abort is unchanged** — a partial device set would tombstone every device whose read failed. Devices with an S3 prefix but no snapshot in the lookback window are now named in a WARN rather than dropped silently.
- **`view_type=msdp|mroute` on the refresh counter and duration histogram**, so the predicted effect (MSDP ~155s → ~20s, mroute ~57s → ~8s) is checkable after deploy rather than assumed.
- **`GetLinkMulticastDelivery` becomes a two-stage fan-out.** Its six sequential queries over the same `enriched_ip_mroute_oifs` and health view chains overran the handler's 30s deadline in CI. Error precedence, source-unavailable downgrades, and response shape are unchanged.
- **The test template-DB clone skips ClickHouse internals.** `dz_device_interface_ips` is a REFRESHABLE materialized view, so the long-lived template DB rebuilds it into a transient `.tmp.inner_id.<uuid>` table every minute; a clone whose `system.tables` read landed in that window died with `code: 62, Empty query`.

## Diff Breakdown
| Category     | Files | Lines (+/-)   | Net   |
|--------------|-------|---------------|-------|
| Tests        |     4 | +748 / -87    |  +661 |
| Core logic   |     3 | +280 / -57    |  +223 |
| Scaffolding  |     3 | +47 / -5      |   +42 |
| **Total**    |    10 | +1075 / -149  |  +926 |

Three concurrency rewrites carrying most of the risk, behind roughly 2.5x their weight in tests.

<details>

- [`api/handlers/multicast_delivery_entity.go`](https://github.com/malbeclabs/lake/pull/744/files#diff-53c4554235bcf03c0235d7eecb420f6427b15de3a360f380cda5eb651b11a29b) — link handler two-stage fan-out; shared spawn/semaphore extracted to `multicastDeliveryFanout`
- [`indexer/pkg/dz/mroute/source_s3.go`](https://github.com/malbeclabs/lake/pull/744/files#diff-4630e62195f6688098893fea2c22ceec727eec37eeb974a69256ffd919dfa11e) — per-device fan-out, retry on listing and per-device reads, stale-device WARN
- [`indexer/pkg/dz/msdp/source_s3.go`](https://github.com/malbeclabs/lake/pull/744/files#diff-36ca940360154671dfe6a001a013e36c56fea411c520e308fefd86de713570ee) — same, per snapshot kind
- [`indexer/pkg/dzingest/activities.go`](https://github.com/malbeclabs/lake/pull/744/files#diff-a2351b69fbbb5752bb5e678ef261f2d1da4e9294f7230534ccb29f76b388147a) — `observeSync` wrapper adding refresh metrics and a duration log line
- [`api/testing/clickhouse.go`](https://github.com/malbeclabs/lake/pull/744/files#diff-3f95e724029ee98315a41802e66b76489d40101859345ed48b4c0094312ab3c6) — `skipTemplateTable` excludes dot-prefixed internals from the clone

</details>

## Testing Verification

- `TestS3Source_FetchLatest_FanOut` in both packages, MinIO-backed with a fault-injecting `http.RoundTripper`: a first-request failure on a device *or on the device listing* is retried and the walk still returns the full device set; a persistent failure aborts after exactly `MaxAttempts` requests; a failure that also cancels the context returns `context.Canceled` after one attempt, with no not-yet-started device reaching S3; 25 devices return in listing order with peak in-flight above 1 and at most 10.
- `TestSkipTemplateTable` pins the clone filter, including the exact `.tmp.inner_id.<uuid>` name from the CI failure, and runs without a container.
- Docker is unavailable in my environment, so CI runs the container-backed tests rather than me. The link handler went from a 500 after 31.9s to passing in 15.9s.

## Follow-up

Collapsing the three per-device lookback LISTs into one LIST at the `device=` prefix would cut round trips a further ~3x. Left out: it changes scan semantics, and concurrency already restores ample headroom.

### DIFF
--- a/indexer/pkg/dz/mroute/source_s3.go
+++ b/indexer/pkg/dz/mroute/source_s3.go
@@ -11,9 +11,26 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/malbeclabs/lake/indexer/pkg/dz/statecollect"
+	"github.com/malbeclabs/lake/utils/pkg/dberror"
 )
+
+// maxConcurrentFetches bounds in-flight per-device S3 reads, matching the limit
+// used by the serviceability and shreds views.
+const maxConcurrentFetches = 10
+
+// deviceFetchRetry bounds per-device retries for transient S3 failures. The fetch
+// is all-or-nothing (see Store.Sync), so without a retry one blip among the
+// hundreds of calls a cycle makes aborts the whole thing. dberror.IsTransient
+// treats context errors as non-transient, so a cancelled activity aborts
+// immediately rather than retrying into its deadline.
+var deviceFetchRetry = dberror.RetryConfig{
+	MaxAttempts: 3,
+	BaseBackoff: 100 * time.Millisecond,
+	MaxBackoff:  time.Second,
+}
 
 // SnapshotKind is the state-collect kind name for mroute dumps. It matches the
 // key used by the telemetry state-ingest server when it presigns the
@@ -100,14 +117,49 @@ func (s *S3Source) FetchLatest(ctx context.Context) ([]*Dump, error) {
 	}
 
 	now := time.Now().UTC()
-	var dumps []*Dump
-	for _, pubkey := range devices {
+
+	// Per-device reads are independent. Collect by device index so the result
+	// order tracks the device listing regardless of completion order.
+	perDevice := make([]*Dump, len(devices))
+	g, gctx := errgroup.WithContext(ctx)
+	g.SetLimit(maxConcurrentFetches)
+	for i, pubkey := range devices {
+		g.Go(func() error {
+			dump, err := s.fetchDevice(gctx, pubkey, now)
+			if err != nil {
+				return err
+			}
+			perDevice[i] = dump
+			return nil
+		})
+	}
+	// One failed device aborts the whole fetch: Store.Sync writes with
+	// MissingMeansDeleted=true, so returning a partial device set would
+	// tombstone the missing devices' prior state.
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
+
+	dumps := make([]*Dump, 0, len(perDevice))
+	for _, d := range perDevice {
+		if d != nil {
+			dumps = append(dumps, d)
+		}
+	}
+	return dumps, nil
+}
+
+// fetchDevice reads the latest snapshot for one device, retrying transient S3
+// failures. A nil Dump with a nil error means the device has no snapshot in the
+// lookback window.
+func (s *S3Source) fetchDevice(ctx context.Context, pubkey string, now time.Time) (*Dump, error) {
+	return dberror.Retry(ctx, deviceFetchRetry, func() (*Dump, error) {
 		key, keyTS, err := s.latestKeyForDevice(ctx, pubkey, now)
 		if err != nil {
 			return nil, err
 		}
 		if key == "" {
-			continue
+			return nil, nil
 		}
 		body, err := s.getObject(ctx, key)
 		if err != nil {
@@ -136,15 +188,14 @@ func (s *S3Source) FetchLatest(ctx context.Context) ([]*Dump, error) {
 			}
 		}
 
-		dumps = append(dumps, &Dump{
+		return &Dump{
 			FetchedAt:    time.Now().UTC(),
 			SnapshotTS:   snapTS,
 			DevicePubkey: devicePK,
 			RawJSON:      env.Data,
 			FileName:     key,
-		})
-	}
-	return dumps, nil
+		}, nil
+	})
 }
 
 // listDevicePubkeys uses delimiter listing to enumerate the device=<pubkey>
--- a/indexer/pkg/dz/mroute/source_s3.go
+++ b/indexer/pkg/dz/mroute/source_s3.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log/slog"
 	"strings"
 	"time"
 
@@ -48,6 +49,9 @@ type S3SourceConfig struct {
 	// finding the latest per device. Three covers clock skew and the
 	// midnight-UTC boundary; tests can shorten this.
 	LookbackDays int
+
+	// Logger is optional; it defaults to slog.Default().
+	Logger *slog.Logger
 }
 
 // S3Source implements Source against an S3 bucket populated by the telemetry
@@ -56,6 +60,7 @@ type S3SourceConfig struct {
 //	<KeyPrefix>/snapshots/<SnapshotKind>/device=<pubkey>/date=YYYY-MM-DD/hour=HH/<ts>.json
 type S3Source struct {
 	client       *s3.Client
+	log          *slog.Logger
 	bucket       string
 	keyPrefix    string // already stripped of trailing slash; empty when no prefix
 	lookbackDays int
@@ -75,6 +80,10 @@ func NewS3Source(ctx context.Context, cfg S3SourceConfig) (*S3Source, error) {
 		cfg.LookbackDays = 3
 	}
 
+	if cfg.Logger == nil {
+		cfg.Logger = slog.Default()
+	}
+
 	awsCfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(cfg.Region))
 	if err != nil {
 		return nil, fmt.Errorf("mroute: load AWS config: %w", err)
@@ -91,6 +100,7 @@ func NewS3Source(ctx context.Context, cfg S3SourceConfig) (*S3Source, error) {
 
 	return &S3Source{
 		client:       s3.NewFromConfig(awsCfg, clientOpts...),
+		log:          cfg.Logger,
 		bucket:       cfg.Bucket,
 		keyPrefix:    strings.TrimRight(cfg.KeyPrefix, "/"),
 		lookbackDays: cfg.LookbackDays,
@@ -125,6 +135,13 @@ func (s *S3Source) FetchLatest(ctx context.Context) ([]*Dump, error) {
 	g.SetLimit(maxConcurrentFetches)
 	for i, pubkey := range devices {
 		g.Go(func() error {
+			// errgroup still invokes every function queued behind SetLimit
+			// after the group context is cancelled. The AWS SDK rejects the
+			// call on a cancelled context anyway; returning here makes the
+			// skip explicit instead of relying on that.
+			if err := gctx.Err(); err != nil {
+				return err
+			}
 			dump, err := s.fetchDevice(gctx, pubkey, now)
 			if err != nil {
 				return err
@@ -135,16 +152,28 @@ func (s *S3Source) FetchLatest(ctx context.Context) ([]*Dump, error) {
 	}
 	// One failed device aborts the whole fetch: Store.Sync writes with
 	// MissingMeansDeleted=true, so returning a partial device set would
-	// tombstone the missing devices' prior state.
+	// tombstone the devices whose reads failed.
 	if err := g.Wait(); err != nil {
 		return nil, err
 	}
 
 	dumps := make([]*Dump, 0, len(perDevice))
-	for _, d := range perDevice {
-		if d != nil {
-			dumps = append(dumps, d)
+	var stale []string
+	for i, d := range perDevice {
+		if d == nil {
+			stale = append(stale, devices[i])
+			continue
 		}
+		dumps = append(dumps, d)
+	}
+	// A device whose prefix exists but holds no snapshot in the lookback window
+	// is dropped here, and MissingMeansDeleted=true then tombstones its rows.
+	// That is the intended outcome for a device that stopped uploading — its
+	// state is no longer current — but it is the one path where rows disappear
+	// without an error, so name the devices rather than let them vanish.
+	if len(stale) > 0 {
+		s.log.Warn("mroute: devices have no snapshot in the lookback window, their prior state will be tombstoned",
+			"lookback_days", s.lookbackDays, "count", len(stale), "devices", strings.Join(stale, ","))
 	}
 	return dumps, nil
 }
--- a/indexer/pkg/dz/mroute/source_s3.go
+++ b/indexer/pkg/dz/mroute/source_s3.go
@@ -22,11 +22,13 @@ import (
 // used by the serviceability and shreds views.
 const maxConcurrentFetches = 10
 
-// deviceFetchRetry bounds per-device retries for transient S3 failures. The fetch
-// is all-or-nothing (see Store.Sync), so without a retry one blip among the
-// hundreds of calls a cycle makes aborts the whole thing. dberror.IsTransient
-// treats context errors as non-transient, so a cancelled activity aborts
-// immediately rather than retrying into its deadline.
+// deviceFetchRetry bounds retries for transient S3 failures, covering both the
+// device listing and each per-device read. The fetch is all-or-nothing (see
+// Store.Sync), so without a retry one blip among the hundreds of calls a cycle
+// makes aborts the whole thing — and nothing above retries either, since
+// Activities.refresh returns nil to Temporal. dberror.IsTransient treats
+// context errors as non-transient, so a cancelled activity aborts immediately
+// rather than retrying into its deadline.
 var deviceFetchRetry = dberror.RetryConfig{
 	MaxAttempts: 3,
 	BaseBackoff: 100 * time.Millisecond,
@@ -121,7 +123,9 @@ func (s *S3Source) snapshotsPrefix() string {
 // window. Each S3 object is a statecollect.Envelope; this method unwraps
 // it so callers receive only the inner Arista JSON in Dump.RawJSON.
 func (s *S3Source) FetchLatest(ctx context.Context) ([]*Dump, error) {
-	devices, err := s.listDevicePubkeys(ctx)
+	devices, err := dberror.Retry(ctx, deviceFetchRetry, func() ([]string, error) {
+		return s.listDevicePubkeys(ctx)
+	})
 	if err != nil {
 		return nil, err
 	}
--- a/indexer/pkg/dz/msdp/source_s3.go
+++ b/indexer/pkg/dz/msdp/source_s3.go
@@ -22,11 +22,13 @@ import (
 // used by the serviceability and shreds views.
 const maxConcurrentFetches = 10
 
-// deviceFetchRetry bounds per-device retries for transient S3 failures. The fetch
-// is all-or-nothing (see Store.Sync), so without a retry one blip among the
-// hundreds of calls a cycle makes aborts the whole thing. dberror.IsTransient
-// treats context errors as non-transient, so a cancelled activity aborts
-// immediately rather than retrying into its deadline.
+// deviceFetchRetry bounds retries for transient S3 failures, covering both the
+// per-kind device listing and each per-device read. The fetch is all-or-nothing
+// (see Store.Sync), so without a retry one blip among the hundreds of calls a
+// cycle makes aborts the whole thing — and nothing above retries either, since
+// Activities.refresh returns nil to Temporal. dberror.IsTransient treats
+// context errors as non-transient, so a cancelled activity aborts immediately
+// rather than retrying into its deadline.
 var deviceFetchRetry = dberror.RetryConfig{
 	MaxAttempts: 3,
 	BaseBackoff: 100 * time.Millisecond,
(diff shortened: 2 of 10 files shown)


nikw9944:
## Review note: the headline claim has no way to be verified after merge

The change looks right and fits the existing error-handling work well (see below). One gap worth closing before merge:

**"MSDP ~155s to roughly 20s, mroute ~57s to roughly 8s" is currently unfalsifiable in prod.** Neither activity is observable:

- They are **not** instrumented with the view-refresh histogram. Live `view_type` label values on `doublezero_data_indexer_view_refresh_total{component="lake-indexer"}` are: `geolocation, serviceability, shreds, shreds_escrow_events, telemetry, telemetry-usage, validatorsapp, geoip, serviceability_permission_events, solana` — no `msdp`, no `mroute`.
- I could not find a duration log line for either sync in prod logs (searched `lake-indexer`/`lake-prod` for msdp duration lines over 3h: none).

So after this merges there is no metric and no log line to confirm the 8× improvement, and no signal if it regresses later. That is the same trap #711 fell into: a plausible post-deploy prediction with nothing attached to check it, which then silently became "assumed done" (see #740, where the convergence claim went unverified for days and the pipeline was actually frozen).

Cheap options, either is fine:
1. A duration log line per sync (`INF msdp: sync completed … duration=…`), matching how the views log; or
2. Bring `SyncMSDP`/`SyncIPMroute` into the refresh metrics so `view_type=msdp|mroute` exists and the p95 is dashboardable next to the others.

(2) is more useful long-term — it also makes the 300s deadline's headroom visible, which is the thing that actually failed here.

## How this fits the recent error-handling work (for reviewers' context)

- **#733** (merged, in prod today) classified the transient S3 shapes so blips log WARN instead of ERROR — symptom management. **This PR removes the cause**: a blip no longer aborts the whole per-device walk, and the walk stops consuming 63% of its deadline. The two compose; with retries in place the WARN volume #733 absorbs should also drop.
- No file overlap with #733 (that touched `utils/pkg/dberror/*` only), so no conflict.
- Reusing `dberror.IsTransient` — and relying on context errors being classified non-transient so a cancelled activity aborts rather than retrying — matches the conventions from #697/#681. Good.

## Pattern observation

This is the third instance of the same failure class in this subsystem: **sequential per-entity I/O inside a fixed activity budget**.

1. telemetry-usage — Flux queries consume 5–6.5 min of a 15m budget → catch-up livelock (#740)
2. MSDP/mroute — ~1,200 sequential S3 round trips consume 155s of a 300s budget → abort on any blip (this PR)
3. page-cache — concurrency cap and interval tunables (#677, #688)

Each was found only after it failed in an environment under load. Worth considering a general guard (e.g. asserting worst-case per-cycle cost against the activity deadline, as #740's plan does for telemetry-usage) rather than discovering the fourth one the same way.

nikw9944:
## Full review

Read the diff against `main`, plus `dberror.Retry`, the dzingest workflow scheduling, and the prod pod spec. **The change is sound and I found no blockers.** One cheap correctness nit, three things worth naming, and an answer to the CPU question below.

### CPU: no, this should not drive meaningful CPU spikes — but it does stack with sibling fan-outs

The work being parallelized is **I/O-bound, not CPU-bound**: per device it's 3 lookback LISTs + 1 GET, where nearly all wall time is spent waiting on S3. Ten goroutines blocked in network reads cost almost no CPU. The only real CPU is the per-snapshot JSON decode, and total decode work is unchanged — it's the same bytes, just compressed from ~155s into ~20s. So expect a modest, brief rise in indexer CPU, not a spike proportional to the concurrency.

Three specifics that make this safe here:

- **Bounded twice.** `SetLimit(10)` caps in-flight reads, and the prod indexer pod is `cpu: 8` with requests == limits (guaranteed QoS). On **Go 1.25.6** the runtime derives `GOMAXPROCS` from the cgroup CPU limit, so the scheduler won't over-provision Ps against the 8-core quota (that container-aware default landed in 1.25 — worth confirming it isn't disabled via `GOMAXPROCS` in the env, which I don't see set).
- **ClickHouse is untouched.** The fan-out is S3-only; the CH write is still one batch insert after `g.Wait()`, same size, same shape. None of this lands on the shared cluster we've been optimizing.
- **Memory is ~unchanged.** `perDevice` holds every dump, but so did the serial `dumps` accumulator; the delta is ~10 concurrent decode buffers.

**The thing actually worth watching** is aggregate concurrency, not CPU: `workflow.go:104-109` launches `RefreshShreds`, `RefreshShredEscrowEvents`, `SyncISIS`, `SyncIPMroute`, `SyncMSDP`, and `SyncGraph` **concurrently** (`ExecuteActivity` returns futures without blocking). So MSDP's 10 and mroute's 10 now overlap each other and the serviceability/shreds fan-outs — potentially 40+ concurrent operations in one pod where previously the MSDP/mroute portions were serial. That's a pod-level resource question (sockets, TLS handshakes, GC pressure), not a ClickHouse one, and it's the reason I'd want the duration metric from my earlier comment: without it you can't see whether these now contend.

Second-order effect worth a sentence in the PR: finishing the fetch ~135s earlier means MSDP's and mroute's ClickHouse inserts now land ~135s earlier and closer together. Small, but it's a real change to write timing.

### Finding 1 (cheap fix): queued goroutines still burn one attempt each after an abort

`errgroup` does **not** skip functions queued behind `SetLimit` once the group context is cancelled — it runs all of them. And `dberror.Retry` doesn't check `ctx` before attempt 1 (the `ctx.Done()` select only guards attempts 2+, `dberror.go:258-266`). So after a persistent failure aborts the walk, every remaining device still issues one doomed S3 call: up to ~90 wasted requests on a 100-device cycle.

Not harmful, but it undercuts the "a cancelled activity aborts immediately" claim. One line in the closure fixes it:

```go
g.Go(func() error {
    if err := gctx.Err(); err != nil {
        return err
    }
    ...
})
```

### Finding 2: retry amplifies load 3× during exactly the incident it protects against

Per device, worst case is now 3 attempts × (3 LISTs + 1 GET) = 12 requests, so a cycle can go from ~1,200 to ~3,600 requests. During a *broad* S3 slowdown — many devices failing transiently rather than one — this triples offered load at the worst moment. It's bounded (3 attempts, 100ms–1s backoff, abort on persistent failure), so I don't think it blocks, but it deserves a sentence in the PR so the trade-off is explicit rather than discovered during an S3 event.

### Finding 3: the retry redoes the LISTs to recover a failed GET

`fetchDevice` wraps LIST(s) + GET + decode in one retry, so a transient GET failure re-runs up to 3 lookback LISTs. Simpler than splitting the retry, and cheap — just noting it's deliberate-looking rather than accidental.

### Finding 4: the tombstone rationale has a pre-existing hole

The comment says a partial device set "would tombstone the missing devices' prior state," which is the reason for all-or-nothing. But a device whose snapshot is merely **absent** from the lookback window returns `(nil, nil)`, gets skipped by the compaction loop, and is omitted from the result — so `MissingMeansDeleted=true` tombstones it anyway. Same outcome, different path, no protection.

That's pre-existing behavior (the serial version `continue`d identically), so not this PR's bug — but the new comment invites the question, and it's worth either acknowledging in the comment or filing separately. If a device stops uploading for >3 days, its state silently disappears rather than going stale.

### Verified correct (so you know these were checked)

- **Loop-variable capture** — safe under `go 1.25.6` per-iteration semantics; `i`/`pubkey` are correctly per-goroutine.
- **`IsTransient` gating** — confirmed at `dberror.go:276`: non-transient errors return without retry, so decode/permission failures don't burn attempts.
- **Cancellation path** — the `ctx.Done()` select at the top of attempts 2+ means a cancel-induced failure returns `context.Canceled` after exactly one attempt, which is what the test asserts.
- **Root-cause error preservation** — `errgroup` records the first error *then* cancels, so `g.Wait()` returns the real failure, not a sibling's `context.Canceled`.
- **Ordering** — index-keyed collection preserves listing order independent of completion order; the test pins it across 25 devices.
- **Concurrency invariant** — `peak <= maxConcurrentFetches` holds because each `fetchDevice` issues its requests sequentially. (`assert.Greater(peak, 1)` is loose but appropriately flake-resistant.)
- **No overlap with #733** (`utils/pkg/dberror/*` only), so no conflict.

Test design is good — the fault-injecting `RoundTripper` driving retry/abort/cancel/ordering is the right shape, and asserting exact request counts is what makes the retry bound meaningful.

ben-dz:
Both addressed. Pushed in 78a0810.

**Observability (your first comment).** Took option 2, plus a log line since it was nearly free. `SyncMSDP`/`SyncIPMroute` now report `view_type=msdp|mroute` on `view_refresh_total` and `view_refresh_duration_seconds` and log `msdp: sync completed duration=… status=…`. They go through a small `observeSync` wrapper in `activities.go` rather than the shared `refresh` path, so the other activities keep reporting from their view layers and no `view_type` gets double-counted. Duration was recoverable from `log_ingestion_runs` (that's where the 155s/57s numbers came from), but it wasn't dashboardable next to the other views and the deadline headroom wasn't visible at all.

**Finding 1 — fixed, but the mechanism isn't what it looked like.** Added the `gctx.Err()` check. I then measured it, because I wanted the test to assert something real: with 25 devices, `SetLimit(10)`, and the first device failing persistently, **all 25 functions run in both cases, and in neither case do the 15 late devices issue an S3 request.** `SetLimit` makes `g.Go` block the launching loop rather than queueing goroutines, and the AWS SDK rejects an operation on an already-cancelled context before it reaches the transport (measured at the `RoundTripper`: 10 devices reached, 25 invocations). So the guard saves SDK stack traversal, not ~90 requests. It's still worth having — the skip is explicit instead of depending on SDK internals — but I named the test for the invariant it actually pins ("no S3 calls are issued after the walk aborts") rather than claiming it covers the guard.

**Finding 2 — no change, and I don't think the 3× materializes.** Reaching ~3,600 requests needs nearly every device to fail twice *and* recover on the third attempt. Any failure more persistent than that aborts the walk at the first device to exhaust its attempts (~300ms in), which cancels the rest. And 3,600 requests spread over the new ~20s walk is ~180 req/s, well under S3's per-prefix ceiling. Bounded retry against a genuinely blipping S3 is the behaviour we want here.

**Finding 3 — leaving as is.** Deliberate: retrying at the device boundary keeps one retry site per device, and re-running three LISTs to recover a GET is cheap now that the walk is parallel.

**Finding 4 — addressed.** A device whose prefix exists but has no snapshot in the window now gets a WARN line naming it, and the comment says what the abort does and does not protect. Tombstoning a device that stopped uploading for 3 days is the right outcome for a current-state dimension; the defect was that it happened with no error and no log line, so a device silently vanishing looked identical to normal operation. Covered by a new subtest (snapshot 5 days out → dropped and named in the log).

**Pattern observation.** Agreed, and this is the third one. A worst-case-cost-vs-deadline assertion doesn't belong in this PR, but it's worth its own issue — happy for that to be filed against the dzingest activity options rather than any single sync.

One caveat on all of the above: Docker isn't available in my environment, so the MinIO-backed tests are exercised by CI, not by me. The measurements I quote (attempt counts, devices reached, peak in-flight, classification) come from running the same logic against a fake S3 endpoint under `-race`.

nikw9944 (commented):
One thing worth fixing before this lands: `listDevicePubkeys` is now the only S3 call in the cycle without a retry, so a single blip on the device listing still aborts the whole sync — the failure mode this PR set out to remove. Everything else checks out: the index-keyed collection is race-free with `g.Wait()` supplying the happens-before, listing order is preserved, and routing retries through `dberror.IsTransient` gives the right context-cancel behavior.

indexer/pkg/dz/msdp/source_s3.go
```diff
@@ -5,16 +5,34 @@ import (
...
 
+// maxConcurrentFetches bounds in-flight per-device S3 reads, matching the limit
+// used by the serviceability and shreds views.
+const maxConcurrentFetches = 10
+
+// deviceFetchRetry bounds per-device retries for transient S3 failures. The fetch
+// is all-or-nothing (see Store.Sync), so without a retry one blip among the
+// hundreds of calls a cycle makes aborts the whole thing. dberror.IsTransient
+// treats context errors as non-transient, so a cancelled activity aborts
+// immediately rather than retrying into its deadline.
```
nikw9944: `fetchDevice` now wraps every per-device LIST and GET in `deviceFetchRetry`, but `listDevicePubkeys` (line 210) runs before the fan-out and is still bare — a blip there aborts the whole cycle with zero retries, which is the failure mode this PR removes everywhere else.

Nothing else covers it. `Activities.refresh` (`indexer/pkg/dzingest/activities.go:65-78`) swallows the error and returns `nil`, so the workflow's `MaximumAttempts: 3` retry policy never applies to these activities; the source-level retry is the only one. The SDK retryer underneath doesn't close the gap for the blip class named in the PR description either: aws-sdk-go-v2 does not retry a 200-with-an-error-body response, which is exactly what `dberror`'s `awsRespErrRe` exists to classify (`utils/pkg/dberror/dberror.go:26`).

One line:

```go
devices, err := dberror.Retry(ctx, deviceFetchRetry, func() ([]string, error) {
    return s.listDevicePubkeys(ctx, kind)
})
```

indexer/pkg/dz/mroute/source_s3.go
```diff
@@ -5,16 +5,34 @@ import (
...
 
+// maxConcurrentFetches bounds in-flight per-device S3 reads, matching the limit
+// used by the serviceability and shreds views.
+const maxConcurrentFetches = 10
+
+// deviceFetchRetry bounds per-device retries for transient S3 failures. The fetch
+// is all-or-nothing (see Store.Sync), so without a retry one blip among the
+// hundreds of calls a cycle makes aborts the whole thing. dberror.IsTransient
+// treats context errors as non-transient, so a cancelled activity aborts
+// immediately rather than retrying into its deadline.
```
nikw9944: Same gap as in the msdp source: `listDevicePubkeys` (line 204) runs before the fan-out and isn't covered by `deviceFetchRetry`, so a transient blip on the device listing still aborts the cycle with no retry at any layer (`Activities.refresh` returns `nil`, so Temporal's `MaximumAttempts: 3` never applies). Wrap it the same way:

```go
devices, err := dberror.Retry(ctx, deviceFetchRetry, func() ([]string, error) {
    return s.listDevicePubkeys(ctx)
})
```

indexer/pkg/dz/msdp/source_s3.go
```diff
@@ -5,16 +5,34 @@ import (
...
+var deviceFetchRetry = dberror.RetryConfig{
```
nikw9944, line 32: Still open in `e57e20a`: `listDevicePubkeys` is called bare at line 137, outside `deviceFetchRetry`, so a transient blip on the device listing aborts the whole cycle with zero retries — the failure mode this PR removes for every other S3 call in the walk.

Nothing else covers it. `Activities.refresh` (`indexer/pkg/dzingest/activities.go`) still returns `nil`, so the workflow's `MaximumAttempts: 3` retry policy never applies to these activities; the source-level retry is the only one. And the SDK retryer underneath doesn't retry a 200-with-an-error-body response, which is exactly what `dberror`'s `awsRespErrRe` exists to classify (`utils/pkg/dberror/dberror.go:26`).

One line:

```go
devices, err := dberror.Retry(ctx, deviceFetchRetry, func() ([]string, error) {
    return s.listDevicePubkeys(ctx, kind)
})
```

indexer/pkg/dz/mroute/source_s3.go
```diff
@@ -5,16 +5,34 @@ import (
...
+var deviceFetchRetry = dberror.RetryConfig{
```
nikw9944, line 32: Same gap as in the msdp source, still open in `e57e20a`: `listDevicePubkeys` is called bare at line 124 and isn't covered by `deviceFetchRetry`, so a transient blip on the device listing aborts the cycle with no retry at any layer (`Activities.refresh` returns `nil`, so Temporal's `MaximumAttempts: 3` never applies). Wrap it the same way:

```go
devices, err := dberror.Retry(ctx, deviceFetchRetry, func() ([]string, error) {
    return s.listDevicePubkeys(ctx)
})
```

nikw9944 (commented):
One finding carried over from the previous round, verified still open in `e57e20a`: `listDevicePubkeys` is the only S3 call in the msdp/mroute cycle without a retry, so a single blip on the device listing still aborts the whole sync with no retry at any layer. The follow-up commit addressed observability, the post-abort skip, and the stale-device log, but not this.

The new commits check out. The link multicast-delivery fan-out preserves error precedence, the source-unavailable downgrades, and the response shape; `queryLinkRelatedGroupHealth` issues a single query and never acquires the delivery semaphore, so chaining it inside the groups slot can't self-deadlock. The stale-device log is only reachable after a successful `g.Wait()`, so a nil entry really does mean "no snapshot in the lookback window". The new `view_type=msdp|mroute` metric labels have no other writer and no alert rule keys off them.

Separately, the title and description now cover about half the diff — the api/handlers and api/testing changes aren't mentioned anywhere in the PR body.

- The title and description now cover roughly half the diff. Alongside the indexer fan-out, the head carries a rewrite of `GetLinkMulticastDelivery` into a two-stage fan-out (`api/handlers/multicast_delivery_entity.go`, +138/-45) and a ClickHouse template-clone fix (`api/testing/clickhouse.go`). Both are well-motivated by CI failures on this branch and both look correct, but nothing in the PR body tells a reviewer they exist. Add them to the summary and testing verification, or split them out.

indexer/pkg/dz/msdp/source_s3.go
ben-dz: Fixed in cd0f407, wrapped exactly as you wrote it. You were right about the layering too — I checked `Activities.refresh` and it does unconditionally `return nil`, so `MaximumAttempts: 3` never reaches these activities and the source-level retry really was the only one.

I also fixed the `deviceFetchRetry` doc comment, which said "per-device retries" and so actively described the gap as if it were the design.

On why no test caught this: `faultTransport` attributes a failure by matching a device pubkey in the request URL, and the per-kind listing carries no pubkey, so the existing fault injection structurally could not reach it. Added a `failListFirst` knob keyed on that same property — safe because `newSource` pins region and credentials statically and installs `aws.NopRetryer`, so every request through the transport originates in the source — plus two subtests mirroring the per-device pair: a blip is retried and the walk still returns the full device set, and a persistent failure aborts after exactly `MaxAttempts`. The listing runs once per kind here, so the retry assertion is `listAttempts() > len(snapshotKinds)` rather than `> 1`.

Docker is unavailable in my environment, so CI runs these rather than me.

indexer/pkg/dz/mroute/source_s3.go
ben-dz: Fixed in cd0f407, same shape as the msdp side. Doc comment corrected here too.

Test coverage matches what I described in the msdp thread, minus the per-kind wrinkle: mroute lists once per cycle, so the retry assertion is just `listAttempts() > 1`.

ben-dz:
Both findings addressed in cd0f407 — details in the two inline threads. Thanks for carrying the listing retry forward a second time; it was a real gap and the previous round's follow-up commit did miss it.

On the description: you were right that it covered about half the diff. Rewritten, and the title is now `indexer,api: ...` so it stops implying the change is indexer-only. The `GetLinkMulticastDelivery` fan-out and the `api/testing` template-clone fix are both in the summary and the testing verification now.

I folded them in rather than splitting them out, on the grounds that neither has motivation independent of this branch — both exist only because CI here went red, and the clone fix in particular is a one-line filter that would be a strange PR on its own. Happy to split if you'd rather review them separately.

Two things I found while in there that I deliberately did **not** change, both filed as #748 rather than expanded into this PR:

- `max_execution_time = 30` exactly equals the handler's `context.WithTimeout(r.Context(), 30*time.Second)` in both multicast-delivery handlers. The client deadline wins that race, so an overrun surfaces as `context.DeadlineExceeded`, which `logError` drops as client-disconnect-class — a production occurrence is a 500 with no log line at all. Lowering the SQL cap to ~20s would make ClickHouse fail first with a diagnosable error, but it would also start failing queries that currently finish in the 20–30s band, so it wants a deliberate decision rather than a drive-by.
- `countMulticastLinkDeliveryBranches` is still called inline at the top of `queryMulticastLinkDeliveryBranches`, so two full `enriched_ip_mroute_oifs` scans run back-to-back in one fan-out slot. The fan-out took the link test from a 500 at 31.93s to passing at 15.86s, so this isn't urgent, but it's the obvious next reduction if it tightens again.

Standing caveat on verification: Docker is unavailable in my environment, so every container-backed test in this PR is run by CI, not by me. Locally I have `go vet`, `gofmt`, and a `-race` test-binary compile, plus `TestSkipTemplateTable`, which is deliberately container-free.

nikw9944 (approved):
lgtm — no new commits since the last round, and both earlier findings remain verified as fixed: the device listing is wrapped in `deviceFetchRetry` in both sources with tests covering the retried-blip and bounded-abort paths, and the title and description now cover the api changes.

PR #745, opened 2026-07-30
shreds: publish retransmit-only metro flag to clickhouse

nikw9944:
## Summary of Changes

* Publish the shred-subscription `retransmit_only_enabled` metro flag (bit 0 of the onchain `MetroHistory.flags`) to ClickHouse as `dim_dz_shred_metro_histories_current.retransmit_only_enabled`, and expose it on the internal `/api/dz/shreds/devices` response, inherited from each device's metro.
* Badge it as `retransmit-only` on the shreds devices tab and in the subscribe flow (device picker and selected-device summary), so a subscriber sees that a seat excludes leader shreds before funding it.
* Operators and subscribers currently have no way to tell which metros serve the retransmit group only — the flag exists onchain ([doublezero-shreds#597](https://github.com/malbeclabs/doublezero-shreds/pull/597)) but was not published anywhere.
* Existing rows default to 0 (the bit was reserved and unset onchain, so no backfill). The public `/api/v1/edge/shreds/devices` contract is unchanged.
* Fixes malbeclabs/infra#2091

## Testing Verification

* New `TestGetShredDevices_RetransmitOnlyFromMetro` covers both paths of the metro join: a device whose metro has the flag set reports 1, and a device whose metro has no history row falls back to 0 while still being listed (presence asserted, so the fallback cannot pass vacuously).
* `TestGetShredMetroHistories_WithData` extended to assert the flag round-trips from the history table through the `_current` view.
* Migration follows the `20260722000002_users_feed_pks.sql` convention — `ALTER TABLE` on both the history and staging tables, with the `AFTER current_usdc_price_dollars` placement matching `MetroHistorySchema.PayloadColumns()` order, since `loadSnapshotIntoStaging` appends a positional batch with no column list. Verified by hand against `PayloadColumns()`/`ToRow()`; the ClickHouse-backed Go tests exercise it via the per-test migrated database.
* Adding a payload column changes every metro's `attrs_hash`, so the first indexer run after deploy writes one new SCD2 history row per metro — expected, same as the prorated-fields change.


indexer/pkg/dz/shreds/schema.go
```diff
@@ -148,6 +148,7 @@ func (s *MetroHistorySchema) PayloadColumns() []string {
 		"total_initialized_devices:INTEGER",
 		"current_epoch:BIGINT",
 		"current_usdc_price_dollars:INTEGER",
+		"retransmit_only_enabled:BOOLEAN",
```
ben-dz, line 151: The positional column-order invariant the migration calls "load-bearing" is pinned by no test. The staging insert is positional (`dim_write.go:527`, `PrepareBatch("INSERT INTO stg_…")` with no column list), so `PayloadColumns()` order must equal the staging table's physical order — verified here only by hand. The handler tests write to the *history* table with an explicit column list, and `dim_write_test.go` uses synthetic tables, so neither would catch drift; `indexer/pkg/dz/shreds/` has no test files at all.

Fix: one integration test in that package following `indexer/pkg/dz/mroute/migration_test.go` — `laketesting.NewClientWithInfo` already runs the real migrations — calling `Store.ReplaceMetroHistories` with `RetransmitOnlyEnabled: true` plus a distinct `CurrentUSDCPriceDollars` and asserting both in `dim_dz_shred_metro_histories_current`. Needs the package's first `TestMain`/`sharedDB` (copy mroute's). Cheaper alternative: assert `system.columns` ordering for `stg_dim_dz_shred_metro_histories_snapshot` equals internal+PK+payload order.

api/handlers/shreds_devices.go
```diff
@@ -114,6 +118,7 @@ func (a *API) GetShredDevices(w http.ResponseWriter, r *http.Request) {
 			&d.IsEnabled,
 			&d.BasePriceDollars, &d.PremiumDollars, &d.TotalPriceDollars,
 			&d.GrantedSeats, &d.Capacity, &d.AvailableSeats,
+			&d.RetransmitOnlyEnabled,
```
ben-dz, line 121: This scan loop never checks `rows.Err()`, so a mid-stream failure — including the 10s `context.WithTimeout` at line 50 firing — returns a truncated device list with HTTP 200, on the endpoint that feeds the subscribe picker. Pre-existing, but the siblings in this same feature do check it (`shred_devices_fetch.go:68`, `shred_subscribers.go`), and 55 of 67 handler files follow that convention.

Fix: after the loop, `if err := rows.Err(); err != nil { logError(...); http.Error(w, dberror.UserMessage(err), 500); return }`. Same for `GetShredMetroHistories` at `shreds.go:541`.

ben-dz (commented):
No blocking findings — the migration's positional/`AFTER` contract holds, the SDK bit-0 → `UInt8` path round-trips, and the v1 public shape is genuinely unchanged.

Two things worth addressing:

- The positional column-order invariant the migration itself calls load-bearing is pinned by no test — nothing exercises `loadSnapshotIntoStaging` for this dimension, so future drift between `PayloadColumns()` and the staging DDL would go uncaught. One integration test in `indexer/pkg/dz/shreds/` following the `mroute/migration_test.go` pattern would close it.
- `FetchEdgeShredsDevices` leaves the new field zero on the shared `ShredDeviceItem`, which is a trap for whoever later adds it to the v1 shape.

Also: the devices scan loop returns a truncated list with HTTP 200 on a mid-stream/timeout failure, unlike its siblings in the same feature.

- `api/handlers/shred_devices_fetch.go:33` — the shared `ShredDeviceItem` leaves `RetransmitOnlyEnabled` zero on the edge path. `FetchEdgeShredsDevices` neither selects nor scans the column, so every row reports "not retransmit-only"; harmless today only because `api/v1/shreds_devices.go` drops the field. Whoever later adds it to `EdgeShredsDevice` gets a silent all-zero public field, and with every metro currently unset that tests clean. Fix: either add `COALESCE(mh.retransmit_only_enabled, 0)` to the select and `&d.RetransmitOnlyEnabled` to the scan (the `mh` join is already there, line 37), or note on the struct field that only `GetShredDevices` populates it.

indexer/pkg/dz/shreds/schema.go
nikw9944: Added in c818db5 — the package's first ClickHouse integration tests (`main_test.go` copies mroute's `TestMain`/`sharedDB`): `TestStore_MetroHistories_RoundTrip` writes through `Store.ReplaceMetroHistories` with `RetransmitOnlyEnabled: true`, opposite-valued flags, and distinct values for every payload column, asserting all of them in `dim_dz_shred_metro_histories_current`; plus your cheaper alternative, `TestMetroHistoriesStagingColumnOrder`, which asserts `system.columns` ordering for the staging table equals internal+PK+payload. I kept the ordering assertion scoped to metro histories rather than sweeping all eight shreds datasets — no Docker in this environment, so I can't locally verify the other seven wouldn't trip on pre-existing drift; happy to extend it in a follow-up if you want the sweep.

api/handlers/shreds_devices.go
nikw9944: Fixed in c818db5 — both `GetShredDevices` and `GetShredMetroHistories` now check `rows.Err()` after the scan loop and return 500 via `dberror.UserMessage`, matching the convention in `access_passes.go` and the v1 fetch path. Noted in the CHANGELOG entry since a truncated-200 on the subscribe picker is user-visible.

nikw9944:
Addressed both review threads in c818db5:

- **Positional column-order invariant now pinned by tests**: added `indexer/pkg/dz/shreds`'s first ClickHouse integration tests — an SCD2 round-trip of every `MetroHistoryRow` payload column through `ReplaceMetroHistories` into the `_current` view, and a `system.columns` ordering assertion for the staging table (both suggested fixes, since each catches a different failure shape). Scoped to metro histories; extending the ordering assertion to the other shreds datasets would be a cheap follow-up.
- **`rows.Err()` checks** added to `GetShredDevices` and `GetShredMetroHistories`, so a mid-scan stream failure returns 500 instead of a truncated 200.

Verified: `go build`, `go vet`, `golangci-lint` clean; the new tests compile and reach `TestMain` (they need Docker for the ClickHouse testcontainer, so they first execute in CI — same as the mroute package they mirror). Nothing was left unaddressed.

ben-dz:
One item from the original review is still open: FetchEdgeShredsDevices leaves RetransmitOnlyEnabled zero on the shared ShredDeviceItem, which is a trap for whoever later adds the field to the v1 shape. Probably worth fixing before merging.

nikw9944:
Fixed in 1baa422 — `FetchEdgeShredsDevices` now selects `COALESCE(mh.retransmit_only_enabled, 0)` and scans it into the shared `ShredDeviceItem`, so the field is populated on the v1 fetch path too. The public v1 response shape is unchanged (it's still shaped explicitly in `api/v1/shreds_devices.go`); a direct test (`TestFetchEdgeShredsDevices_PopulatesRetransmitOnly`) pins the populated value so it can't silently regress to zero. Branch is also rebased onto main (CHANGELOG conflicts only).

ben-dz (commented):
Re-checked at `8042314` (branch rebased onto current `main`). All three findings from the previous round are fixed in code, and CI is green on this head — so the new ClickHouse integration tests have actually executed, closing my earlier caveat that they were unproven:

One new low-severity item: because goose tracks applied migrations by version id with no filename or checksum check, any database that ran this branch under the old `20260730000001` will now silently skip `main`'s `20260730000001_entity_changes_feeds.sql`. Staging and prod only ever ran `main` and are unaffected; local dev databases of anyone who checked the branch out are not. Worth a line in the PR body telling them to reset before pulling `main`.

- The migration renumber in `8042314` strands any database that already applied the migration under the old number. `goose_db_version` is keyed by version id with no filename or checksum check, so a database that ran this branch before the rename has `20260730000001` recorded from the shreds migration and will skip `main`'s 681-line `20260730000001_entity_changes_feeds.sql` as already applied, while re-running the renamed `20260805000001` (harmless — `ADD COLUMN IF NOT EXISTS` and `CREATE OR REPLACE VIEW` are idempotent). Staging and prod only ever ran `main`, so they are unaffected; local k3d dev databases of anyone who checked this branch out are. Fix: a line in the PR body telling them to `./scripts/k8s.sh down --clean` (or delete the `20260730000001` row) before pulling `main`.

ben-dz (approved):
See previous nit, but stamping so you can merge after that.

PR #746, opened 2026-07-30
api: finish the vote/stake/gossip cache work

### LINKED ISSUE #743: api: finish the vote/stake/gossip cache work — 3 total_stake CTEs outside timeline.go + validators cache keyed to the wrong limit (~7% of cluster)
## Summary

#736 (PR A of #732) delivered one of its three targets. Post-deploy measurement in prod (2026-07-30, equal 20-min windows) — [full data](https://github.com/malbeclabs/lake/issues/732#issuecomment-5132863276):

| Member | Result |
|---|---|
| `current_slot_info` ×2 (15s scalar cache) | ✅ **eliminated** — 186 runs → 0, ~91K CPU-sec/day (~4% of cluster) reclaimed |
| `total_stake` | ❌ unchanged — 129 runs / 1,677 CPU-sec per 20 min (~120K/day, ~5% of cluster) |
| validators listing | ❌ unchanged — 119 runs / 920 CPU-sec per 20 min (~66K/day, ~3%) |

**~7% of cluster CPU is still on the table**, and both misses are now diagnosed precisely.

## Miss 1: three `total_stake` CTE copies were outside `timeline.go`

#736 converted the `timeline.go` copies, but on current `main` the CTE still appears in:

- `api/handlers/validators.go:175`
- `api/handlers/validators.go:418`
- `api/handlers/stake.go:315`

These still inline `WITH total_stake AS (SELECT sum(activated_stake_lamports) as total …)`, which is why qhash `12413159124723682455` survives at its pre-deploy rate. Fix: point the existing cached scalar (from #736) at these three call sites — no new mechanism needed.

## Miss 2: the validators page cache is keyed to a request shape nobody sends

The `cache=validators` entry is live and refreshing every ~60s, but **100% of prod traffic to `/api/solana/validators` uses `limit=900`** (measured post-deploy: 150 req/25m = 6.0/min, matching the 119 surviving queries exactly). #736 made the *default first page* (`limit=100` = `PAGE_SIZE`) cache-eligible; the external poller that drives ~95% of requests uses `limit=900` and bypasses the cache on every call.

Pre-PR evidence showed this split (8,639 req/24h at `limit=900` vs 491 at `limit=100`, [data](https://github.com/malbeclabs/lake/issues/732#issuecomment-5107339303)) — the eligibility rule was simply built around our own UI's shape rather than the measured traffic.

Fix options (pick one, or 1+3):
1. **Widen eligibility** to the polled shape — a `limit=900`, `offset=0`, stake-desc, unfiltered request is exactly as cacheable as `limit=100`. Simplest.
2. Cache the full stake-ordered set once and slice pages from it — covers any `limit`, at the cost of holding more in memory.
3. **Add `Cache-Control: max-age=60`** on the cached response so the client self-throttles. Worth doing regardless of 1 or 2; free.

## Verification

Each shape has its own `normalized_query_hash`, so measure them **separately** post-deploy:

- `12413159124723682455` (`total_stake`) — expect ~0 outside cold-start
- `2239025009229925591` (validators listing) — expect ~1/min (the 60s cache refresh) instead of ~6/min

Careful with attribution: `validators.go` holds two of the CTE copies *and* is the listing handler, so the two fixes may overlap. Don't assume one fix covers both — the per-hash numbers will tell you.

Baselines to compare against are pinned on #732; snapshot fresh ones at PR time too, since replica scaling wipes `query_log`.

## Context

#732 (parent, PR A = #736) · #738 (PR B: edge-scoreboard leader map + v1 metadata) · infra#1837 (cluster tracking). Cluster is currently ~2.2M CPU-sec/day (~26 cores), down ~65% from 6.3M three weeks ago.

nikw9944:
## Summary of Changes

* Point the three `total_stake` CTE copies that survived #736 (validators listing, validator detail, DZ user detail) at a new TTL-cached scalar, and cache the **complete** stake-desc validator set so the handler can slice any requested page out of it.
* #736 delivered one of its three targets; the other two left ~7% of shared-ClickHouse CPU on the table. The CTE re-scanned a window-function view per request, and the page cache was keyed to `limit=100` while 100% of measured production traffic uses `limit=900` and bypassed it on every call.
* Caching the whole set rather than widening the gate to `limit=900` means eligibility no longer depends on the client's page size, so this doesn't re-break when the external poller changes its shape.
* The new scalar is deliberately separate from #736's `cachedTotalStake`: these sites filter on `epoch_vote_account`, and reusing the unfiltered total would have enlarged every denominator and shrunk every `stake_share`.
* Cached responses carry `Cache-Control: public, max-age=60` so pollers self-throttle, with `Vary: X-DZ-Env` since the env arrives as a header and isn't part of the URL.
* Fixes #743

## Testing Verification

* Handler-level test (`NewTestAPIAll`, both containers) asserts a cached page equals the live page field-for-field for `limit=900&offset=0` and `limit=100&offset=1`, and pins `X-Cache`/`Cache-Control`/`Vary` on both the HIT and MISS paths. Nothing exercised the HIT path before — the other validators tests use `NewTestAPI`, which leaves `PgPool` nil.
* One known live/cached divergence is pinned rather than fixed: for `offset >= total` the live path reports `total: 0`, because `fetchValidatorsPage` only reads its window aggregates off returned rows, while the cached path reports the real count. The cached path is the more correct of the two; fixing the live path needs a whole-set count for a page that returns no rows (a second full query) and applies equally to filtered and non-stake-sorted shapes, so it belongs in its own change. Worth filing as a follow-up.
* A stale-entry test pins that an entry older than 3x the refresh cadence falls through to live, so a stopped worker can't pin an answer indefinitely.
* `sliceCachedValidators` table test covers the polled `limit=900` shape, a page clamped at the end of the set, an offset past the end (encodes `[]`, not `null`), and the three fall-through-to-live cases: a truncated entry, an empty entry, and an unparseable payload. Verified passing by extracting the real function into a standalone module, since `api/handlers`' `TestMain` starts containers and Docker is unavailable locally — CI runs them in place.
* `isCacheableValidatorsRequest` test pins that `limit=900`, `limit=1000`, and deep offsets are all eligible (the shapes the old gate sent live), while any filter or non-stake-desc sort still bypasses the cache.
* Added an invariant test that `validatorsCacheMaxRows > MaxLimit` — lowering the cap to a request-sized value would leave every entry incomplete and silently revert this fix with only a WARN to show it.
* Measured the cached payload at ~390 B/row: ~490 KB at today's ~1,300 validators, ~1.9 MB at the 5,000-row cap. Comparable to existing `edge_scoreboard`/`topology` entries.
* `api/worker` tests pass with `-race` (they run without Docker); the `validators` entry still refreshes every other slow cycle, now asserted against the exported key constant rather than a literal.
* Deploy-window behavior reasoned through in both directions: the new key means neither old nor new pods can serve the other's payload, and the completeness check catches a capped entry.

## How to verify this worked in production

**Don't measure it by the two query hashes pinned on the issue.** They will both go to zero
runs, and that proves nothing. ClickHouse groups `query_log` rows by a fingerprint of the SQL
text; this PR edits that text at every site, so the old fingerprints vanish on deploy no matter
what the rewritten queries cost. Zero would look like a total win even if nothing got cheaper.

Two ways to actually measure it, easiest first.

**1. API Prometheus counters.** These are labeled by handler name, not by SQL text, so a query
rewrite can't disturb them. No ClickHouse access needed. Compare the rate before and after:

- `doublezero_lake_api_clickhouse_queries_total{name="validators"}` — the listing and detail queries
- `{name="users"}` — the DZ user detail query
- `{name="scalar_cache:epoch_vote_total_stake"}` — the new cached scalar

**2. `query_log`, finding the new queries by text** instead of looking up the old fingerprint:

- the cached scalar: `query LIKE 'SELECT sum(activated_stake_lamports) FROM solana_vote_accounts_current WHERE epoch_vote_account%'`
- the rewritten listing: `query LIKE '%validators_with_gossip%' AND query NOT LIKE '%total_stake%'`

**What success looks like:**

| Metric | Before | After |
|---|---|---|
| Validators listing query | ~6/min (every poll hits ClickHouse) | ~1/min (only the cache refresh) |
| Total-stake scan | Once per request | ≤1 per 5 min per env (its cache TTL) |
| `total_stake` CPU | 111,643 CPU-sec/day | ~0 outside cold start |
| Listing CPU | 65,728 CPU-sec/day | ~0 outside cold start |

If the external poller honors the new `Cache-Control` header, that shows up as the same 6/min → 1/min
drop measured at the endpoint rather than at the query.

**One caveat on attributing the win.** `validators.go` receives both fixes, so the drop on that
endpoint can't be split between them. The CTE removal still earns its place on its own: it also
covers `GetValidator`, `GetUser`, and the cache refresh itself, none of which the caching fix touches.


api/handlers/validators.go
```diff
@@ -93,27 +94,51 @@ var validatorsQuerySortFields = map[string]string{
...
 
-	// The default shape (first page, stake desc, no filters) is polled continuously
-	// and served from the page cache on mainnet. Other shapes bypass the cache.
-	if isMainnet(r.Context()) && isDefaultValidatorsRequest(pagination, sort, filters) {
+	// The unfiltered stake-desc listing is polled continuously (by the UI at
+	// limit=100 and by an external consumer at limit=900). The cache holds the
+	// complete set, so any page of it can be served without a ClickHouse query.
+	// Filtered or differently-sorted shapes bypass the cache.
+	if isMainnet(r.Context()) && isCacheableValidatorsRequest(sort, filters) {
 		if data, err := a.readPageCache(r.Context(), ValidatorsPageCacheKey); err == nil {
```
ben-dz: This read runs on the caller's `r.Context()` — the 20s bound isn't established until line 146, after this branch — and `readPageCache` returns whatever row exists with no `updated_at` check.

That combination mattered little before: only `limit=100&offset=0` was cache-served, so the external ~10s poller at `limit=900` and every deeper page went to ClickHouse under a deadline. Now 100% of unfiltered stake-desc traffic reads a payload that grew from ~40 KB to ~490 KB out of a pool capped at `MaxConns = 10` (`api/config/postgres.go:79`), shared with the auth and nonce queries. A Postgres slowdown that previously degraded one request shape now stalls the whole endpoint for as long as clients wait, unbounded.

The staleness half is narrower than it looks — query and write failures both escalate via `logger.Escalator` (`api/worker/workflow.go:466`, `:475`), so an indefinitely frozen entry needs the worker to not run at all — but it is now the whole endpoint rather than one shape, with `Cache-Control: max-age=60` stacked on top.

`fetchPublisherCheckCachedOrLive` (`api/handlers/publisher_check.go:136-155`) already has the shape: it bounds the read ctx ("it must not run on the caller's unbounded ctx") and rejects a payload older than `publisherCheckStaleAfter` with a single WARN. Swap in `readPageCacheWithAge` under a `context.WithTimeout(r.Context(), …)` and fall through to live past ~3x the refresh cadence.

api/handlers/validators.go
```diff
@@ -141,28 +168,88 @@ func (a *API) GetValidators(w http.ResponseWriter, r *http.Request) {
...
+
+	// Both bounds are clamped to the set, so a page past the end is an empty (but
+	// non-nil, hence still JSON "[]") slice rather than an out-of-range panic.
+	start := min(p.Offset, len(cached.Items))
+	end := min(start+p.Limit, len(cached.Items))
+
+	return &ValidatorListResponse{
+		Items: cached.Items[start:end],
+		// Whole-set aggregates, independent of the page being served.
+		Total:     cached.Total,
```
ben-dz, line 264: These whole-set aggregates diverge from the live path when `offset >= Total`. In `fetchValidatorsPage`, `total` and `onDZCount` are only assigned inside the `rows.Next()` loop (`validators.go:373`, `:409-410`), so a page past the end returns zero rows and reports `total: 0, on_dz_count: 0`, while this returns the whole-set values. Same URL, different `total`, decided only by cache warmth — and the cache is guaranteed cold for one refresh cycle after every deploy because of the key rename. The web pager reads `response.total` (`web/src/components/validators-page.tsx:244`).

This path is the more correct of the two; the live path is the wart. Either way the "byte-for-byte equivalent page" premise doesn't hold, and nothing would catch it: the new tests cover only the pure slicer, and `TestGetValidators` uses `NewTestAPI`, which leaves `PgPool` nil (`api/testing/api.go:36-52`), so `readPageCache` short-circuits on `errNoPgPool` and the live path is the only one ever exercised. Nothing asserts a cached page equals the live page, nor that `X-Cache`/`Cache-Control`/`Vary` are emitted.

`NewTestAPIAll(t, chDB, pgDB, nil, nil)` (`api/testing/api.go:70`) wires both containers. Seed >1 validator, `WritePageCache(ValidatorsPageCacheKey, FetchValidatorsData(ctx))`, then assert the `limit=900&offset=0`, `limit=100&offset=1`, and `offset=Total` responses field-for-field against the same requests with the row deleted.

api/handlers/validators.go
```diff
@@ -93,27 +94,51 @@ var validatorsQuerySortFields = map[string]string{
...
+				w.Header().Set("Content-Type", "application/json")
+				w.Header().Set("X-Cache", "HIT")
+				// Let the polling clients self-throttle to the refresh cadence.
+				w.Header().Set("Cache-Control", "public, max-age=60")
+				// The response body depends on the env header, which is not in the URL
+				// (EnvMiddleware reads X-DZ-Env first, ?env= only as a fallback). Without
+				// this, a cache would replay this mainnet payload for a devnet request to
+				// the same URL — starting with the browser's own cache when the UI's env
+				// selector switches.
+				w.Header().Add("Vary", "X-DZ-Env")
```
ben-dz: `Vary` is only emitted on the HIT path — the MISS response (line 144 onward) carries none, so whether a shared cache can distinguish envs depends on server cache state. Move this above the cache branch, alongside the freshness headers.

ben-dz (commented):
Two things worth changing before this lands, neither blocking.

1. The cache read at `validators.go:125` runs on the caller's unbounded `r.Context()` (the 20s bound is only established at line 146, after the branch) and has no age gate. That was tolerable when only `limit=100&offset=0` was cache-served; now 100% of unfiltered stake-desc traffic reads a ~490 KB payload — up from ~40 KB — out of a 10-connection pool shared with the auth queries. `fetchPublisherCheckCachedOrLive` (`publisher_check.go:136-155`) already does exactly the right thing with `readPageCacheWithAge` under a bounded ctx.

2. `Total`/`OnDZCount` diverge between the cached and live paths for `offset >= Total`, and no test exercises the HIT path through the handler at all — `TestGetValidators` uses `NewTestAPI`, which leaves `PgPool` nil, so `readPageCache` short-circuits and only the live path ever runs. `NewTestAPIAll(t, chDB, pgDB, nil, nil)` wires both containers, so one test can pin cached-page == live-page.

The injection surface, the NULL/zero denominator, the slice bounds, the completeness gate, cross-env keying, the key rename's rolling-deploy rationale, and the separation of `cachedEpochVoteTotalStake` from `cachedTotalStake` all check out.

api/handlers/validators.go
nikw9944: Fixed in 41d80ed. The read moved into `cachedValidatorsPage`, which bounds it at `validatorsCacheReadTimeout` (2s) and rejects an entry older than `validatorsCacheStaleAfter` (3 min = ~3x the `everyN=2` x 30s cadence, the same ratio `publisherCheckStaleAfter` uses against its own). Every failure mode — no entry, read error, stale, incomplete — returns false and falls through to live.

2s rather than reusing the 20s live bound: it's a single-row point lookup, so that's 20x+ headroom, and a 20s stall is close to unbounded from the caller's side. Worth naming the tradeoff you get either way — on Postgres slowness we now bail to the 13-CPU-sec ClickHouse query, which is the expensive fallback. That's inherent to bounding it, and it's what the endpoint did for every shape before this PR.

api/handlers/validators.go
nikw9944: Fixed in 41d80ed — `Vary` is now set once before the cache branch, so MISS carries it too. `Cache-Control` stays HIT-only, since a MISS shouldn't advertise freshness. Both are asserted on both paths in `TestGetValidators_CachedPageMatchesLive`.

api/handlers/validators.go
nikw9944: Test added in 41d80ed; the divergence I left in place, deliberately.

`TestGetValidators_CachedPageMatchesLive` uses `NewTestAPIAll`, seeds 3 validators at distinct stakes, captures the live responses for `limit=900&offset=0`, `limit=100&offset=1` and `offset=3`, then writes the entry via `FetchValidatorsData` and asserts each cached response field-for-field against its live baseline, plus `X-Cache`/`Cache-Control`/`Vary` on both paths. You were right that nothing exercised the HIT path before — `NewTestAPI` leaves `PgPool` nil.

On the divergence: I agree this path is the more correct one, which is why I didn't "fix" it by making the cached path report 0 as well. But fixing the live path needs the whole-set count for a page that returns no rows, i.e. a second full query, and it applies equally to filtered and non-stake-sorted shapes that this PR doesn't otherwise touch. So it's now pinned on both sides (the `offset=3` case asserts live `total: 0` and cached `total: 3`) with the reasoning in a comment on the struct literal, and left for its own change. Happy to file that as a follow-up issue — say the word and I will.

You're also right that the "byte-for-byte" framing in my PR description was too strong; it holds for pages inside the set, not past the end. Corrected there.

One thing your suggestion surfaced: both cache tests had to drop `t.Parallel()` and delete the row on cleanup. `SetupPostgresForTest` has no per-test isolation, so two parallel tests writing `ValidatorsPageCacheKey` race each other — the same convention `cleanupPublisherCheckCache` documents.

nikw9944:
Thanks — both points were real. Addressed in 41d80ed, on top of a rebase onto current `main` (#742).

**Bounded + age-gated cache read.** The read moved into `cachedValidatorsPage`, with its own 2s deadline and a 3-minute staleness bound (~3x the `everyN=2` x 30s cadence, matching `publisherCheckStaleAfter`'s ratio). No entry, read error, stale entry, or incomplete payload all fall through to live.

**`Vary` on both paths.** Set once before the cache branch. `Cache-Control` stays HIT-only — a MISS shouldn't advertise freshness.

**The missing HIT-path test.** `TestGetValidators_CachedPageMatchesLive` uses `NewTestAPIAll`, seeds 3 validators, and asserts each cached response field-for-field against its live baseline plus the headers on both paths. Your read was right that `NewTestAPI` left `PgPool` nil so only the live path ever ran.

**What I did not change:** the `total`/`on_dz_count` divergence at `offset >= total`. I agree the cached path is the more correct one, so I didn't level it down — but fixing the live path needs a whole-set count for a page that returns no rows, i.e. a second full query, and it applies equally to the filtered and non-stake-sorted shapes this PR doesn't touch. It's now pinned on both sides with the reasoning in a comment, and flagged in the PR description as a follow-up. Happy to file the issue if you want it tracked.

Your suggestion also surfaced something I'd have shipped broken: both cache tests had to drop `t.Parallel()` and delete the row on cleanup, since `SetupPostgresForTest` has no per-test isolation and two parallel tests writing `ValidatorsPageCacheKey` race each other — the convention `cleanupPublisherCheckCache` already documents.

Note on verification: Docker isn't available in my environment, so the two new container-backed tests are the one part I could not execute locally — they're verified by CI. Everything else (build, vet, lint, `api/worker` with `-race`, and the pure slicer logic) is green locally.

api/handlers/validators.go
```diff
@@ -93,26 +94,64 @@ var validatorsQuerySortFields = map[string]string{
...
+	// caller's unbounded ctx: every unfiltered stake-desc request now reads a ~490 KB
+	// payload from a pool capped at MaxConns=10 and shared with the auth queries, so
+	// a Postgres slowdown would otherwise stall the whole endpoint for as long as
+	// clients wait. Falling through to the live query is the better failure mode.
+	validatorsCacheReadTimeout = 2 * time.Second
+
+	// validatorsCacheStaleAfter rejects a frozen entry: ~3× the refresh cadence
+	// (validatorsListingEveryN=2 × the 30s default interval), the same ratio
+	// publisherCheckStaleAfter uses. Refresh failures already escalate via
+	// logger.Escalator, so this only catches a worker that isn't running at all —
```
ben-dz: This bound is hardcoded at 3 minutes and the comment derives it from "`validatorsListingEveryN=2` × the 30s default interval" — but the interval is configurable and the gate doesn't track it. `loadRefreshConfig` reads `PAGE_CACHE_REFRESH_INTERVAL` clamped to 5s–10min (`api/worker/pagecache.go:174`, `api/worker/workflow.go:53-54`), and the effective cadence is `everyN × interval` (`dueThisCycle`, `api/worker/workflow.go:139-144`).

An entry's age sweeps from 0 to the full cadence across each cycle, so the degradation is proportional rather than a hard cutoff:

| interval | cadence | share of each cycle rejected as stale |
|---|---|---|
| 30s (default) | 60s | 0% |
| 60s | 120s | 0% |
| 90s | 180s | ~0%, at the boundary |
| 120s | 240s | ~25% |
| 600s (max) | 1200s | ~85% |

An operator who raises the interval — plausibly to relieve the very ClickHouse pressure this work exists to reduce, and a move the suite already exercises at 90s (`api/worker/pagecache_test.go:134`) — silently sends a growing fraction of `/api/solana/validators` back to the 13-CPU-sec live query. The only signal is the `logWarn` below, which then fires once per affected request (~6/min from the poller alone), with no counter and nothing that alerts.

For context: `publisherCheckStaleAfter = 6 * time.Minute` against a 2-minute cadence (`publisher_check.go:94-99`) has the identical coupling, so the pattern is pre-existing and followed faithfully here. What makes this instance worth closing is that the bound is tighter, so it breaks at a lower interval, and the consequence is the reversal of this PR's stated purpose rather than one page getting slower.

Suggested fix, in the spirit of what's already here: add the same kind of invariant test the row cap got in `TestValidatorsCacheHoldsMoreThanAnyRequest`, asserting the bound covers the worst configurable cadence. It fails today, which forces the choice to be explicit — either raise this constant past `validatorsListingEveryN × maxRefreshInterval`, or lower `maxRefreshInterval` (`api/worker/workflow.go:54`) to a ceiling the consumers' bounds actually cover. Deriving the bound from the live interval is cleanest but needs plumbing, since `api/worker` imports `api/handlers` and not the reverse.

api/handlers/validators.go
```diff
@@ -141,28 +182,117 @@ func (a *API) GetValidators(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	// Both bounds are clamped to the set, so a page past the end is an empty (but
+	// non-nil, hence still JSON "[]") slice rather than an out-of-range panic.
+	start := min(p.Offset, len(cached.Items))
+	end := min(start+p.Limit, len(cached.Items))
+
+	return &ValidatorListResponse{
+		Items: cached.Items[start:end],
+		// Whole-set aggregates, independent of the page being served.
```
ben-dz, line 254: Answering the offer in the thread: yes, please file the follow-up. The deferral reasoning checks out — reporting a whole-set count for a page that returns no rows genuinely needs a second query, and the wart applies equally to the filtered and non-stake-sorted shapes this PR doesn't touch — and the containment here (this comment plus a test pinning both sides) is the right call for now. The reason to have an issue anyway is that a code comment on the *cached* path is the one place a future reader of the *live* path won't look.

ben-dz (commented):
Re-reviewed at `41d80ed`. Both prior Mediums and the Low are verified fixed in code, not just claimed: the cache read is now bounded at 2s and age-gated via `readPageCacheWithAge` in `cachedValidatorsPage`, `Vary` moved above the cache branch, and `TestGetValidators_CachedPageMatchesLive` / `TestGetValidators_StaleCacheFallsThroughToLive` exercise the HIT path field-for-field with correct cleanup ordering and non-parallel execution.

One new issue in the added code: `validatorsCacheStaleAfter` is a hardcoded 3 minutes derived from the *default* 30s refresh interval, but `PAGE_CACHE_REFRESH_INTERVAL` is configurable to 10 minutes and the effective cadence is `everyN × interval`. Above a 90s interval a growing share of each cycle reads as stale and goes live — at the configurable max, ~85% — which silently reverts this PR's CPU win with only a per-request WARN as signal. `publisherCheckStaleAfter` has the same coupling, so the pattern is pre-existing, but this bound is tighter and the consequence is larger.

Also: yes, please file the follow-up issue for the `offset >= Total` live/cached divergence. The deferral itself is well-reasoned and the containment (comment plus a test pinning both sides) is right, but a code comment is the one place a future reader of the live path won't look.

api/handlers/validators.go
nikw9944: Fixed in 8ef1e42 — took the "raise the constant" option, now 45 minutes.

Your table made me re-examine what the gate is actually for, and the answer argues for a loose bound rather than a cadence-tracking one. It's a dead-worker backstop: refresh failures already escalate via `logger.Escalator`, and stake/geo move on ~2-day timescales, so a 45-minute-stale entry is immaterial. Rejecting a *healthy* entry, by contrast, costs a ~13 CPU-sec query and has no upside at all. The asymmetry says size it so it can never fire in healthy operation at any permitted interval, and 45 min clears the 20-minute worst-case cadence plus `maxActivityTimeout` for the refresh itself.

`TestValidatorsCacheBound` in `api/worker` asserts it against `validatorsListingEveryN`, `maxRefreshInterval` and `maxActivityTimeout` rather than a second hardcoded duration — that package is the only one that sees both sides, per your note about the import direction, which is why the constant is now exported. I confirmed it fails at the old 3-minute value before keeping the new one:

```
handlers.ValidatorsCacheStaleAfter (3m0s) must exceed the worst-case healthy entry age
(20m0s cadence + 10m0s refresh); otherwise raising PAGE_CACHE_REFRESH_INTERVAL silently
sends validators listing traffic back to the live query
```

I didn't add a counter for the stale-fallthrough. With the bound this loose it can only fire when the worker isn't running, which the api-down/crash-loop alerts already cover, so a metric would be measuring a condition that's alerted on elsewhere.

Left `publisherCheckStaleAfter` alone — same coupling, but changing it is a behavior change to an endpoint this PR doesn't touch.

api/handlers/validators.go
nikw9944: Filed as #750, and 8ef1e42 adds the pointer where you said a reader would actually be — on the live path's `var total, onDZCount uint64`, noting the aggregates ride on the returned rows and that it affects every shape reaching that path, not just this listing.

The issue writes up the cause, the live-vs-cached table, why it was split out, and a fix sketch — including the prior question of whether a page past the end should return an empty success at all, versus clamping to the last page or 404ing.

nikw9944:
Both points addressed in 8ef1e42, rebased onto current `main` (#744). CI was fully green on `41d80ed`, including `go-test` — the two new container-backed tests passed, which closes the one thing I couldn't verify locally last round.

**Staleness bound (`ValidatorsCacheStaleAfter`): 3 min → 45 min.** Your table was right, and working through it changed how I think the bound should be chosen. The gate is a dead-worker backstop, not a freshness mechanism — refresh failures already escalate via `logger.Escalator`, and this data moves on ~2-day timescales. So a stale entry costs nothing, while rejecting a *healthy* one costs a ~13 CPU-sec query. That asymmetry says size it to never fire in healthy operation at any permitted interval, rather than track the cadence. 45 minutes clears the 20-minute worst-case cadence plus `maxActivityTimeout`.

`TestValidatorsCacheBound` asserts that against the worker's own constants instead of a second hardcoded duration; it lives in `api/worker` since that's the only package seeing both sides, which is why the constant is exported. Verified it fails at the old 3-minute value.

**Follow-up filed: #750**, with the pointer added on the live path's `var total, onDZCount uint64` where a reader of that code will actually hit it.

**Not changed:** `publisherCheckStaleAfter` (same coupling, but a behavior change to an endpoint outside this PR), and no counter for the stale fallthrough — with the bound this loose it can only fire when the worker isn't running, which the api-down alerts already cover.

ben-dz (approved):
lgtm. Both prior rounds' findings are verified fixed in code at `8ef1e42`.

The staleness bound is now `ValidatorsCacheStaleAfter = 45m`, sized so it can never reject a healthy entry: `validatorsListingEveryN=2 × maxRefreshInterval(10m) + maxActivityTimeout(10m)` = a 30-minute worst case. `TestValidatorsCacheBound` asserts that against the worker's own constants rather than a second hardcoded number, from `api/worker` since that's the only package that sees both sides — ran it locally with `-race`, passes. Reframing the gate as a dead-worker backstop rather than a freshness mechanism is the right read, and declining the counter is consistent with it.

#750 is filed and the pointer sits on the live path's `var total, onDZCount uint64`, which is where a future reader of that code will actually be.

One thing to glance at before merging: `go-test` is still pending on this head. It passed on the previous one and the delta is a constant rename, comments, and one pure-constant test, so I don't expect it to catch anything.

PR #747, opened 2026-07-30
indexer: break the telemetry-usage catch-up livelock with an adaptive span

### LINKED ISSUE #740: indexer: telemetry-usage catch-up livelocks when a cycle exceeds the activity budget — staging watermark frozen 22.6h across 76 identical windows
## Summary

Telemetry-usage catch-up can **livelock**: when a catch-up cycle's work exceeds the activity's `StartToClose` budget, the insert is cancelled, the watermark never advances, and the **next cycle re-queries the identical window** — forever. Observed on staging: the watermark sat frozen at exactly `2026-07-28T18:11:36Z` for **~22.6 hours across 76 identical cycles** (07-28 19:22 → 07-29 17:58), ingesting nothing, until the 24h horizon-skip finally tripped.

This was found while running the infra#2106 capacity experiment, and it **falsifies the capacity framing**: more ClickHouse CPU cannot fix it (see below).

## Mechanism

Each cycle, in a 10-minute activity budget (`StartToClose: 10m`, `MaximumAttempts: 1`, from #711):

1. Catch-up window is **15 min** (#714's overlap-skip: `[maxTime, maxTime+15m)`) → ~6.9M InfluxDB rows.
2. `main influxdb query completed … rows=6880210 duration=6m29s` (also seen: `5m22s`) — the Flux query alone eats 50–65% of the budget.
3. The remainder is not enough for the link-lookup query + a ~4.6M-row ClickHouse batch insert, so the work dies on the deadline:
   - `failed to insert interface usage data to clickhouse: … context cancelled during batch insert: context deadline exceeded`
   - or `failed to query influxdb: … error iterating flux intf counters results: context deadline exceeded`
   - plus `activity error … activity StartToClose timeout` and `Activity complete after timeout`
4. **No insert → `max(event_ts)` unchanged → identical window next cycle.** No retry (`MaximumAttempts: 1`), and nothing shrinks the window on repeated failure.

Log evidence (staging, all 76 cycles identical):
```
17:52:49  capping catch-up window … queryStart=2026-07-28T18:11:36.582Z queryEnd=2026-07-28T18:26:36.582Z target=2026-07-29T17:52:42Z
17:58:15  capping catch-up window … queryStart=2026-07-28T18:11:36.582Z queryEnd=2026-07-28T18:26:36.582Z target=2026-07-29T17:58:08Z
```

## Why this is not a capacity problem

infra#2106's experiment raised staging ClickHouse 16→24 CPU / 32→48Gi (infra#2122, applied 2026-07-29 17:15Z). ClickHouse came up healthy and CH-side latency improved (baseline query 15.1s → 7.2s), but **the livelock persisted unchanged** — because the majority of the exhausted budget is the **InfluxDB** query, which the resize didn't touch. The binding constraint is *work-per-cycle vs. activity budget*, not CH cores.

## Interaction that created it

- **#711** gave the activity a dedicated 10-min `StartToClose` with `MaximumAttempts: 1`.
- **#714** widened catch-up windows from 10 → 15 min (overlap skip), raising per-cycle rows ~1.5×.
- Neither change is wrong alone; together, on any environment where a 15-min window costs >10 min, catch-up cannot make progress — and the failure is **silent** (no data loss alert; just a frozen watermark) until the 24h horizon fires ~a day later.

## Prod exposure (latent)

Prod currently fits: its 15-min-window Flux query measured ~2m0s with a ~50s insert (2026-07-25), well inside 10 min. But the livelock triggers precisely **during catch-up after an outage** — i.e. when prod is already degraded — and any ~2× InfluxDB slowdown in that state would freeze prod ingest for up to 24h before the horizon skip drops the span. Worth fixing before it's discovered in prod.

## Fix options

1. **Adaptive window**: on a deadline failure, halve the catch-up window for the next attempt (and recover upward on success). Directly breaks the livelock and self-tunes per environment.
2. **Make progress durable**: insert in sub-batches and advance the watermark for what was actually written, so a partial cycle still moves forward. (Also stops throwing away 5+ minutes of completed Flux work on every failed cycle.)
3. **Budget coherence**: derive the activity `StartToClose` from the window size (or cap the window from the budget) so the two can't drift apart, and assert the relationship in a test — the current values were chosen independently in #711 and #714.
4. **Detect the freeze**: alert when the watermark is unchanged across N consecutive cycles. 22.6 hours of zero ingest produced no distinct signal — the escalating errors looked like ordinary transient failures.

## Acceptance

- An induced slow environment (Flux query > budget) still makes forward progress per cycle, or shrinks its window until it does; no identical-window repetition.
- A frozen watermark is detectable from logs/metrics without waiting for the 24h horizon.

## Context

#708 → #711 → #713/#714 → #718/#719 (catch-up/honest-lag chain); infra#2106 (staging capacity, now partly falsified); infra#2122 (the resize experiment). Prod ingest is currently healthy (single-digit-minute staleness since 07-25).

nikw9944:
## Summary of Changes

* The telemetry-usage catch-up span is now adaptive (AIMD): a failed capped cycle halves it down to a 1m floor, each successful capped cycle adds 1m back up to the previous fixed 10m cap. Steady-state (uncapped) cycles never touch it, so a healthy environment behaves exactly as before.
* Fixes a livelock, not a slowdown. A capped cycle that failed after its window was computed re-queried the **identical** window forever, because `max(event_ts)` only advances on a successful insert and nothing about the window depended on the previous outcome. Staging sat frozen at `2026-07-28T18:11:36Z` for ~22.6h across 76 cycles, ingesting nothing. Prod fits today, but the trigger is catch-up after an outage — i.e. when prod is already degraded.
* Adds `doublezero_data_indexer_telemetry_usage_watermark_timestamp_seconds{dz_env}`, published before any expensive work so failing cycles report it too. Deliberately an absolute timestamp rather than a precomputed lag: `time() - <gauge>` keeps climbing even if refreshes stop running altogether, which a lag gauge would not.
* Ties the activity deadline to the work it bounds. `WorstCaseRefreshFluxBudget()` names the worst-case InfluxDB cost of one capped refresh (14m) and a test asserts `telemUsageStartToCloseTimeout` exceeds it. #711's 10m and #714's 15m-of-data were chosen independently; that is how they drifted into incoherence.
* Fixes #740

Deliberately out of scope, worth filing as follow-ups:

1. **Durable partial progress** — sub-batch inserts that advance the watermark for what was actually written, so a failed cycle stops discarding 5+ minutes of completed Flux work. Needs the sparse-counter baseline-cache interaction designed properly (`updateBaselineCache` must only run once rows are durably written, and deltas depend on ordered whole-window processing).
2. **A `deriv()`-based frozen-vs-converging detector plus a staging PodMonitor** — the pair that would give staging and devnet real coverage against their structurally-stale baselines, and the only thing that would flag the shrink walk-down as an incident. Worth designing after the new gauge has a few days of observed behavior.
3. **A scrape-health alert** for `lake-indexer` app metrics — closes the `noDataState: OK` gap that `lake-indexer-down` does not cover (a broken PodMonitor leaves the pod healthy and both alerts quiet).
4. **Derive the `lake-admin` InfluxDB bucket from `--dz-env`**, as the ClickHouse database already is. Today `--dz-env testnet` redirects the write target but not the read bucket, so a testnet usage backfill silently reads mainnet. The infra runbook works around it; the CLI should not need one.

## Testing Verification

* Six new tests cover each branch: shrink on a deadline failure, shrink on a non-deadline failure (the issue saw both shapes), forward progress under an induced slow source, additive-not-multiplicative recovery, no span change on an uncapped failure, and the repeated-window WARN gate.


indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -209,6 +213,12 @@ func (cfg *ViewConfig) Validate() error {
 		return fmt.Errorf("query window %s plus one capped span (%s) must stay below maxCatchupHorizon (%s)",
 			cfg.QueryWindow, cfg.QueryChunk*maxCatchupChunks, maxCatchupHorizon)
 	}
+	// A ceiling at or below minCatchupSpan silently disables the adaptive shrink
+	// (see adjustCatchupSpan) and restores the #740 livelock.
+	if ceiling := cfg.QueryChunk * maxCatchupChunks; ceiling <= minCatchupSpan {
+		return fmt.Errorf("query chunk %s is too small: the catch-up span ceiling (QueryChunk×maxCatchupChunks = %s) must exceed minCatchupSpan (%s) or the adaptive shrink cannot fire",
+			cfg.QueryChunk, ceiling, minCatchupSpan)
+	}
```
ben-dz, line 230: This guard doesn't guard the drift it exists for. `WorstCaseRefreshFluxBudget` hardcodes `defaultQueryChunk` (`view.go:348`), while this check only rejects `QueryChunk ≤ 30s` — the harmless direction, whose sole consequence is reverting to pre-#740 behavior.

The real hazard is a small-but-legal chunk. Sub-queries are `ceil((refreshOverlap + 2·QueryChunk)/QueryChunk)` (`view.go:1036`), so `QueryChunk = 4m` gives a 13m window in **4** sub-queries = 16m + 2m fallback = **18m against the 15m deadline** — and it passes all three `Validate` checks while `TestLake_DZIngest_TelemetryUsageBudgetCoversWorstCaseRefresh` stays green at 14m, because the helper never sees the configured value.

Fix: replace this check with `cfg.QueryChunk >= refreshOverlap`. That pins sub-queries at exactly 3 across the whole legal range (the `baselineCacheMaxLag` guard already caps the chunk below 7.5m), makes the helper's 14m correct for any accepted config, and subsumes the too-small case this check covers.

ben-dz (changes requested):
One high finding to resolve before merge, plus two mediums on the new guardrails.

The AIMD recovery rate ties the steady-state span to the capped-cycle **failure rate** rather than to what the environment can afford: with *k* successes between failures the span has a fixed point at exactly *k* minutes (`(x+k)/2 = x`). At the real cadence (`refreshInterval = 60s`, `telemUsageEveryN = 1`, future awaited in-iteration), a sustained ~50% failure rate pins the span at the 1m floor and catch-up **diverges** — 0.5m of watermark advance per cycle against ≥1.5m of wall clock — where the pre-PR fixed 10m span converged at 5m/cycle. That regime needs failures uncorrelated with span size, which is exactly what InfluxDB rate limiting looks like (already a named cause in `activities.go`), and catch-up issues 3× the steady-state query count. Nothing pages before `maxCatchupHorizon` permanently drops the data: the escalator resets on every success, and the shrink WARN, the divergence WARN, and the repeated-window gate are all WARN or unreachable in that state. Gating the shrink on `sameWindowCycles >= 2` fixes it using state the PR already tracks, without weakening the #740 property.

Separately, the new drift guard does not guard the drift it exists for (a legal `QueryChunk = 4m` yields an 18m worst case against the 15m deadline while the new assertion stays green at 14m), and the budget's premise that Flux queries ignore the activity deadline is contradicted by the client library.

- `indexer/pkg/dz/telemetry/usage/view.go:939` — high: **Additive recovery ties the steady-state span to the failure *rate*, not to what the environment can afford — at a sustained ~50% capped-cycle failure rate catch-up now diverges where it previously converged.**

Halving on failure and adding `minCatchupSpan` (1m) on success gives the span a fixed point at exactly *k* minutes, where *k* is the number of successful capped cycles between failures: after *k* successes the span is `x + k`, and the next failure halves it, so `(x + k)/2 = x` ⇒ `x = k`. The span settles at a value determined by the failure rate alone — not by the window cost the shrink is meant to be probing for.

`dzingest` runs this activity every iteration at `refreshInterval = 60s` (`workflow.go:13`, `telemUsageEveryN = 1`) and awaits the future in the same iteration, so cycle wall time is `60s + refresh duration`. Simulating 600 cycles at that cadence with cheap failures:

| capped-cycle failure rate | span pins at | adaptive | pre-PR fixed 10m |
|---|---|---|---|
| 1 in 2 | **1m (floor)** | **diverges** (0.5m advance/cycle vs ≥1.5m wall) | converges (5m/cycle) |
| 1 in 3 | 2m | converges | converges |
| 1 in 4 | 3m | converges | converges |
| 1 in 6 | 5m | converges | converges |

So the regression is confined to roughly a ≥50% failure rate on capped cycles — but that regime is reachable and it is not the #740 regime. It needs failures *uncorrelated with span size*, i.e. the 10m ceiling would have succeeded half the time: a rate-limiting InfluxDB is the obvious one, and this code already names it as a known cause (`activities.go:70`, `cause=influxdb_rate_limit`). Catch-up issues 3× the steady-state query count, so it is exactly when the org budget gets hit. There the pre-PR fixed span advanced 10m per successful cycle and closed the gap; the adaptive span walks to the 1m floor, advances 0.5m per cycle against 1.5m of real time, and diverges monotonically to `maxCatchupHorizon`, where the skip **permanently drops** the intervening data.

Nothing pages before that loss. `Activities.refresh` calls `esc.Reset` on every success (`activities.go:76`), so a 50% failure rate never reaches three consecutive failures and the escalator never escalates to ERROR. The shrink WARN is WARN. The divergence WARN does fire (a failed cycle returns before `lastCatchupLag` is touched, so consecutive successes do compare a growing lag) — also WARN. The repeated-window escalation cannot fire: the successful cycles advance the watermark, so the window changes every cycle and `sameWindowCycles` never leaves 1. The first alert-bearing line is the horizon ERROR, concurrent with the data loss.

**Fix:** gate the shrink on the condition that actually defines #740 — the identical window genuinely repeating — instead of on any single failure. The code already tracks it:

```go
case v.sameWindowCycles < 2:
    // One failure proves nothing: if the next cycle succeeds the watermark
    // advances and the window changes. Only a window that has actually
    // repeated is the #740 livelock.
case v.catchupSpan <= minCatchupSpan:
    ...
```

A failure followed by a success changes the window and resets the count, so the intermittent case keeps the ceiling span and its old convergence. A sustained failure still walks down, now two cycles per halving (~8 cycles to the floor instead of 4) — still bounded, still strictly cheaper than repeating identical work, which is the property the issue asks for. If you would rather keep the per-failure trigger, the increase has to be a fraction of the ceiling rather than a fixed 1m so the fixed point cannot land below what a cycle needs.

**Test:** a table-driven test directly on `adjustCatchupSpan` asserting the span after an `F,S,S,F,S,S,…` pattern stays at the ceiling. That needs no ClickHouse container, unlike the `Refresh`-level tests.
- `indexer/pkg/dz/telemetry/usage/view.go:341` — medium: The premise here is contradicted by the client library. `queryAPI.Query(ctx, …)` (`flux_influx_client.go:70`) reaches `influxdb-client-go/v2@v2.14.0/api/http/service.go:102`, which builds the request with `http.NewRequestWithContext(ctx, …)` — so both the Flux request and the streaming body read behind `result.Next()` **are** cancelled by the activity deadline. Total Flux time is bounded by the 15m deadline, not by 3×4m.

The `15m > 14m` assertion is therefore conservative (safe direction) but models a state that cannot occur, and its failure message claims "with margin for the ClickHouse insert" when the residual is 1m — smaller than the 60s `max_execution_time` the baseline scan alone can consume. Carried as prose this was harmless; encoded in a helper and a CI assertion it will be used to size the next deadline.

Fix: correct the "does not abort on the activity context deadline" claim, and either drop the margin wording from the assertion message or assert against `WorstCaseRefreshFluxBudget() + <named ClickHouse allowance>`.

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: Fixed in `cb84327` — replaced the check with `cfg.QueryChunk >= refreshOverlap`, as suggested.

Confirmed your numbers: `QueryChunk` of 3m or 4m both give 4 sub-queries = 18m against the 15m deadline while passing every other `Validate` check. The new guard pins sub-queries at exactly 3 across the whole remaining legal range (5m through just under 7m30s, where the `baselineCacheMaxLag` check takes over), so `WorstCaseRefreshFluxBudget`'s 14m is correct for any accepted config — and it subsumes the too-small case my original check covered, since a chunk ≥ 5m puts the ceiling at ≥ 10m.

Added a `Validate` subtest on the 4m case so CI catches the direction that actually hurts.

nikw9944:
All three findings addressed in `cb84327`, plus a rebase onto `main` (`24d9625`) resolving a CHANGELOG conflict against #742.

**High — additive recovery ties the span to the failure rate.** Fixed as you proposed: the shrink is now gated on `sameWindowCycles >= 2`. I reproduced the regression before changing anything — simulating 600 cycles at the real cadence (90s per cycle), a 1-in-2 failure rate advanced the watermark 0.51 m/cycle against 1.50m of wall clock, and the gate restores 5.00 m/cycle. Your `x = k` fixed point is exactly right, and I verified the two supporting claims rather than assuming them: `esc.Reset` on every success (`activities.go:76`) means a 50% failure rate never reaches three consecutive failures, so nothing pages, and `influxdb_rate_limit` is already a named cause at `activities.go:71` — so span-independent failures are a real mode, not a hypothetical.

The #740 property holds: a sustained failure still walks 10m → 5m → 2m30s → 1m15s → 1m, now at two failed cycles per halving, reaching the floor in 8 cycles instead of 4. Bounded, and each retry is still strictly cheaper than the one that failed. Staging's 76 consecutive identical windows would have tripped it on cycle 2.

I took the `sameWindowCycles` gate over your alternative of a fractional increase because the fixed point is `s·k` for step `s` — converging at `k = 1` needs `s > 3m`, which reintroduces the oscillation additive recovery exists to avoid. Gating the trigger fixes the cause; tuning the step only moves the fixed point.

**Medium — the Flux/ctx premise.** You're right, and it was wrong in the codebase before this PR too. Confirmed the chain: `queryAPI.Query(ctx, …)` → `DoPostRequest` → `doHTTPRequestWithURL` → `http.NewRequestWithContext` at `api/http/service.go:102`. Corrected the claim in `workflow.go` and reframed the invariant around what actually matters — the deadline *does* cancel Flux, which is precisely why it must exceed the budget: set below it, cancellation lands mid-query and the insert never runs. Dropped "with margin for the ClickHouse insert" from the assertion message; the 1m residual is stated plainly instead, including that it is tighter than the 60s `max_execution_time` the baseline scan alone can use.

**Medium — the drift guard.** Replied on the thread; `cfg.QueryChunk >= refreshOverlap` as suggested.

**Tests.** Added the table-driven test on `adjustCatchupSpan` you asked for, including an `F,S,S…` case over 30 cycles asserting the span holds at the ceiling. It needs no ClickHouse container — though the package's `TestMain` still starts one, so it does not run in this environment either. **No test in these two packages ran locally: there is no Docker daemon here.** Everything compiles (`go test -c`), vet and `golangci-lint` are clean, and I validated every changed expected value by replicating `adjustCatchupSpan` plus `Refresh`'s window tracking verbatim in a standalone program — the span sequences, the `sameWindowCycles` trajectory (≤ 2 through the walk-down, WARN at cycle 12), and the empty-source-stall case all match what the tests assert. CI is the real check.

Nothing else in your review left unaddressed.

indexer/pkg/dz/telemetry/usage/view.go
```diff
@@ -754,6 +876,68 @@ func (v *View) Refresh(ctx context.Context) (ingestionlog.RefreshResult, error)
...
+	ceiling := v.cfg.QueryChunk * maxCatchupChunks
+	switch {
+	case err == nil:
+		if v.catchupSpan >= ceiling {
+			return
+		}
+		v.catchupSpan = min(v.catchupSpan+minCatchupSpan, ceiling)
+	case errors.Is(err, context.Canceled):
+		// Pod shutdown says nothing about the cost of the window.
+	case v.sameWindowCycles < 2:
```
ben-dz: **This gate can never fire when the window is anchored at `now − QueryWindow`, so those cycles repeat the same cost forever with no shrink.**

Two branches anchor `queryStart` at `queryWindowStart` (`= now − QueryWindow`) rather than at the watermark: the initial full refresh when `maxTime == nil` (`view.go:617`) and the horizon-skip cycle (`view.go:568`). On both, `newDataStart` falls back to `queryStart` (`view.go:741`) and `queryEnd = newDataStart + catchupSpan`, so **both ends of the window slide forward with `now` every cycle**. `queryStart.Equal(v.lastWindowStart)` is false on every cycle, `sameWindowCycles` is pinned at 1, and this case matches forever — the shrink never fires, however many consecutive capped cycles fail.

That matters most on the most expensive cycle in the system. A fresh `dz_env` with an empty table has `maxTime == nil` and returns 0 ClickHouse baselines, so it takes the 2m InfluxDB baseline fallback (`view.go:648`) *on top of* the 3 chunked Flux reads — the full 14m `WorstCaseRefreshFluxBudget` against a 15m deadline, the path the code itself documents as tight. If that overruns, the cycle fails and nothing shrinks it: the same 10m-of-data window at the same cost is retried indefinitely. The window is not byte-identical each time, but the work is, which is the property this PR sets out to guarantee.

Not a regression against `main` (no shrink existed there), and not the #740 frozen-watermark case — that one the gate handles correctly, since `queryStart = watermark.Add(-refreshOverlap)` (`view.go:608`) is frozen while the watermark is, so the window truly repeats and the shrink fires on cycle 2.

**Fix:** trigger on consecutive capped failures rather than window identity. Track a `consecutiveCappedFailures` counter — increment on a capped failure, reset on any capped success — and gate on `>= 2`:

```go
case v.consecutiveCappedFailures < 2:
    // One failure proves nothing: a success advances the watermark and the
    // next window is cheaper anyway.
```

Cost determines whether a cycle fits and the span determines cost, so this predicate covers the frozen and sliding anchors identically. It preserves everything the current gate buys — an intervening success still resets it, so the intermittent-failure fixed point stays at the ceiling — and it drops the shrink's dependency on `sameWindowCycles`, which exists for the WARN and is maintained mid-`Refresh`. The surgical alternative is a `|| newDataStart.Equal(queryStart)` disjunct, but that bolts a second predicate onto a condition that was already the wrong one.

**Test:** the table test at `view_test.go:2564` covers this with one more row — two consecutive capped failures with `sameWindowCycles` held at 1 must shrink. Every adaptive-span test currently calls `seedMaxTime`, so the `maxTime == nil` path has no coverage at all.

ben-dz (commented):
All three round-1 findings verified fixed in `cb84327` — the shrink is gated on a repeated window, `Validate` now rejects a chunk below `refreshOverlap`, and the false Flux/ctx premise is corrected in both the helper doc and the assertion message.

One gap the new gate introduces: `sameWindowCycles >= 2` can never be satisfied when the window is anchored at `now − QueryWindow` rather than at the watermark — the initial full refresh (`maxTime == nil`) and the horizon-skip cycle. Both ends of that window slide with `now`, so the count is pinned at 1 and the shrink never fires, while the cycle repeats the same cost indefinitely. It bites hardest on the bootstrap of a fresh `dz_env`, which is the single most expensive cycle in the system (0 ClickHouse baselines → the 2m InfluxDB fallback on top of 3 chunked reads = the full 14m budget against a 15m deadline). Not a regression against `main`, but a gap in the property the PR sets out to guarantee. Triggering on consecutive capped failures instead of window identity closes it and keeps everything the current gate buys.

indexer/pkg/dz/telemetry/usage/view.go
nikw9944: Fixed in `b5e1640` — switched to `consecutiveCappedFailures >= 2` as suggested, and dropped the shrink's dependency on `sameWindowCycles` entirely (it now serves only the WARN).

Confirmed the gap before changing anything: on both branches `newDataStart` falls back to the sliding `queryStart`, so `queryStart.Equal(v.lastWindowStart)` is false every cycle and the old gate matched forever. Simulated all-failing cycles on the sliding anchor — the window-identity gate produced **0 shrinks over 20 cycles** (span stuck at 10m); the streak trigger walks `10m → 5m → 2m30s → 1m15s → 1m`. On the frozen-watermark #740 case both reach the floor with 4 halvings, and the round-1 `F,S,S` case still holds at the ceiling and converges at 5.00 m/cycle, since a success clears the streak.

Your predicate is also simpler in practice: after the first free failure every consecutive failure halves, so the walk-down is 5 cycles instead of 8.

Tests: added the sliding-anchor row you asked for (`sameWindow` held at 1, two consecutive failures, must shrink), a `wantStreak` column so the counter's own transitions are pinned, and a `Refresh`-level test on the `maxTime == nil` path — you were right that it had no coverage at all.

One residual, not code I'm adding: on the sliding anchor `sameWindowCycles` stays 1, so the repeated-window WARN can't fire there. Coverage on that path comes from the per-halving shrink WARN plus the activity escalator, which does reach ERROR because these failures are consecutive by construction.

nikw9944:
Round-2 finding fixed in `b5e1640`, plus a rebase onto `main` (`24d9625` → current, no conflicts this time).

**The sliding-anchor gap is real and I reproduced it.** Simulating all-failing capped cycles with the window anchored at `now − QueryWindow`, the `sameWindowCycles >= 2` gate produced **0 shrinks over 20 cycles** — span stuck at the 10m ceiling, exactly as you described. Switched the trigger to `consecutiveCappedFailures >= 2`: increment on a capped failure, clear on any success, `context.Canceled` neither counts nor shrinks.

Verified the swap keeps everything round 1 bought:

| scenario | window-identity gate | failure-streak trigger |
|---|---|---|
| #740 frozen watermark, all fail | floor, 4 halvings | floor, 4 halvings |
| sliding anchor, all fail | **0 shrinks** | floor, 4 halvings |
| `F,S` alternating (round-1 case) | ceiling, 5.00 m/cyc, converges | ceiling, 5.00 m/cyc, converges |

The streak trigger is also simpler in effect — after the first free failure every consecutive failure halves, so the walk-down is 5 cycles rather than 8 — and the shrink no longer reads `sameWindowCycles` at all, which now serves only the capping WARN as you noted it should.

**Tests.** Added your sliding-anchor row to the table test (`sameWindow` pinned at 1, two consecutive failures, must shrink), a `wantStreak` column pinning the counter's own transitions, and a `Refresh`-level test for `maxTime == nil` — correct that it had zero coverage, since every other adaptive-span test seeds a watermark.

**Two things I am not changing, stated rather than buried:**

1. On the sliding anchor `sameWindowCycles` stays 1, so the repeated-window WARN cannot fire there. That path is covered by the per-halving shrink WARN and by the activity escalator reaching ERROR — these failures are consecutive by construction, unlike the intermittent case.
2. Extending round 1's analysis: at random failure rates the divergence boundary sits near p ≈ 0.6 for **both** the window-identity gate and the streak trigger (p=0.5 converges at 2.3–2.6 m/cycle against 1.5m of wall clock; p=0.6 gives ~1.3 m/cycle). So the streak trigger is not a regression against the gate you approved, but neither fully immunizes AIMD-with-additive-recovery against a badly degraded source. Above p≈0.6 the escalator does page, since runs of three consecutive failures become common — which is the part of round 1's "nothing pages" concern that does not carry to that regime.

**Verification unchanged and still limited: no test in these two packages ran locally — there is no Docker daemon here**, and both `TestMain`s start a ClickHouse container. `go build`, `go vet`, `golangci-lint` (0 issues) and `go test -c` all pass. Every changed expected value was derived by replicating `adjustCatchupSpan` plus `Refresh`'s window tracking verbatim in a standalone program: the span sequences, the `sameWindowCycles` trajectory (floor at cycle 5, WARN at cycle 9), the sliding-anchor walk-down, and the empty-source-stall case all match the assertions. CI is the real check.

nikw9944:
Status: no code changes this round — CI is green and the branch is current with `main` (`8b4348f`, linear, no merge commits).

**The tests I could not run locally now pass in CI**, which closes the caveat I carried through both earlier rounds (no Docker daemon in my environment, and both packages start a ClickHouse testcontainer in `TestMain`). Under `-race`:

- `adjustCatchupSpan` — all 12 table subtests, including `shrinks_on_a_sliding_anchor,_where_the_window_never_repeats` and `first_capped_failure_does_not_shrink`
- `adjustCatchupSpan_IntermittentFailureHoldsCeiling` — the round-1 `F,S,S` fixed-point case
- `Refresh_InitialRefreshShrinksDespiteSlidingWindow` — the round-2 `maxTime == nil` gap
- `Refresh_FailedCappedCycleShrinksWindow`, `_NonDeadlineFailureShrinksWindow`, `_InducedSlowSourceMakesProgress`, `_SuccessAdditivelyRestoresCatchupSpan`, `_UncappedFailureKeepsCatchupSpan`
- `Refresh_RepeatedWindowWarnsOnlyAtFloor` — both subtests
- `TestLake_DZIngest_TelemetryUsageBudgetCoversWorstCaseRefresh`

Every span sequence I had derived by simulation matches real execution, so the expected values in those tests are now verified rather than reasoned.

@ben-dz — both findings from your second pass are fixed (`b5e1640`) and replied to on their threads. Merge is currently gated on two things only you or a maintainer can clear: the ruleset on `main` requires one approving review, and your round-1 `CHANGES_REQUESTED` still stands formally even though your follow-up reviews confirmed the fixes. It also requires all review threads resolved — I'm leaving both threads for you to resolve rather than closing them myself.

ben-dz (approved):
lgtm — the shrink trigger is now `consecutiveCappedFailures >= 2`, which fires on both the frozen-watermark and sliding-anchor windows while a success still clears the streak, and the previously uncovered `maxTime == nil` path has a `Refresh`-level test. All findings from both prior rounds verified fixed in code.

PR #749, opened 2026-07-30
api/handlers: fold multicast-delivery totals into their page queries

### LINKED ISSUE #748: api/handlers: device multicast-delivery tests run close to the 30s handler deadline
The device multicast-delivery handler tests are passing close enough to their deadline that they are likely to start flaking.

Observed in the `go-test` job of https://github.com/malbeclabs/lake/actions/runs/30568486876 (branch `bdz/adhoc-100`), both passing:

- `TestDeviceMulticastDelivery_HealthFiltersUseHealthSemantics` — 34.56s
- `TestDeviceMulticastDelivery_EmptyState` — 26.06s

Those are whole-test wall times including per-test ClickHouse setup, so the handler itself stayed under its 30s budget. But they sit right at the edge, and the equivalent link test already crossed it: it failed with a 500 after 31.93s earlier on the same branch, which is what prompted the fan-out in #744 (that one now passes in 15.86s).

Two things make this worse than a slow test:

1. `GetDeviceMulticastDelivery` sets `context.WithTimeout(r.Context(), 30*time.Second)` (`api/handlers/multicast_delivery_entity.go:93`) while its queries set `max_execution_time = 30`. The client deadline wins that race, so an overrun surfaces as `context.DeadlineExceeded` rather than a ClickHouse error naming the query. `logError` classifies that as client-disconnect and drops it, so a production occurrence returns a 500 with **no log line at all**. Lowering the SQL cap to ~20s would make ClickHouse fail first and produce a diagnosable error. Same pattern at `:347` for the link handler.
2. `countMulticastLinkDeliveryBranches` is called inline at the top of `queryMulticastLinkDeliveryBranches` (`api/handlers/multicast_delivery_entity_link_queries.go:14`), so two full `enriched_ip_mroute_oifs` scans still run back-to-back inside a single fan-out slot. Spawning the count as its own branch would take that side of the fan-out to one query deep.

Worth checking whether the device handler's remaining depth has the same shape, and whether the CI runner's `-p 16` package parallelism (16 ClickHouse containers on 16 cores) means these tests are measuring contention as much as query cost.

Not urgent — nothing is red. Filing so it is not rediscovered the next time CI goes red on a 500 with no log line.

ben-dz:
## Summary of Changes

* Each paged multicast-delivery section now carries its own total as a window aggregate (`count() OVER ()`, and `countIf(health_status = '…') OVER ()` per bucket for the health legs), instead of running a separate `count()` over the same view and filter first. Five full scans of `enriched_ip_mroute` / `enriched_ip_mroute_oifs` / the health tables disappear per request, and every fan-out branch is now one query deep.
* Per-query `max_execution_time` drops 30s → 20s, below the handler's own 30s deadline. Previously the client deadline won that race, so an overrun arrived as `context.DeadlineExceeded`, which `logError` classified as a client disconnect and dropped — a 500 with **no log line at all**. Now ClickHouse fails first with `TIMEOUT_EXCEEDED` naming the query.
* That classification is also fixed generally in `api/handlers`: net/http never sets a deadline on `r.Context()`, so a request-path deadline is the handler's own budget expiring and logs at WARN rather than being discarded. Cancellations and broken connections are still dropped.
* CI `go test` drops to `-p 6`: ~20 packages start their own database container, and the GOMAXPROCS default let 16 of them size for the whole 16-core runner — the contention that pushed these tests toward their deadline in the first place.
* Fixes #748

## Testing Verification

* Window semantics validated against production ClickHouse: `count() OVER ()` on the busiest device's real OIF filter returns exactly the same total as the standalone `count()` (43,683 rows), the link query's window total is taken after the direction filter, and `countIf(...) OVER ()` matches the `GROUP BY health_status` buckets.
* New test covers the one semantic edge the fold introduces: a page whose offset lands past the end carries no total, so those legs fall back to the count query. The route and endpoint-health pagers are moved past the end independently, since they page on different offsets.
* New unit tests pin the query cap below the request deadline (including the fallback's remaining-deadline budget), that the health buckets match `addEntityStatusCount` with `unknown` as the remainder, and that a handler deadline now produces a WARN line.
* Not verified locally: the container-backed `api/handlers` tests and the `-p 6` effect on total CI wall time. Compare the `go-test` step against the baseline in #748 (`_EmptyState` 26.06s, `_HealthFilters…` 34.56s, whole step 331s).


api/handlers/multicast_delivery_entity.go
```diff
@@ -31,6 +31,47 @@ const multicastDeliveryEntityCoverageNote = "Current observed route state is lim
...
+const (
+	// multicastDeliveryRequestTimeout bounds a whole entity request.
+	multicastDeliveryRequestTimeout = 30 * time.Second
+
+	// multicastDeliveryQueryTimeoutSeconds must stay strictly below
+	// multicastDeliveryRequestTimeout so an overrun surfaces as a ClickHouse
+	// TIMEOUT_EXCEEDED naming the query (logged at WARN, transient) rather than
+	// a bare context.DeadlineExceeded from the handler's own deadline winning
+	// the race. See TestMulticastDeliveryQueryTimeoutLosesToRequestDeadline.
+	multicastDeliveryQueryTimeoutSeconds = 20
```
nikw9944, line 43: The cap is a fixed 20s and ClickHouse starts counting it when the query *starts executing* — but `multicastDeliveryFanout.spawn` (`multicast_delivery_entity.go:95-109`) first blocks on a per-env semaphore sized `maxOpenConns * 6 / 10` (6 slots on devnet/testnet), stored on the `API` (`api/handlers/api.go:105`) and therefore shared across all concurrent multicast-delivery requests in the env. The device handler alone spawns ~9 stage-2 queries against those 6 slots.

A query that waits 12s for a slot then gets a 20s cap outlives the 30s request budget, so the handler deadline wins after all and the caller gets the bare `context.DeadlineExceeded` naming no query instead of the `TIMEOUT_EXCEEDED` this change is buying. The line is no longer dropped, but *which* query overran — the diagnostic the PR is after — is still lost, and queueing is likeliest under exactly the loaded conditions that motivated #748.

The fix is code already in this PR: make the remaining-deadline computation the default rather than fallback-only. Replace the `multicastDeliveryQuerySettings` package var with `multicastDeliveryQuerySettings(ctx)` carrying the current body of `multicastDeliveryFallbackQuerySettings`, thread `ctx` at the call sites (all are already methods with `ctx` in scope), and drop the separate fallback helper. Two helpers collapse into one and the cap becomes correct for queued and second-in-branch queries alike.

.github/workflows/go.yml
```diff
@@ -49,7 +49,13 @@ jobs:
         working-directory: dev/controlcenter/ui
       - run: bun run build
         working-directory: dev/controlcenter/ui
-      - run: go test -race -v ./...
+      # -p bounds how many packages run at once. ~20 packages start their own
+      # database container, so the GOMAXPROCS default (16 on this runner) had 16
+      # ClickHouse/Postgres/Neo4j servers each sizing thread pools and caches for
+      # the whole machine, which is what pushed container-backed tests toward
+      # their handler deadlines. Fewer packages at once means more CPU per
+      # container; watch total step time here against the pre-change baseline.
```
nikw9944: "watch total step time here against the pre-change baseline" reads as a standing instruction, so the next reader re-runs the comparison. This PR's own run already answers it: the `go test` step took 390s here (run 30578854767) against 402s on main (run 30478997187) — the wall-time cost of `-p 6` is nil. Replace the "watch" clause with the measured result.

nikw9944 (changes requested):
One gap to close before merge: `logWarn` was left on the old classification, so the "500 with no log line" failure this PR sets out to kill is still live on `GetMulticastGroupMembers` — an endpoint whose only timeout bound *is* the handler's own budget. The window-aggregate fold itself checks out: all five paged legs guard on the offset their `OFFSET ?` actually binds, `limit=0` is unreachable so an empty page can't fake a zero total, the link-branch window sits after the direction filter with filters and arg order matching the count query it replaces, and `unknown`-as-remainder is pinned against `addEntityStatusCount`. One medium: the new per-query cap is measured from query start, so a query that queues on the fan-out semaphore can still lose the race to the request deadline — the remaining-deadline helper this PR already wrote fixes it if made the default rather than fallback-only.

- `api/handlers/errors.go:41` — high: `logWarn` still delegates straight to `logger.Warn`, which skips on `logger.IsClientDisconnect` (`utils/pkg/logger/errlog.go:46-52`) — and that predicate still matches deadlines. So the fix landed in `logError` and `internalError` but not in the third helper in this file, and the classification is not yet fixed "generally in `api/handlers`".

Live reproduction: `GetMulticastGroupMembers` runs under its own 20s budget (`api/handlers/multicast.go:278`), and neither members query carries a `max_execution_time` setting, so that handler budget is the *only* bound. On expiry the driver returns a bare `context.DeadlineExceeded`, `logWarn` at `multicast.go:506` (and `:511`) returns without logging, and the handler still writes a 500 — HTTP 500 with zero log lines at any level, exactly the shape `isCallerGone` was introduced to prevent. Same shape at `multicast.go:660`/`:678` and at `swap_rate.go:64`, where the served-stale fallback's likeliest cause is the oracle client's own 10s timeout (`swap_rate.go:39`); CLAUDE.md wants that WARN line, and today it never appears.

Fix — same guard, leaving `logger.Warn` untouched for its non-handler callers:

```go
func logWarn(msg string, args ...any) {
	if err := logger.ErrorFromArgs(args); err != nil && isCallerGone(err) {
		return
	}
	slog.Default().Warn(msg, args...)
}
```

Related, same root cause: `logger.Error`'s doc comment at `utils/pkg/logger/errlog.go:29-33` still tells request handlers to "check `IsClientDisconnect` first (see `api/handlers.logError`)". `logError` no longer does that, and a future caller in this shared package who follows the comment reintroduces the drop — point it at `isCallerGone` / `IsDeadlineExceeded`.

api/handlers/multicast_delivery_entity.go
ben-dz: Fixed in f824c27. `multicastDeliveryQuerySettings` now takes `ctx` and carries the remaining-deadline computation; the separate fallback helper is gone, so queued and second-in-branch queries get the same treatment. Every call site is inside the spawned closure, i.e. after the semaphore acquire, so the cap is measured from the point the query actually starts.

`TestMulticastDeliveryFallbackQuerySettingsUseRemainingDeadline` became `TestMulticastDeliveryQuerySettingsUseRemainingDeadline` and now asserts the general case.

.github/workflows/go.yml
ben-dz: Fixed in f824c27 — comment now states the measured result (390s at `-p 6` vs 402s at the default) instead of asking the next reader to re-run it. Confirmed both step timings from the run IDs you cited.

ben-dz:
All three addressed in f824c27.

**`logWarn` (high).** You're right that the fix wasn't general — `logWarn` went straight to `logger.Warn`, which drops on `IsClientDisconnect`, so `GetMulticastGroupMembers` (whose only bound *is* the handler budget, no `max_execution_time` on either query) kept returning a silent 500. It now uses the same `isCallerGone` guard, so all three helpers in `errors.go` agree, and `TestLogWarn_KeepsHandlerDeadline` pins it: deadline → one WARN, cancel → nothing. `logger.Warn` is untouched for its non-handler callers.

**Query cap vs semaphore queueing (medium).** Took your suggestion — `multicastDeliveryQuerySettings(ctx)` now carries the remaining-deadline computation and the fallback-only helper is gone, so two helpers collapse into one. Worth noting for anyone reading later: the fix works because every call site builds its query inside the spawned closure, i.e. after the slot is acquired, so the cap is computed from the time actually left.

**`logger.Error` doc comment.** Repointed at `isCallerGone` / `IsDeadlineExceeded`, with the reason spelled out, so a future caller in the shared package doesn't reintroduce the drop by following it.

**Workflow comment.** Replaced the standing "watch this" instruction with your measured numbers (390s vs 402s); I re-derived both from the step timestamps in the runs you cited.

nikw9944 (commented):
All three findings from the previous round are fixed in `f824c27`, verified in the code: `logWarn` now uses the same `isCallerGone` guard as `logError` (with `TestLogWarn_KeepsHandlerDeadline` covering both the deadline-kept and cancel-dropped halves); `multicastDeliveryQuerySettings` takes `ctx` and caps every query at the remaining request budget, with all 13 call sites converted and the clause built inside the `spawn` closure so the cap really is measured after the semaphore acquire; and the `-p 6` comment states the measured 390s-vs-402s result.

One new issue that commit introduces: moving `api/handlers.logWarn` off `logger.Warn` left `logger.Warn` with zero production callers repo-wide, still gating on `IsClientDisconnect` as a whole and still recommending itself "for request paths" — four lines below the `logger.Error` doc comment this same commit updated to warn against exactly that. Deleting it (nothing calls it) or giving it the `&& !IsDeadlineExceeded` carve-out would stop the next request-path caller re-creating the silent 500.

- **`utils/pkg/logger/errlog.go:49` (medium)** — `logger.Warn` is now caller-less and still documents the pattern this PR removed.

`f824c27` moved `api/handlers.logWarn` off `logger.Warn`, which was the right fix. But `logger.Warn` was its only production consumer repo-wide, and the helper itself was left untouched. It is now an exported helper in a shared package with zero production callers (only its own test at `errlog_test.go:65,68`), still gating on `IsClientDisconnect(err)` as a whole:

```go
// Warn logs at WARN level, silently skipping client disconnects. Meant for
// request paths, where a disconnect-class error means the client is gone and
// there is nothing to log; elsewhere prefer a plain log.Warn.
func Warn(log Logger, msg string, args ...any) {
	if err := ErrorFromArgs(args); err != nil && IsClientDisconnect(err) {
		return
	}
	log.Warn(msg, args...)
}
```

Four lines above it, the `logger.Error` doc comment updated in this same commit (`errlog.go:33-37`) now says the opposite: request handlers "should check first — but on the cancel/broken-connection half only, not on `IsClientDisconnect` as a whole: a deadline there is the handler's own budget expiring, and dropping it leaves a 500 with no log line at any level."

So the file now holds guidance against the drop and, immediately below it, an exported helper whose doc explicitly recommends itself "for request paths" while implementing that exact drop. `logger.Warn` is the obvious name to reach for from a request path, and this package is shared with the indexer, slack, and telemetry callers — the next handler that uses it re-creates the silent 500 the last two commits removed.

Fix: delete `logger.Warn` and its test; nothing calls it, and `api/handlers.logWarn` is now the working pattern. If it is being kept deliberately for future callers, give it the same carve-out (`IsClientDisconnect(err) && !IsDeadlineExceeded(err)`) so it matches `isCallerGone`, and drop the "Meant for request paths" recommendation from a helper that has none.

ben-dz:
Fixed in ed67256 — took the delete option. Confirmed `logger.Warn` had no production callers repo-wide (the one `logger.Warn(` grep hit is `dev/controlcenter`'s own slog field, not this helper) and only its own test in-package, so it and `TestWarn_SkipsClientDisconnect` are gone. `api/handlers.logWarn` is the working pattern, and the trap you describe — the obvious name to reach for from a handler, documented "for request paths", implementing the drop — no longer exists to be reached for. The now-stale `logger.Warn` mention in `logWarn`'s own comment went with it.

The `Warn` method on the `Logger` interface stays: `logger.Error` and `Escalator.Fail` both use it to emit their downgraded lines.

nikw9944 (approved):
lgtm — no changes since the previous round (head is still `ed67256`) and every earlier finding stays verified-fixed, with CI green on that head. Main has advanced to `86c2d43`, so this needs a rebase before landing; I checked that commit for interaction and found none — it adds no reference to the `logger.Warn` this branch deletes, and the two `logWarn` sites it does add carry no `error` argument, so the new `isCallerGone` guard never applies to them.

PR #754, opened 2026-08-06
indexer: stop a transient blip permanently losing escrow events

bgm-malbeclabs:
## Summary

- A transient `getTransaction` failure in the escrow-events indexer **permanently lost** an escrow event, and reported the refresh as successful. This is silent data loss on a financial audit trail.
- `fetchEscrowEvents` skipped the failed signature and committed the rest. But `GetHighWaterMarks` derives the resume point from `max(slot)` of the rows *actually written* — and signatures arrive newest-first, so committing the skipped signature's newer siblings dragged the high-water mark past it. The next refresh passed that mark as `Until` and never fetched it again.
- A transient failure now discards the escrow's batch, leaving the high-water mark below the affected signature so the next refresh re-fetches everything. Inserts are idempotent via the `(escrow_pk, slot, tx_signature, event_type)` ReplacingMergeTree key.
- Non-transient failures still skip. An unretrievable pruned transaction cannot succeed on retry and would wedge the escrow forever.

`permissionevents` documents this exact hazard and deliberately guards against it by discarding a partially-decoded chunk:

> Decode order within a chunk is not cursor order: persisting a partially decoded chunk could advance the fact-derived high-water mark past never-decoded signatures — a silent, permanent gap.

Same risk, opposite decision. This aligns them.

## Why discard the batch rather than chunk oldest-first

`permissionevents` reverses to oldest-first and commits per chunk because it grinds a real multi-hour backlog. Escrow histories are small and steady-state is near-empty, so re-doing a batch next cycle costs nothing and needs none of that machinery. The caller's existing per-escrow tolerance (`return nil // Don't fail the entire refresh for one escrow`) means one throttled escrow still doesn't fail the refresh.

This also covers `BackfillRefresh`, which routes through the same function — a blip mid-backfill was stranding signatures there too.

## Testing Verification

- New test reproduces the loss: with the middle of three signatures throttled once, the pre-change code indexes slots `{100, 300}` and slot 200 never returns even after the throttle lifts. Post-change all three land.
- Adds the package's first view-level test infrastructure (`main_test.go`, `view_test.go`) — previously only `parser_test.go` existed, so the refresh path had no coverage at all.
- `go test -race` green across `dz/shreds/...`.

## Found by

An audit of every batch path in the indexer for the invariant *item-failure handling must match how the resume point is derived*. Two other paths fail it and are getting their own PRs: the rollup backfill chunk loop (permanent gaps), and `telemetry/latency` silent per-circuit drops (observability only).


indexer/pkg/dz/shreds/escrowevents/view.go
```diff
@@ -337,8 +338,25 @@ func (v *View) fetchEscrowEvents(ctx context.Context, escrow EscrowInfo, hwm Hig
...
+			// Until and never fetches it again, permanently losing an escrow event
+			// while reporting success. Discard the whole batch instead: the mark stays
+			// put, and the next refresh re-fetches every signature (inserts are
+			// idempotent via the ReplacingMergeTree unique key). Escrow histories are
+			// small and steady-state is near-empty, so re-doing a batch is cheap.
+			//
+			// Non-transient failures (notably an unretrievable pruned transaction)
+			// still skip: they cannot succeed on retry and would wedge the escrow
+			// forever.
+			if dberror.IsTransient(err) {
```
ben-dz: The transient gate doesn't recognize the errors a Solana RPC client actually returns, so the permanent gap this PR targets is still open for the common failure modes. `dberror.IsTransient` is a message-substring classifier built for ClickHouse/Neo4j/Influx/S3 errors; running the real `solana-go v1.21.0` constructors through it, `rpc.ErrNotFound`, `jsonrpc.HTTPError{429/502/503}`, and `RPCError{-32005 node behind / -32004 block not available / -32603 internal}` all return false and fall through to the skip at line 357. (`HTTPError`'s message reads `status code: 429`; the pattern in `dberror.go:167` is the literal `status 429`. `RPCError.Error()` is a `spew.Sdump`, so it degenerates to substring-matching provider prose.) Only socket-level faults and the exact string the new test injects pass.

`rpc.ErrNotFound` also makes the comment above wrong: `getTransaction` returns it for any null result, and `permissionevents/view.go:40-47` documents a near-tip null as a load-balanced backend lagging finalization that *must* retry — not pruned history. That sibling distinguishes the two by slot age, and its polarity is the inverse of this one: fail the chunk on any error, with a single narrow escape hatch.

Mirror it. `allSigs` is newest-first, so `allSigs[0].Slot` is the tip:

```go
if err != nil {
    if errors.Is(err, rpc.ErrNotFound) && sig.Slot < skipNotFoundBelow {
        metrics.EscrowEventsSkippedTx.Inc()
        v.log.Warn("shreds/escrow-events: transaction unretrievable upstream, skipping",
            "signature", sig.Signature.String(), "slot", sig.Slot, "error", err)
        continue
    }
    return nil, fmt.Errorf("get transaction %s: %w", sig.Signature.String(), err)
}
```

Hoist `notFoundSkipSlotLag` next to `sol.TxHistoryRPC` rather than redeclaring it. Land this together with the partial-progress fix — fail-closed alone is not safe on the largest escrow.

indexer/pkg/dz/shreds/escrowevents/view_test.go
```diff
@@ -0,0 +1,199 @@
...
+func TestLake_EscrowEvents_View_TransientFetchFailureMustNotStrandSignature(t *testing.T) {
+	t.Parallel()
+
+	rpc := &fakeViewRPC{
+		history: newViewHistory([]uint64{100, 200, 300}),
+		txErrs:  map[solana.Signature]error{},
+	}
+	// Throttle the MIDDLE signature: slot 300 still commits and would drag the
+	// fact-derived high-water mark past slot 200.
+	rpc.txErrs[mkViewSig(200)] = &soljsonrpc.RPCError{
```
ben-dz: The injected error was chosen to match a `dberror` substring rather than what the RPC client returns, so the test passes with either polarity of the gate for every real failure shape. Table-drive `txErrs` over `rpc.ErrNotFound` (near tip must converge; far below tip must skip and let the batch commit) and `jsonrpc.NewHTTPError(503, …)` (must converge) — each of those fails against the current diff. Nothing yet covers the surviving skip path, `BackfillRefresh`/`SkipHighWaterMarks`, or the pagination loop at `view.go:298-322`.

indexer/pkg/dz/shreds/escrowevents/view.go
```diff
@@ -337,8 +338,25 @@ func (v *View) fetchEscrowEvents(ctx context.Context, escrow EscrowInfo, hwm Hig
...
+				return nil, fmt.Errorf("get transaction %s: %w", sig.Signature.String(), err)
+			}
+			v.log.Warn("shreds/escrow-events: failed to fetch transaction, skipping",
```
ben-dz: On shutdown the loop emits one "skipping" WARN per remaining signature, producing thousands of lines that read like data loss. `IsTransient` deliberately returns false for `context.Canceled`/`DeadlineExceeded` (`dberror.go:71-73`), so a canceled context or an expired Temporal `StartToClose` walks every remaining signature against a dead context instead of aborting. Add `if ctx.Err() != nil { return nil, ctx.Err() }` at the top of the error branch; the fail-closed rewrite above also subsumes this.

ben-dz (changes requested):
The diagnosis is right and the regression test genuinely pins the bug, but the fix does not fire for the failures that actually happen in production, and the all-or-nothing discard introduces a new wedge on the one escrow that isn't small.

**The gate misses every application-level RPC error.** `dberror.IsTransient` is a message-substring classifier written for ClickHouse/Neo4j/Influx/S3. Running the real `solana-go v1.21.0` constructors through it: `rpc.ErrNotFound`, `jsonrpc.HTTPError{429/502/503}`, and `RPCError{-32005/-32004/-32603}` all classify as **not** transient and still take the skip path — the permanent gap. Only socket-level faults and the exact string the new test injects survive the gate. `rpc.ErrNotFound` is the sharp one: `permissionevents/view.go:40-47` documents a near-tip `getTransaction` null as a lagging backend that *must* retry, which contradicts the new comment's claim that non-transient means pruned history.

**Fail-closed without partial progress wedges the largest escrow.** The largest escrow has 4,471 distinct signatures (growing ~40/day, third-party-appendable, uncapped) fetched serially. On the no-high-water-mark path — `BackfillRefresh` and any cold start — one blip discards all 4,471 and the retry restarts from zero, so it never converges. This gets worse once the gate is fixed. `permissionevents` chunks oldest-first at 200 with a durable cursor precisely for this.

**Neither outcome is observable.** The discard is swallowed per-escrow, so the refresh reports success, the escalator resets, and `SourceMaxEventTS` is restamped. The truncating backfill workflow can empty the table, write nothing for an escrow, and log "backfill complete".

- `indexer/pkg/dz/shreds/escrowevents/view.go:336` — high: Discarding the whole batch with no checkpoint means the largest escrow may never finish indexing on a cold start. This loop is serial (only escrows are concurrent), and on the no-high-water-mark path — `BackfillRefresh` or any run against an empty table — it walks the escrow's entire history in one all-or-nothing pass.

The comment's premise doesn't hold: there are 691 escrows in prod, and the largest (`ExeAy1LJ…ves9F`) has 4,471 distinct signatures, growing ~40/day with no bound (a third party can append cheap transactions referencing an escrow, and nothing caps `allSigs` at line 318). One blip anywhere in 4,471 serial calls throws away the entire walk; the retry restarts from zero and trips again. This gets strictly worse once the gate above is corrected, since far more error shapes will take the discard path.

`permissionevents` reverses to oldest-first, chunks at `scanChunkSize = 200`, and commits a durable cursor per chunk (`permissionevents/view.go:363-370, 482-529`) for exactly this reason. Process `allSigs` oldest-first and insert per chunk so an aborted batch still advances the fact-derived mark over a real prefix — that gets most of the benefit without a cursor table. Test: a fake with more than one chunk of history failing mid-walk, asserting the second attempt does not restart from zero.
- `indexer/pkg/dz/shreds/escrowevents/view.go:215` — high: An escrow that keeps failing now stalls indefinitely with every success signal intact, so nobody finds out. The discard is swallowed here (`return nil`), which means the refresh returns success, `ViewRefreshTotal{success}` increments, the escalator resets, and `SourceMaxEventTS` is restamped with `time.Now()` — freshness never regresses even for an escrow stuck for days. The skip path has no counter either, unlike `PermissionEventsSkippedTx` (`indexer/pkg/metrics/metrics.go:50`). The PR's whole recovery story is "the next refresh re-fetches everything", and there is no signal for when that never happens.

The sharpest case is `BackfillEscrowEventsWorkflow`: `Truncate: true` empties the fact table (`indexer/pkg/dzingest/backfill_escrow_events.go:56`), then `BackfillRefresh` runs with no mark. One blip discards that escrow's entire history, the activity returns nil, `MaximumAttempts: 2` never fires, and the workflow logs "escrow events backfill complete" over a table missing an escrow. The periodic ingest workflow eventually re-derives the absent mark and retries, which bounds the damage — but it retries the same all-or-nothing walk, and the operator was told it succeeded.

Add an `EscrowEventsSkippedTx` counter alongside `PermissionEventsSkippedTx`; have `fetchEscrowEvents` report discarded escrows back to `refresh` so a persistently stuck one escalates instead of warning forever; and when `ignoreCursor` is set, return an error so the backfill activity fails and retries rather than reporting complete.

bgm-malbeclabs:
Pushed `104ab5e`. Both parts of the review hold up, and I checked each one before rewriting rather than taking them on trust.

**The gate did not fire.** I ran the real solana-go v1.21.0 error shapes through `dberror.IsTransient`:

```
rpc.ErrNotFound              IsTransient=false
HTTPError{429}               IsTransient=false
HTTPError{503}               IsTransient=false
RPCError{-32005 node behind} IsTransient=false
RPCError{-32603 internal}    IsTransient=false
RPCError{429 too many}       IsTransient=true    <- the only one
wrapped ErrNotFound          IsTransient=false
```

One shape out of seven, and it is the shape my own test injected. `IsTransient` is a message-substring classifier written for ClickHouse, Neo4j, Influx and S3. It was never going to read Solana errors.

**"Escrow histories are small" was wrong.** I asserted it from the median without checking the tail:

```
escrows: 693 | median sigs: 4 | p99: 87 | max: 4471 | escrows over one chunk: 2
```

Two escrows exceed a chunk today and the largest grows without bound. All-or-nothing discard means one blip on that escrow throws away the whole walk, and the retry starts from zero. It never converges on a cold start or a `BackfillRefresh`.

**What the rewrite does.** `fetchEscrowEvents` becomes `drainEscrow`, with the same polarity as permissionevents:

- Any fetch error fails its chunk. One escape: `rpc.ErrNotFound` more than `notFoundSkipSlotLag` (300 slots) below the tip is pruned history, which no retry recovers and which would wedge that escrow forever. Near the tip it is a lagging backend and must be retried. That also corrects the old comment, which claimed non-transient meant pruned history.
- The walk reverses to oldest-first and commits per chunk of `scanChunkSize` (200). The resume point is `max(slot)` of rows written, so oldest-first is the only order where a partial commit is a real advance. Newest-first would write a high watermark over a hole.
- `ctx.Err()` aborts instead of skipping.
- An `EscrowEventsSkippedTx` metric counts the escape, one summary WARN per refresh names the failed escrows, and `SourceMaxEventTS` is restamped only on a clean sweep.

**Tests.** Three replace the old single-shape test. Against the reconstructed original fix, 5 of the 6 error shapes in the new table fail — every one except `RPCError{429}`.

`golangci-lint` clean, `go test -race ./indexer/pkg/dz/shreds/...` green.

indexer/pkg/dz/shreds/escrowevents/view.go
```diff
@@ -231,21 +251,32 @@ func (v *View) refresh(ctx context.Context, ignoreCursor bool) (ingestionlog.Ref
...
-	}
+	result.RowsAffected = totalInserted
 
-	result.RowsAffected = int64(len(allEvents))
-	fetchedAt := time.Now().UTC()
-	result.SourceMaxEventTS = &fetchedAt
+	// One summary line per refresh rather than one per escrow, so a partial run is
+	// visible without burying the log. A failed escrow keeps its committed chunks and
+	// resumes next refresh, so the count matters more than any single error.
+	if n := failedEscrows; n > 0 {
```
ben-dz, line 259: A permanently stuck escrow still never escalates past WARN, so nothing pages when one stops advancing for good. `refresh` returns nil on a partial run, so `v.esc.Observe` at `view.go:166` sees success and resets the escalator; an escrow that fails on its first signature every cycle makes zero progress while emitting one WARN and one `partial` increment forever. Feed the consecutive-partial count through the existing `logger.Escalator` so a single blip stays at WARN and a sustained stall reaches ERROR.

indexer/pkg/dz/shreds/escrowevents/view.go
```diff
@@ -321,65 +352,114 @@ func (v *View) fetchEscrowEvents(ctx context.Context, escrow EscrowInfo, hwm Hig
...
+			txResult, err := v.cfg.RPC.GetTransaction(ctx, sig.Signature, &rpc.GetTransactionOpts{
+				MaxSupportedTransactionVersion: &maxVersion,
+			})
+			if err != nil {
+				// One narrow escape hatch: a finalized signature the RPC cannot serve
+				// from far enough below the tip is pruned history. Retrying can never
+				// recover it and failing forever would wedge the escrow, so it is
+				// skipped and counted. Every other error fails the chunk.
+				//
+				// Failing closed is what closes the gap. The previous version gated on
```
ben-dz, line 403: These six lines describe how the deleted version behaved, inside code that no longer calls `dberror.IsTransient`. A reader who greps for `IsTransient` after reading this finds nothing and has to work out which version is being described. Delete the paragraph — the four lines above it state the invariant on their own; the same applies to the `view_test.go:161-171` header.

indexer/pkg/dz/shreds/escrowevents/view.go
```diff
@@ -321,65 +352,114 @@ func (v *View) fetchEscrowEvents(ctx context.Context, escrow EscrowInfo, hwm Hig
...
+	// exactly the range that is done. Committing newest-first would push the mark past
+	// signatures never decoded, which is the permanent gap this whole change is about.
+	for i, j := 0, len(allSigs)-1; i < j; i, j = i+1, j-1 {
+		allSigs[i], allSigs[j] = allSigs[j], allSigs[i]
+	}
+
+	// The tip is now the last element. Not-founds more than notFoundSkipSlotLag below it
+	// are pruned history and safe to skip.
+	var skipNotFoundBelow uint64
+	if tip := allSigs[len(allSigs)-1].Slot; tip > notFoundSkipSlotLag {
```
ben-dz, line 375: An escrow whose newest unindexed signature is itself unretrievable can never reach the escape hatch and so never advances. `skipNotFoundBelow` is derived from `allSigs[len(allSigs)-1].Slot` after the reversal, so for the tip signature `sig.Slot < skipNotFoundBelow` is always false and the chunk fails on every refresh. Chunk commits bound the loss to the newest ~300 slots, and the same arithmetic exists in `permissionevents`; deriving the tip from the chain rather than the batch would close it in both.

PR #755, opened 2026-08-06
rollup: fail a backfill that skipped a chunk instead of reporting success

bgm-malbeclabs:
## Summary

- `BackfillRollupWorkflow` advanced `chunkStart` unconditionally after `runIteration` swallowed the activity error, so a transient failure on one chunk left a **permanent hole** in the backfilled rollups — nothing ever recomputes it — and the workflow returned `nil`, reporting a complete backfill.
- Failed chunks are now collected and the workflow fails with the count and the range to re-run. The loop keeps going, so one bad window doesn't abandon the rest of the range.
- `runIteration` now returns the activity error in addition to logging it via the escalator. The **live** `ComputeRollupWorkflow` discards the return and keeps looping, unchanged — its overlap window (`rollupWindow`) gives each bucket several chances, so swallowing is correct there. The backfill has no such second pass, which is the whole difference.
- A cancellation is reported as `ctx.Err()` rather than a chunk failure, so shutdown doesn't get recorded as a gap.

## Why the backfill is different from the live rollup

The live rollup recomputes a sliding window every 30s, so a bucket that fails one cycle is retried on the next several. The backfill is a single pass over a historical range: a window it skips is never revisited by anything. Silently skipping it while reporting success is the worst of both — the data is missing and nobody knows.

Re-running is safe: `link_rollup_5m` and `device_interface_rollup_5m` are both `ReplacingMergeTree(ingested_at)` ordered by the bucket key, so recomputing a window is idempotent. The error message says so.

No ERROR log is added here — the escalator already logged each chunk failure with its `window_start`/`window_end`, and this is a manually-triggered workflow whose error Temporal surfaces to the operator. A second alert-bearing line would page on-call for a hand-run backfill.

## Testing Verification

- Adds the package's first workflow-level tests (`workflow_test.go`); previously only the activities had coverage.
- `FailedChunkMustNotBeSilentlySkipped` — three chunks with the middle one failing. Asserts the workflow errors, that the message names the failed window so an operator can re-run it, and that it reports `1 of 3 chunks failed` — the last assertion is what stops the test passing for the wrong reason if *every* chunk fails instead.
- Confirmed the test fails on the pre-change code (`An error is expected but got nil`) and passes after, by stashing only the `workflow.go` change.
- `AllChunksSucceedReportsSuccess` pins the other side: the new failure path must not make a clean backfill look broken.
- `go test -race` green on `indexer/pkg/rollup`.

## Found by

An audit of every batch path in the indexer for the invariant *item-failure handling must match how the resume point is derived*. Sibling PR #754 fixes the same family in the escrow-events indexer (a skipped signature dragged the fact-derived high-water mark past itself).

### DIFF
--- a/indexer/pkg/rollup/workflow.go
+++ b/indexer/pkg/rollup/workflow.go
@@ -1,6 +1,8 @@
 package rollup
 
 import (
+	"errors"
+	"fmt"
 	"time"
 
 	"go.temporal.io/sdk/log"
@@ -65,7 +67,10 @@ func ComputeRollupWorkflow(ctx temporalworkflow.Context, iteration int) error {
 			WindowEnd:   now,
 		}
 
-		runIteration(ctx, log, esc, window)
+		// The live loop must keep running: a bucket stays in the overlap window
+		// (rollupWindow) for several iterations, so a failed cycle gets another
+		// chance on its own. The escalator already logged it.
+		_ = runIteration(ctx, log, esc, window)
 
 		iteration++
 
@@ -79,20 +84,27 @@ func ComputeRollupWorkflow(ctx temporalworkflow.Context, iteration int) error {
 	return temporalworkflow.NewContinueAsNewError(ctx, ComputeRollupWorkflow, 0)
 }
 
-// runIteration executes one rollup cycle. Errors are logged, not returned,
-// so the workflow loop continues on failure.
-func runIteration(ctx temporalworkflow.Context, log log.Logger, esc *lakelogger.Escalator, window BackfillChunkInput) {
-	err := temporalworkflow.ExecuteActivity(ctx, (*Activities).RollupLinks, window).Get(ctx, nil)
-	if err != nil && ctx.Err() != nil {
-		return // workflow context done; the error is just the cancellation
+// runIteration executes one rollup cycle. Errors are logged via the escalator,
+// which owns the paging decision, and also returned so a caller that must not
+// silently skip a window can act on them. The live loop discards the return and
+// continues; the backfill records it (see BackfillRollupWorkflow).
+//
+// A cancellation is reported as ctx.Err() rather than an activity failure: the
+// window did not fail, the workflow is shutting down.
+func runIteration(ctx temporalworkflow.Context, log log.Logger, esc *lakelogger.Escalator, window BackfillChunkInput) error {
+	linkErr := temporalworkflow.ExecuteActivity(ctx, (*Activities).RollupLinks, window).Get(ctx, nil)
+	if linkErr != nil && ctx.Err() != nil {
+		return ctx.Err() // workflow context done; the error is just the cancellation
 	}
-	esc.Observe(log, "link_rollup", "link rollup failed", err, "window_start", window.WindowStart, "window_end", window.WindowEnd)
+	esc.Observe(log, "link_rollup", "link rollup failed", linkErr, "window_start", window.WindowStart, "window_end", window.WindowEnd)
 
-	err = temporalworkflow.ExecuteActivity(ctx, (*Activities).RollupDeviceInterfaces, window).Get(ctx, nil)
-	if err != nil && ctx.Err() != nil {
-		return
+	intfErr := temporalworkflow.ExecuteActivity(ctx, (*Activities).RollupDeviceInterfaces, window).Get(ctx, nil)
+	if intfErr != nil && ctx.Err() != nil {
+		return ctx.Err()
 	}
-	esc.Observe(log, "device_interface_rollup", "device interface rollup failed", err, "window_start", window.WindowStart, "window_end", window.WindowEnd)
+	esc.Observe(log, "device_interface_rollup", "device interface rollup failed", intfErr, "window_start", window.WindowStart, "window_end", window.WindowEnd)
+
+	return errors.Join(linkErr, intfErr)
 }
 
 // BackfillRollupWorkflow processes historical data in time chunks.
@@ -131,6 +143,16 @@ func BackfillRollupWorkflow(ctx temporalworkflow.Context, input BackfillInput) e
 		log.Info("capped backfill end time to avoid live rollup overlap", "end_time", endTime)
 	}
 
+	// A backfill is a one-shot pass: unlike the live rollup there is no overlap
+	// window and no later cycle to recompute a window that failed. Advancing past
+	// a failed chunk therefore leaves a permanent hole in the rollups. Collect the
+	// failures instead — keep going so one bad window doesn't abandon the rest of
+	// the range, then fail with the windows to re-run. Re-running is safe: both
+	// rollup tables are ReplacingMergeTree keyed on the bucket, so recomputing a
+	// window is idempotent.
+	var failed []BackfillChunkInput
+	totalChunks := 0
+
 	chunkStart := input.StartTime
 	for chunkStart.Before(endTime) {
 		chunkEnd := chunkStart.Add(input.ChunkSize)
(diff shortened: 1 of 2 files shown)


indexer/pkg/rollup/workflow.go
```diff
@@ -65,7 +67,10 @@ func ComputeRollupWorkflow(ctx temporalworkflow.Context, iteration int) error {
 			WindowEnd:   now,
 		}
 
-		runIteration(ctx, log, esc, window)
+		// The live loop must keep running: a bucket stays in the overlap window
+		// (rollupWindow) for several iterations, so a failed cycle gets another
+		// chance on its own. The escalator already logged it.
+		_ = runIteration(ctx, log, esc, window)
```
ben-dz, line 73: Nothing tests that the live rollup keeps looping when a rollup activity fails. Before this diff `runIteration` had no return value, so the property was structural and unbreakable; now a discarded error is the only thing holding it, and a discarded error looks like a bug to the next reader.

If someone "fixes" this to `if err := ...; err != nil { return err }`, the live rollup dies on the first ClickHouse blip — rollups stop until someone notices stale dashboards, and nothing in CI fails. Add the sibling test to the new file: execute `ComputeRollupWorkflow` at `continueAsNewThreshold-2` with both activity mocks always failing, assert `errors.As(env.GetWorkflowError(), **workflow.ContinueAsNewError)`.

indexer/pkg/rollup/workflow.go
```diff
@@ -144,10 +166,25 @@ func BackfillRollupWorkflow(ctx temporalworkflow.Context, input BackfillInput) e
 			SourceDatabase: input.SourceDatabase,
 		}
 
-		runIteration(ctx, log, esc, chunk)
+		totalChunks++
+		if err := runIteration(ctx, log, esc, chunk); err != nil {
+			if ctx.Err() != nil {
+				return err // shutting down, not a chunk failure
```
ben-dz: On cancellation this returns the chunk error rather than the cancellation, and discards everything collected in `failed`. The branch is reachable without a race: `RollupLinks` fails while the context is live, `RollupDeviceInterfaces` then succeeds, and the cancel arrives during that second activity — `runIteration` returns the link error, this check fires, and the workflow terminates Failed with a chunk error instead of Canceled.

The same `return err` skips the `len(failed) > 0` block, so a backfill that failed 12 chunks and is cancelled at chunk 400 reports nothing about those gaps — the exact information this PR exists to surface. Return `ctx.Err()` explicitly here, and when `len(failed) > 0` wrap it with the same count/range summary, keeping the `*temporal.CanceledError` in the chain so the workflow still ends Canceled.

indexer/pkg/rollup/workflow.go
```diff
@@ -144,10 +166,25 @@ func BackfillRollupWorkflow(ctx temporalworkflow.Context, input BackfillInput) e
...
+			failed = append(failed, chunk)
+		}
 
 		chunkStart = chunkEnd
 	}
 
+	if len(failed) > 0 {
+		// Each failure was already logged with its window by the escalator, so this
+		// carries the summary and the range to re-run rather than repeating them.
+		return fmt.Errorf("backfill incomplete: %d of %d chunks failed, leaving gaps between %s and %s; re-run the backfill over the failed windows (each is logged with window_start/window_end)",
```
ben-dz: The message names a range spanning the successful chunks between the failures, and the per-window detail it defers to isn't reachable from where the operator is standing. With 10 chunks where only the first and last fail it reads `2 of 10 chunks failed, leaving gaps between <t0> and <t10>` — the entire backfill range, for two one-hour holes.

The `window_start`/`window_end` values live in indexer worker logs in the cluster, while the operator who ran `admin --start-backfill-rollup` sees only this string (`admin/internal/admin/backfill_rollup.go:78`), so re-running means redoing the whole range. Append the failed window starts to the message, capped at ~10 plus "and N more", and extend the existing test to assert both failed windows appear.

indexer/pkg/rollup/workflow.go
```diff
@@ -144,10 +166,25 @@ func BackfillRollupWorkflow(ctx temporalworkflow.Context, input BackfillInput) e
...
+			if ctx.Err() != nil {
+				return err // shutting down, not a chunk failure
+			}
+			failed = append(failed, chunk)
+		}
 
 		chunkStart = chunkEnd
 	}
 
+	if len(failed) > 0 {
```
ben-dz, line 207: A backfill that processes zero chunks still returns `nil`, so the CLI prints "rollup backfill complete" for a run that covered nothing — the same failure class this PR exists to eliminate. A reversed or mistyped range reaches here unvalidated: neither `resolveTimeFlags` (`admin/cmd/admin/main.go:540-576`) nor `BackfillRollup` checks that start precedes end.

`totalChunks` is already in hand, so a guard before this block closes it: `if totalChunks == 0 { return fmt.Errorf("backfill covered no chunks: start %s is not before end %s", ...) }`.

indexer/pkg/rollup/workflow.go
```diff
@@ -131,6 +143,16 @@ func BackfillRollupWorkflow(ctx temporalworkflow.Context, input BackfillInput) e
 		log.Info("capped backfill end time to avoid live rollup overlap", "end_time", endTime)
 	}
 
+	// A backfill is a one-shot pass: unlike the live rollup there is no overlap
```
ben-dz: The same rationale is restated four times — "the live loop has an overlap window so it self-heals; the backfill is one-shot; ReplacingMergeTree makes re-running safe" appears here and at lines 70-72 and 87-93, plus `workflow_test.go:22-34`. That's four copies to keep in sync as the code moves; keep the canonical statement on `runIteration` and let the other three point at it.

ben-dz (commented):
No blocking findings. Four things worth addressing before merge. Inline below

indexer/pkg/rollup/workflow.go
bgm-malbeclabs: Fixed in c9216a6. The loop now returns `ctx.Err()` on shutdown, wrapped with the same count/range summary when chunks failed, so the workflow still ends Canceled (`errors.As` on `*CanceledError`) and still names the gaps. A chunk that failed on its own before the cancel is recorded as a gap; the cancellation itself is not.

New test `CancellationStillReportsGaps` covers it: chunk 1 fails, the cancel lands during chunk 2. On the pre-change code it fails with `"...: canceled" does not contain "2026-08-01T00:00:00Z"`.

indexer/pkg/rollup/workflow.go
bgm-malbeclabs: Fixed in c9216a6. The full statement now lives on `runIteration` as a two-bullet contrast (live loop self-heals via the overlap window; backfill is one-shot, so re-run is needed and safe under ReplacingMergeTree). The two call sites and the test comment point at it.

indexer/pkg/rollup/workflow.go
bgm-malbeclabs: Done in 7fb7668, not just resolved — the claim held. The message now lists each failed window start (capped at 10, then `and N more`) plus the chunk size, instead of the outer range.

New test `NamesEveryFailedWindow`: 4 chunks, first and last fail. It asserts both failed starts appear **and** that the two that succeeded do not, so the range-spanning form cannot come back. On the old code it fails with `...leaving gaps between 2026-08-01T00:00:00Z and 2026-08-01T04:00:00Z... does not contain "2026-08-01T03:00:00Z"` — the exact shape you described.

indexer/pkg/rollup/workflow.go
bgm-malbeclabs: Done in 7fb7668. Guard added, using the capped `endTime` rather than `input.EndTime`: the cap reaches the same dead end without any typo, e.g. `--start-time-ago 20m` moves the end 30m behind the start.

New test `ZeroChunksIsNotSuccess` (reversed range) fails on the old code with `An error is expected but got nil`.

indexer/pkg/rollup/workflow.go
bgm-malbeclabs: Done in 0ef4147. `KeepsLoopingWhenActivitiesFail` executes `ComputeRollupWorkflow` at `continueAsNewThreshold-2` with both activity mocks always failing and requires `*workflow.ContinueAsNewError` in the chain.

Checked it catches the wrong fix, not just the current code: swapping the discard for `if err := ...; err != nil { return err }` makes it fail with the RollupLinks error in place of the continue-as-new.

PR #756, opened 2026-08-06
indexer: make telemetry latency fetch skips visible

bgm-malbeclabs:
## Summary

- Every per-circuit RPC error in the device-link and internet-metro fan-outs hit a bare `continue` — no log, no metric, no counter. The circuit was dropped and the activity reported success.
- Dropping is genuinely **safe** here: each circuit resumes from its own stored max sample index, so the next refresh re-fetches what this one missed. This is not data loss. It is a pure observability hole — a circuit failing persistently, or a throttled subset, under-collects indefinitely with nothing anywhere to say so.
- Unclassified fetch errors now increment `doublezero_data_indexer_telemetry_fetch_skipped_total{source}` and log at WARN with the circuit and epoch.
- Expected conditions stay uncounted and unchanged: worker shutdown, and no telemetry account for a circuit/epoch (routine and very common — counting it would bury the signal).
- The three-way classification was written out identically at both call sites. It now lives in one `classifyFetchErr` so the two cannot drift apart.

WARN, not ERROR: a single skipped circuit self-heals next refresh and is not actionable on its own. The counter is what makes a *sustained* rate visible, which is the actionable condition.

## Why this is separate from the other audit PRs

The sibling PRs (#754 escrow events, #755 rollup backfill) fix real data loss — a resume point that advances past work that was never done. This path gets that part right; the only thing wrong is that nobody can see it happening. No data is at risk, so it ships last.

## Testing Verification

- `ClassifyFetchErr` table test covers all three outcomes including wrapped errors: cancel/deadline → abort, `ErrAccountNotFound` → expected miss, and the cases that used to vanish silently (RPC throttle, service-unavailable, connection reset, decode failure) → counted.
- The classification is what the test pins deliberately: an error class landing in the wrong bucket either hides a real outage (unclassified treated as expected) or floods the counter with routine misses. Both are silent failures of the fix itself.
- `go test -race` green on `dz/telemetry/...`.

## Found by

An audit of every batch path in the indexer for the invariant *item-failure handling must match how the resume point is derived*. This path satisfies the invariant; it just did so invisibly.

### DIFF
--- a/indexer/pkg/dz/telemetry/latency/view.go
+++ b/indexer/pkg/dz/telemetry/latency/view.go
@@ -25,7 +25,9 @@ import (
 type fetchOutcome int
 
 const (
-	// fetchAbort means the worker is shutting down: stop, don't record anything.
+	// fetchAbort means the refresh context is done — worker shutdown or the refresh
+	// deadline — so stop and record nothing. Shutdown is not a collection failure and
+	// must not be counted as one.
 	fetchAbort fetchOutcome = iota
 	// fetchExpectedMiss means there is simply no telemetry account for this
 	// circuit/epoch. Routine and very common — not worth counting.
@@ -38,15 +40,30 @@ const (
 	fetchUnclassified
 )
 
-func classifyFetchErr(err error) fetchOutcome {
-	switch {
-	case errors.Is(err, context.Canceled), errors.Is(err, context.DeadlineExceeded):
+// classifyFetchErr decides what a sample-fetch error means. Shutdown is decided by
+// the context, never by the error's shape.
+//
+// The distinction matters because an http.Client timeout also satisfies
+// errors.Is(err, context.DeadlineExceeded) — verified: a Client.Timeout abort
+// produces "context deadline exceeded (Client.Timeout exceeded while awaiting
+// headers)" and errors.Is against context.DeadlineExceeded is true. Keying on the
+// error therefore read a slow endpoint as "we are shutting down" and returned from
+// the worker goroutine, which sits before the append of everything already collected
+// for that circuit — so a timeout silently discarded the earlier epochs' samples
+// while the refresh still reported a clean cycle.
+//
+// With the context as the authority, a live context means the fetch genuinely failed:
+// it is counted, logged, and the circuit resumes from its stored max sample index next
+// refresh. Only a cancelled or expired context aborts, and the loop's own
+// select-on-ctx.Done already handles that case first.
+func classifyFetchErr(ctx context.Context, err error) fetchOutcome {
+	if ctx.Err() != nil {
 		return fetchAbort
-	case errors.Is(err, telemetry.ErrAccountNotFound):
-		return fetchExpectedMiss
-	default:
-		return fetchUnclassified
 	}
+	if errors.Is(err, telemetry.ErrAccountNotFound) {
+		return fetchExpectedMiss
+	}
+	return fetchUnclassified
 }
 
 type TelemetryRPC interface {
--- a/indexer/pkg/dz/telemetry/latency/view.go
+++ b/indexer/pkg/dz/telemetry/latency/view.go
@@ -19,6 +19,36 @@ import (
 	"github.com/malbeclabs/lake/utils/pkg/logger"
 )
 
+// fetchOutcome classifies a telemetry sample-fetch error. Both the device-link
+// and internet-metro fan-outs need the same three-way split, so it lives here
+// rather than being spelled out (and drifting) at each call site.
+type fetchOutcome int
+
+const (
+	// fetchAbort means the worker is shutting down: stop, don't record anything.
+	fetchAbort fetchOutcome = iota
+	// fetchExpectedMiss means there is simply no telemetry account for this
+	// circuit/epoch. Routine and very common — not worth counting.
+	fetchExpectedMiss
+	// fetchUnclassified is any other error. Skipping it is safe (each circuit
+	// resumes from its own stored max sample index, so the next refresh re-fetches
+	// what this one missed) but it MUST be counted and logged: the refresh still
+	// reports success, so an uncounted skip lets a persistently failing circuit
+	// under-collect indefinitely with no signal anywhere.
+	fetchUnclassified
+)
+
+func classifyFetchErr(err error) fetchOutcome {
+	switch {
+	case errors.Is(err, context.Canceled), errors.Is(err, context.DeadlineExceeded):
+		return fetchAbort
+	case errors.Is(err, telemetry.ErrAccountNotFound):
+		return fetchExpectedMiss
+	default:
+		return fetchUnclassified
+	}
+}
+
 type TelemetryRPC interface {
 	GetDeviceLatencySamplesTail(ctx context.Context, originDevicePK, targetDevicePK, linkPK solana.PublicKey, epoch uint64, existingMaxIdx int) (*telemetry.DeviceLatencySamplesHeader, int, []uint32, error)
 	GetInternetLatencySamples(ctx context.Context, dataProviderName string, originLocationPK, targetLocationPK, agentPK solana.PublicKey, epoch uint64) (*telemetry.InternetLatencySamples, error)
--- a/indexer/pkg/dz/telemetry/latency/view.go
+++ b/indexer/pkg/dz/telemetry/latency/view.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log/slog"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/gagliardetto/solana-go"
@@ -291,4 +292,17 @@ func (v *View) WaitReady(ctx context.Context) error {
 	case <-ctx.Done():
 		return fmt.Errorf("context cancelled while waiting for telemetry view: %w", ctx.Err())
 	}
+}
+
+// noteSkip records a skipped fetch for the per-refresh summary. Both fan-outs call it
+// from the same place, the arm that counts an unclassified error, so the two cannot
+// drift: a previous version of this accounting sat in the wrong branch on one path and
+// silenced that path's summary entirely.
+//
+// It stores a copy of err rather than the caller's variable, so the retained error
+// cannot be affected by later loop iterations.
+func noteSkip(count *atomic.Int64, first *atomic.Pointer[error], err error) {
+	count.Add(1)
+	e := err
+	first.CompareAndSwap(nil, &e)
 }
--- a/indexer/pkg/dz/telemetry/latency/view.go
+++ b/indexer/pkg/dz/telemetry/latency/view.go
@@ -32,30 +32,33 @@ const (
 	// fetchExpectedMiss means there is simply no telemetry account for this
 	// circuit/epoch. Routine and very common — not worth counting.
 	fetchExpectedMiss
-	// fetchUnclassified is any other error. Skipping it is safe (each circuit
-	// resumes from its own stored max sample index, so the next refresh re-fetches
-	// what this one missed) but it MUST be counted and logged: the refresh still
-	// reports success, so an uncounted skip lets a persistently failing circuit
-	// under-collect indefinitely with no signal anywhere.
+	// fetchUnclassified is any other error. The circuit resumes from its own stored
+	// max sample index, so the next refresh re-fetches what this one missed — but only
+	// while the epoch is still in the fetch window (current and current-1). Past that,
+	// roughly two days, only the admin backfill recovers it. So the skip must be
+	// counted and logged: the refresh still reports success, and an uncounted skip lets
+	// a failing circuit under-collect with no signal anywhere.
 	fetchUnclassified
 )
 
-// classifyFetchErr decides what a sample-fetch error means. Shutdown is decided by
-// the context, never by the error's shape.
+// classifyFetchErr decides what a sample-fetch error means. The context decides
+// shutdown, never the error's shape.
 //
-// The distinction matters because an http.Client timeout also satisfies
-// errors.Is(err, context.DeadlineExceeded) — verified: a Client.Timeout abort
-// produces "context deadline exceeded (Client.Timeout exceeded while awaiting
-// headers)" and errors.Is against context.DeadlineExceeded is true. Keying on the
-// error therefore read a slow endpoint as "we are shutting down" and returned from
-// the worker goroutine, which sits before the append of everything already collected
-// for that circuit — so a timeout silently discarded the earlier epochs' samples
-// while the refresh still reported a clean cycle.
+// The invariant: an http.Client timeout also satisfies
+// errors.Is(err, context.DeadlineExceeded). Keying the abort on the error therefore
+// reads a slow endpoint as shutdown and returns from the worker goroutine, and that
+// return sits before the append of everything already collected for the circuit. The
+// refresh still reports success, so the cycle looks clean with a hole in it.
 //
-// With the context as the authority, a live context means the fetch genuinely failed:
-// it is counted, logged, and the circuit resumes from its stored max sample index next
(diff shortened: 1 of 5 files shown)


bgm-malbeclabs:
Added `37491bf`: shutdown is now decided by `ctx.Err()` rather than the error's shape.

This is a finding from #757's review that belongs here, since this PR already rewrites these lines. An `http.Client` timeout also satisfies `errors.Is(err, context.DeadlineExceeded)` — verified:

```
err: context deadline exceeded (Client.Timeout exceeded while awaiting headers)
errors.Is(err, context.DeadlineExceeded) = true
```

So the old branch read a slow endpoint as "we're shutting down" and returned from the worker goroutine — and that return sits *before* the `samplesMu.Lock()` append, so every sample already collected for that circuit went with it. `Refresh` still returned nil: no log, no counter, no escalator observation. A clean cycle with a hole in it.

`classifyFetchErr` now takes the context. A cancelled or expired context aborts silently; a live context means the fetch genuinely failed, so it is counted and logged by the path this PR adds and the circuit resumes from its stored max sample index next refresh.

I kept the classification rather than deleting the branch outright. Deleting works for the data loss, but with this PR's counter in place every in-flight fetch during a real SIGTERM would then log a WARN and increment the skipped counter — turning shutdown into a burst of false collection failures, which CLAUDE.md explicitly places outside what should be recorded. Gating on `ctx.Err()` gets both: silent on shutdown, counted on failure.

The test now pins the regression directly rather than the old error-shape table: a `Client.Timeout` error (constructed against a real `httptest` server, so it keeps matching whatever `net/http` actually returns) on a live context must classify as a countable failure, and the same error on a cancelled context must abort.

**Merge order:** this should land before or with #757. That PR lowers the SDK's per-request timeout from 5 minutes to 10s, which is what makes this reachable — the fetch context carries no deadline of its own, so under the old default the branch effectively only fired on real SIGTERM.

indexer/pkg/metrics/metrics.go
```diff
@@ -54,6 +54,22 @@ var (
...
+	// success, so without this counter a circuit (or a throttled subset) can
+	// under-collect indefinitely with no signal at all. Expected conditions
+	// (no telemetry account for a circuit/epoch, worker shutdown) are not counted.
+	// A sustained non-zero rate means a subset of circuits is not being collected.
+	TelemetryFetchSkipped = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "doublezero_data_indexer_telemetry_fetch_skipped_total",
+			Help: "Telemetry sample fetches skipped on an unclassified RPC error, by source",
+		},
+		[]string{"source"},
```
ben-dz: The counter can't tell you which network it's reporting on, which defeats the alerting it exists to enable. One indexer process runs mainnet plus devnet and testnet together (`indexer/cmd/indexer/main.go:743`), so a devnet RPC hiccup and a real mainnet under-collect land in the same `{source="device_link"}` series — an operator watching a `rate()` sees devnet noise and mutes it. Every other per-env metric in this file carries the label for this reason (lines 150, 159, 179, 211, 224, with the rationale at line 204), and this package's store already does `WithLabelValues(s.cfg.DZEnv, "device_link", ...)` at `store.go:182`.

Fix: `[]string{"dz_env", "source"}` here, and pass `v.cfg.DZEnv` (already on `ViewConfig`, `view.go:91`) at both call sites. 3 envs × 2 sources = 6 series. Better before merge — adding a label to a counter someone has queried is a breaking change.

indexer/pkg/dz/telemetry/latency/view_test.go
```diff
@@ -1740,3 +1743,68 @@ func TestLake_TelemetryLatency_View_AgentVersionCommit(t *testing.T) {
...
+
+// TestLake_TelemetryLatency_ClassifyFetchErr pins that shutdown is decided by the
+// context, not by the error's shape.
+//
+// The regression this guards: an http.Client timeout also satisfies
+// errors.Is(err, context.DeadlineExceeded), so classifying on the error read a slow
+// endpoint as shutdown and returned from the worker goroutine — before the append of
+// everything already collected for that circuit. A timeout therefore discarded earlier
+// epochs' samples silently, with the refresh still reporting a clean cycle.
+func TestLake_TelemetryLatency_ClassifyFetchErr(t *testing.T) {
```
ben-dz, line 1760: The test pins the four-line classifier but not the call-site behavior the PR is actually about. Revert `view_device.go` and `view_internet.go` to `main` while keeping `classifyFetchErr` and the suite still passes — no mock in this file ever returns a non-nil error from a fetch, so a refactor that restores `errors.Is(err, context.DeadlineExceeded) → return` reintroduces the silent sample loss with everything green.

The harness exists: `mockTelemetryRPCWithIncrementalSamples.getSamplesFunc` (line 1133) plus the ClickHouse-backed setup around line 760. `epochsToFetch` is current-first (`view_device.go:35`), so have epoch N return samples and epoch N−1 return the timeout-shaped error, then assert epoch N's rows still land in `fact_dz_device_link_latency` and the counter went up by 1. ~40 lines; the testcontainer is already paid for.

indexer/pkg/dz/telemetry/latency/view.go
```diff
@@ -19,6 +19,53 @@ import (
...
+	// what this one missed) but it MUST be counted and logged: the refresh still
+	// reports success, so an uncounted skip lets a persistently failing circuit
+	// under-collect indefinitely with no signal anywhere.
+	fetchUnclassified
+)
+
+// classifyFetchErr decides what a sample-fetch error means. Shutdown is decided by
+// the context, never by the error's shape.
+//
+// The distinction matters because an http.Client timeout also satisfies
```
ben-dz: This states as fact a production data-loss event the deployed wiring cannot produce. The activity ctx carries a 5-minute `StartToCloseTimeout` (`indexer/pkg/dzingest/workflow.go:93`) and the RPC client's own `Timeout` is also 5 minutes (`doublezero/tools/solana/pkg/rpc/retry.go:16,48`), with the HTTP request always starting later — so `ctx.Err()` is non-nil first and the old code took the abort branch anyway.

The invariant is right and becomes load-bearing the moment anyone tightens the client timeout below the activity deadline or calls `Refresh` from a deadline-free ctx; the history attached to it isn't, and the next maintainer sizing the risk here will act on it. Fix: state the invariant plus the condition under which it bites (the 5m/5m relationship is why it hadn't fired), drop the past-tense claim, and keep one canonical copy — `view_test.go:1747-1754` and `metrics.go:57-64` restate the same paragraph and will diverge on the first behavior change.

indexer/pkg/dz/telemetry/latency/view_device.go
```diff
@@ -123,13 +122,18 @@ func (v *View) refreshDeviceLinkTelemetrySamples(ctx context.Context, devices []
...
+							switch classifyFetchErr(ctx, err) {
+							case fetchAbort:
 								return
-							}
-							if errors.Is(err, telemetry.ErrAccountNotFound) {
+							case fetchExpectedMiss:
+								continue
+							default:
+								metrics.TelemetryFetchSkipped.WithLabelValues("device_link").Inc()
+								v.log.Warn("telemetry/device-link: sample fetch failed, skipping until next refresh",
```
ben-dz: The WARN fires per circuit *per epoch*, so it floods logs in exactly the case the counter is designed to catch. A total outage is bounded (`GetEpochInfo` fails first and the refresh returns before the fan-out), so this is specifically the partial/throttle case, which is when someone is reading the logs.

Fix: keep `TelemetryFetchSkipped.Inc()` per fetch, demote this line to Debug, and emit one WARN per source after `wg.Wait()` with the skip count, the total, and the first error. Same at `view_internet.go:151`.

indexer/pkg/metrics/metrics.go
```diff
@@ -54,6 +54,22 @@ var (
 		},
 	)
 
+	// TelemetryFetchSkipped counts telemetry sample fetches the latency indexer
+	// dropped on an unclassified RPC error. The drop itself is safe — each circuit
```
ben-dz: "The drop itself is safe" is stated without its expiry date, understating how urgently an operator should respond. `epochsToFetch` is `{current, current−1}` (`view_device.go:35`, `view_internet.go:36`), so the automatic re-fetch only recovers a skipped circuit while that epoch stays in the two-epoch window — past roughly two days only the admin backfill gets it back. Add that bound here and at `view.go:35-39`.

ben-dz (changes requested):
The counter needs a `dz_env` label before merge — one indexer process runs mainnet, devnet and testnet together, so as written devnet noise and a real mainnet under-collect are the same series and the alert this metric exists for isn't actionable. Adding the label later breaks any query already written against it.

Beyond that: the call-site behavior the PR is about is untested (only the pure classifier is covered, and the harness for the real test already exists); the source comments assert a production data-loss event that the deployed 5m-activity / 5m-client-timeout wiring cannot actually produce; and the per-epoch WARN emits thousands of lines a minute under exactly the throttling case the counter targets.

The classifier itself is correct — verified `ErrAccountNotFound` is the SDK's only routine miss and is `errors.Is`-matchable, that no other routine condition lands in the counted bucket, and that the resume point genuinely cannot advance past a skipped fetch on either path.

bgm-malbeclabs:
All five fixed in `80869b2`. The second one was the important catch, and I verified it before fixing it.

**The call-site test was missing, and you were right that the suite proved nothing.** I reverted `view_device.go` and `view_internet.go` to `main`, keeping `classifyFetchErr`, and the entire latency suite passed. The silent loss was fully restored and everything was green. No mock in the file returns a non-nil fetch error, so the call site was never exercised.

The new test serves samples for epoch 100 and a real `Client.Timeout` error for epoch 99, relying on `epochsToFetch` being current-first, then asserts epoch 100's rows still land. It discriminates:

```
with the fix:                    ok
view_device.go reverted to main: FAIL — 0 rows where 3 belong
```

The error message names the mechanism, so a future failure explains itself. I also added `latencyFixture` while in there, since nine tests inline the same serviceability block and I did not want to add a tenth.

**`dz_env` added.** `[]string{"dz_env", "source"}`, passed `v.cfg.DZEnv` at both call sites. `DZEnv` was already on `ViewConfig` and set at `indexer.go:174`, so this was purely an omission on my part. Agreed it had to happen before merge rather than after someone queries the series.

**The comment claimed history that cannot have happened.** Your reasoning holds: client timeout 5m, activity `StartToCloseTimeout` 5m, request starts later, so `ctx.Err()` was non-nil first and the abort was genuine every time. The comment now states the invariant and the condition that makes it bite, which is #757 lowering the client timeout below the activity deadline. I kept one canonical copy instead of the three I had written, which would have diverged on the first behaviour change.

**Per-epoch WARN demoted to Debug**, with one WARN after `wg.Wait()` carrying the skip count and first error. You are right that a total outage is bounded because `GetEpochInfo` fails first, so this only ever floods in the partial case, which is exactly when someone is reading.

**Recovery bound recorded.** `epochsToFetch` is `{current, current-1}`, so a skip self-heals only while its epoch stays in the window. Past roughly two days only the admin backfill recovers it. That is now stated on the metric and on `fetchUnclassified`, because "safe to skip" without the expiry understates how fast someone should respond.

One correction to something I told you on #757: I said Finding 2 there was the load-bearing one because a correctly classified timeout becomes a counted, self-healing skip. The two-day window makes "self-healing" narrower than I implied. It heals within the window and not after.

`go test -race` green across `dz/telemetry/...`; lint clean.

indexer/pkg/dz/telemetry/latency/view_device.go
```diff
@@ -123,15 +125,22 @@ func (v *View) refreshDeviceLinkTelemetrySamples(ctx context.Context, devices []
...
+								metrics.TelemetryFetchSkipped.WithLabelValues(v.cfg.DZEnv, "device_link").Inc()
+								v.log.Debug("telemetry/device-link: sample fetch failed, skipping until next refresh",
+									"origin_device_pk", originDevicePK, "target_device_pk", targetDevicePK,
+									"link_pk", linkPKStr, "epoch", epoch, "error", err)
 								continue
 							}
-							continue
 						}
 						if hdr == nil {
+							skipped.Add(1)
```
ben-dz: The skip accounting landed in the wrong branch, so device-link fetch failures now produce no WARN at all. `skipped.Add(1)` and `firstSkipErr` are inside `if hdr == nil` rather than the `default:` arm on line 133 where the counter is incremented — `view_internet.go:157-158` places the identical pair correctly inside `default:`.

Two effects: an unclassified fetch error increments the metric and logs at Debug but never touches `skipped`, so the summary WARN on line 205 never fires; and `hdr == nil` is only reachable when `err == nil`, so the captured `&err` points at nil and the summary would log `first_error=<nil>`. The SDK returns `ErrAccountNotFound` whenever the account is missing (`smartcontract/sdk/go/telemetry/client.go:183,188`) and never nil-header-with-nil-error, so that branch does not fire against the real client — the device-link summary WARN is dead code.

Device-link is the larger fan-out (~672 fetches per refresh across 168 links), and it is now strictly less visible than before this commit, which at least had a per-fetch WARN. CI stays green because the new test asserts the counter, not the log. Fix: move both statements into the `default:` arm next to `metrics.TelemetryFetchSkipped.WithLabelValues(...).Inc()` and delete them from the `hdr == nil` branch.

ben-dz (changes requested):
One new problem: the per-refresh skip accounting for device-link landed in the wrong branch, so device-link fetch failures now reach no WARN at all — only Debug and the counter. That is less visible than before this commit. The internet-metro path has it right and shows exactly what the device path should look like.

bgm-malbeclabs:
Fixed in `c4f8101`. Your diagnosis is exactly right, including the `<nil>` consequence and why CI stayed green.

**Cause was my patching method.** I moved that accounting by matching indentation depth. The marker I used, a 7-tab `continue` followed by a 6-tab `}`, matches the `hdr == nil` block in `view_device.go` and the intended `default:` arm in `view_internet.go`, because the internet fan-out nests one level deeper. My assertion that the marker was unique passed on both files, so nothing warned me. That is a bad way to edit Go and I should have anchored on the surrounding statements instead.

**Fix.** Both paths now call one helper from the arm that counts:

```go
func noteSkip(count *atomic.Int64, first *atomic.Pointer[error], err error) {
	count.Add(1)
	e := err
	first.CompareAndSwap(nil, &e)
}
```

One call site each, immediately after `TelemetryFetchSkipped...Inc()`, so the two cannot drift again. It also stores a copy of the error rather than the caller's variable, which removes any question about the retained pointer and later loop iterations.

**Added the coverage that would have caught it.** You noted CI stayed green because the new test asserts the counter, which still moved. That is the real gap, so the fix is not enough on its own. There is now a `slog.Handler` that records emitted records, and a test asserting exactly one summary WARN per refresh carrying a non-zero count and a real first error.

Verified by reintroducing the bug exactly as it was, moving `noteSkip` back into `hdr == nil`:

```
with the bug:  FAIL — no summary WARN was emitted for a skipped fetch
restored:      ok
```

That handler is reusable, and it closes a category rather than a case: this package had no way to assert on log output at all, which is why an operator-facing regression could pass every existing test.

Two things I got wrong across this review worth stating plainly. I claimed a timeout was self-healing, when the fetch window is `{current, current-1}` and it only heals inside roughly two days. And I introduced this bug in the commit that was meant to make skips more visible, on the larger of the two fan-outs. Both were caught by reading rather than by the suite, which is the argument for the log assertion now being there.

`go test -race` green across `dz/telemetry/...`; lint clean.

ben-dz (approved):
lgtm — the skip accounting now calls `noteSkip` from the counting arm on both paths, and the new `capturingHandler` test asserts the summary WARN fires exactly once with a real count and error, which is the assertion that was missing when the counter-only test stayed green through the regression.

PR #757, opened 2026-08-06
deps: bump doublezero SDK to get working RPC retry classification

bgm-malbeclabs:
## Summary

One-line `go.mod` bump, but it is the closest thing to a root-cause fix for the permission-events throttle incident. **On the SDK version we are running, an RPCPool 429 is not retryable at all** — every cycle got exactly one attempt and aborted.

Two independent reasons, both fixed upstream:

**1. Retry classification is dead code.** The pinned version tests retryability with interface assertions:

```go
type hasStatusCode interface{ StatusCode() int }
type hasCode interface{ Code() int }
```

`jsonrpc.RPCError` and `jsonrpc.HTTPError` expose `Code` as a **struct field, not a method**, so `errors.As` matches neither. The newer version reads the codes off the concrete types and says so explicitly:

> Both expose their code as a struct field, not a method, so an interface assertion on `StatusCode()`/`Code()` matches nothing in production — that was the bug that made every Go ledger reader give up after one attempt on RPCPool's 503s.

**2. The message fallback does not cover it either.** Pinned list includes `"rate limited"`; the actual error reads `"Too many requests for a specific RPC call"`. `"too many requests"` and `"service unavailable"` are both added in the newer version.

Which explains the shape of the incident precisely: runs failing in 8–14 seconds, nowhere near any budget, for 4.5 hours.

## Also in the bump

- **Jittered backoff** — `d/2 + rand.N(d/2+1)`, added so *"the ~60 doublezerod hosts and dozen services reading the same ledger endpoint do not retry in lockstep and re-spike an endpoint that is already shedding load."*
- **Per-request timeout 5m → 10s.** The pinned client uses `defaultTimeout = 5 * time.Minute` for the HTTP client, dial, and idle timeout alike, so one hung request can stall a drain goroutine for five minutes.
- **`doublezero_solana_rpc_retries_total{method}`** and **`_retries_exhausted_total{method}`**. Every other DoubleZero service already publishes these — `telemetry-state-ingest`, `controller`, `devices`, `doublezero_client`, `internet_latency_collector`, `monitor`, `device-health-oracle`. lake published neither, which is why a 4.5-hour sustained throttle produced no RPC-layer signal: two ERROR log lines and a staleness alert were the entire evidence base. Note that no `getTransaction` series exists from *any* service — lake is the only consumer of it, and the only one not reporting.
- Non-idempotent methods (`sendTransaction`, `requestAirdrop`) are now never retried regardless of options.

## Review surface

Narrower than a four-day version jump suggests. Of the nine SDK packages lake imports, **six are byte-identical**:

| package | changed |
|---|---|
| `config` | — |
| `controlplane/telemetry/pkg/config` | — |
| `sdk/geolocation/go` | — |
| `smartcontract/sdk/go/serviceability` | — |
| `smartcontract/sdk/go/telemetry` | — |
| `tools/maxmind/pkg/{geoip,metrodb}` | — |
| `tools/solana/pkg/jsonrpc` | retry.go, + metrics.go |
| `tools/solana/pkg/rpc` | retry.go |
| `sdk/shreds/go` | rpc.go, state.go |

`sdk/shreds/go` is the one worth a look beyond the RPC layer, since `rpc.go` and `state.go` both changed.

## Testing Verification

- Full `go test ./indexer/...` suite run against the bump. One failure, `TestHealthMulticastUserRate` in `dz/mroute` — confirmed **pre-existing** by running it on `main` without the bump and getting the identical ClickHouse error (`code: 27, Cannot parse input: expected '(' before: '-- grp-rate-gap: …'` — a `--` comment inside a `VALUES` list in `health_multicast_user_rate_test.go:42`). Unrelated to this change, but it means that package has no working local coverage right now.
- Builds clean across `indexer`, `api`, `utils`, `admin`. No source changes required — `go.mod` and `go.sum` only.

## Relationship to #753

#753 (graceful stop on throttle + request pacing) is still correct and still wanted, but its description overstates the deployed retry behavior — I have posted a correction there. The two are complementary: this PR makes a throttled request actually get retried, #753 makes a cycle bank its committed work when retries do run out, and #753's pacing keeps us from tripping the limit in the first place. **This one should land first**, or at least alongside.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/joho/godotenv v1.5.1
 	github.com/jonboulle/clockwork v0.5.0
 	github.com/lmittmann/tint v1.1.3
-	github.com/malbeclabs/doublezero v0.8.1-0.20260724204114-f7981ab6f7a3
+	github.com/malbeclabs/doublezero v0.8.1-0.20260728210138-d904a1969083
 	github.com/modelcontextprotocol/go-sdk v1.6.0
 	github.com/mr-tron/base58 v1.3.0
 	github.com/neo4j/neo4j-go-driver/v5 v5.28.4
--- a/admin/internal/admin/backfill_device_link_latency.go
+++ b/admin/internal/admin/backfill_device_link_latency.go
@@ -52,8 +52,16 @@ func BackfillDeviceLinkLatency(
 	}
 	defer chDB.Close()
 
-	// Create RPC clients
-	dzRPCClient := rpc.NewWithRetries(networkConfig.LedgerPublicRPCURL, nil)
+	// Resolved here rather than at first use: it sizes the connection pool below.
+	maxConcurrency := cfg.MaxConcurrency
+	if maxConcurrency <= 0 {
+		maxConcurrency = defaultBackfillMaxConcurrency
+	}
+
+	// Create RPC clients. Pool sized to the backfill fan-out: the SDK's per-request
+	// timeout includes time queued for a connection, so a smaller pool turns a slow
+	// backend into terminal timeouts rather than slow successes.
+	dzRPCClient := rpc.New(networkConfig.LedgerPublicRPCURL, rpc.Options{MaxConnsPerHost: maxConcurrency})
 	defer dzRPCClient.Close()
 
 	telemetryClient := telemetry.New(log, dzRPCClient, nil, networkConfig.TelemetryProgramID)
@@ -142,11 +150,6 @@ func BackfillDeviceLinkLatency(
 		fmt.Printf("No epochs available to backfill.\n")
 		fmt.Printf("To backfill specific epochs, use --start-epoch and --end-epoch flags.\n")
 		return nil
-	}
-
-	maxConcurrency := cfg.MaxConcurrency
-	if maxConcurrency <= 0 {
-		maxConcurrency = defaultBackfillMaxConcurrency
 	}
 
 	fmt.Printf("Backfill Device Link Latency\n")
--- a/admin/internal/admin/backfill_internet_metro_latency.go
+++ b/admin/internal/admin/backfill_internet_metro_latency.go
@@ -48,8 +48,16 @@ func BackfillInternetMetroLatency(
 	}
 	defer chDB.Close()
 
-	// Create RPC clients
-	dzRPCClient := rpc.NewWithRetries(networkConfig.LedgerPublicRPCURL, nil)
+	// Resolved here rather than at first use: it sizes the connection pool below.
+	maxConcurrency := cfg.MaxConcurrency
+	if maxConcurrency <= 0 {
+		maxConcurrency = defaultBackfillMaxConcurrency
+	}
+
+	// Create RPC clients. Pool sized to the backfill fan-out: the SDK's per-request
+	// timeout includes time queued for a connection, so a smaller pool turns a slow
+	// backend into terminal timeouts rather than slow successes.
+	dzRPCClient := rpc.New(networkConfig.LedgerPublicRPCURL, rpc.Options{MaxConnsPerHost: maxConcurrency})
 	defer dzRPCClient.Close()
 
 	telemetryClient := telemetry.New(log, dzRPCClient, nil, networkConfig.TelemetryProgramID)
@@ -138,11 +146,6 @@ func BackfillInternetMetroLatency(
 		fmt.Printf("No epochs available to backfill.\n")
 		fmt.Printf("To backfill specific epochs, use --start-epoch and --end-epoch flags.\n")
 		return nil
-	}
-
-	maxConcurrency := cfg.MaxConcurrency
-	if maxConcurrency <= 0 {
-		maxConcurrency = defaultBackfillMaxConcurrency
 	}
 
 	fmt.Printf("Backfill Internet Metro Latency\n")
--- a/go.sum
+++ b/go.sum
@@ -231,6 +231,8 @@ github.com/magiconair/properties v1.8.10 h1:s31yESBquKXCV9a/ScB3ESkOjUYYv+X0rg8S
 github.com/magiconair/properties v1.8.10/go.mod h1:Dhd985XPs7jluiymwWYZ0G4Z61jb3vdS329zhj2hYo0=
 github.com/malbeclabs/doublezero v0.8.1-0.20260724204114-f7981ab6f7a3 h1:67hJvUmdRNtXcW2BbUCBrH6kANhOL6tlqsg3lYD1Akk=
 github.com/malbeclabs/doublezero v0.8.1-0.20260724204114-f7981ab6f7a3/go.mod h1:TIfH1zvFiSLdeSNVr6HM7LbwNfkZXlHSTiDWdNpxE80=
+github.com/malbeclabs/doublezero v0.8.1-0.20260728210138-d904a1969083 h1:iBYx+Czk20vq7065s5nIqBjuTxdoG5B6+H+/aTfwurw=
+github.com/malbeclabs/doublezero v0.8.1-0.20260728210138-d904a1969083/go.mod h1:TIfH1zvFiSLdeSNVr6HM7LbwNfkZXlHSTiDWdNpxE80=
 github.com/mattn/go-colorable v0.1.14 h1:9A9LHSqF/7dyVVX6g0U9cwm9pG3kP9gSzcuIPHPsaIE=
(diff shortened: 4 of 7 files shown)


go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/joho/godotenv v1.5.1
 	github.com/jonboulle/clockwork v0.5.0
 	github.com/lmittmann/tint v1.1.3
-	github.com/malbeclabs/doublezero v0.8.1-0.20260724204114-f7981ab6f7a3
+	github.com/malbeclabs/doublezero v0.8.1-0.20260728210138-d904a1969083
```
ben-dz, line 25: This bump lowers the SDK's default per-request timeout from 5 minutes to 10 seconds, and that budget includes time spent waiting for a free connection, not just server work. Lake runs 64 concurrent RPC goroutines against one shared client with a 9-connection pool (`indexer/cmd/indexer/main.go:336`, fan-out at `indexer/pkg/dz/telemetry/latency/view_device.go:57`, default 64 from `main.go:59`), so each request's wall time is roughly 7x the server's service time — measured against the real client: 1s service time gives a slowest request of 8.0s, 1.5s gives 10/64 timeouts, 2s gives 28/64. The failure is also terminal, since `isRetryableJSONRPC` returns false on `context.DeadlineExceeded` before any other check, so a timed-out request gets one attempt and no retry.

Fix: the new SDK exposes `rpc.New(url, rpc.Options{...})`. At `main.go:336`, set `MaxConnsPerHost` to match `maxConcurrencyFlag`, or an explicit `RequestTimeout` sized for the queue depth — `NewWithRetries` can express neither. `admin/internal/admin/backfill_device_link_latency.go:56` has the same shape at concurrency 32. Pin it with an `httptest` backend at a fixed handler delay, the real constructor, `MaxConcurrency` goroutines, asserting zero timeouts.

ben-dz (changes requested):
A second change rides along in this bump that needs handling before it lands: the default per-request timeout drops from 5m to 10s, and that 10s includes time queued for a connection. Lake fans out 64 goroutines at a shared client with a 9-connection pool, so per-request wall time is ~7× the server's service time — measured against the real client, a 1s backend already burns 80% of the budget and a 1.5s backend fails 10/64 requests. The resulting error is then misread as shutdown at `view_device.go:126`, which abandons the goroutine before the append and silently drops every sample already collected for that circuit. Two fixes: pass `rpc.New(url, rpc.Options{...})` with a pool or timeout sized for the fan-out, and classify on `ctx.Err()` rather than on the error's shape.

- A request that hits the new 10s timeout is read as "we're shutting down" and silently throws away every sample already collected for that circuit. `indexer/pkg/dz/telemetry/latency/view_device.go:126` treats `context.DeadlineExceeded` as caller cancellation and returns from the worker goroutine — and that return sits before the `samplesMu.Lock()` append at `:177`, so the headers and samples gathered for earlier epochs go with it. `Refresh` still returns nil: no log, no error counter, no escalator observation, so the indexer reports a clean cycle with a hole in it.

This is pre-existing code that the bump activates. The fetch context carries no deadline of its own, so under the old 5-minute client timeout the branch effectively only fired on real SIGTERM; a `Client.Timeout` abort does satisfy `errors.Is(err, context.DeadlineExceeded)` after solana-go and the retry layer wrap it (verified end-to-end). `view_internet.go:144` is the same pattern, discarding `metroSamples`/`metroHeaders`. The table is append-only so a one-off gap self-heals, but sustained endpoint slowness stops telemetry for the affected circuits while every health signal stays green.

Fix: view_device.go:112-115 already handles real cancellation with case <-ctx.Done(): return at the top of the same loop. So the error-shape branch at :126 is redundant for genuine shutdown and harmful for client timeouts — every other error on that path already falls through to continue. The fix is to delete the branch, not to re-gate it. (I confirmed the return sits before the samplesMu.Lock() append at :177; I took the 64-goroutine latency table on trust.)

bgm-malbeclabs:
Both fixed. Pool sizing here in `9c61220`; the misclassification in #756 as `37491bf`, since that PR already rewrites those lines.

First — I verified your linchpin rather than taking it on trust, because both findings rest on it. A `Client.Timeout` abort does satisfy `errors.Is(err, context.DeadlineExceeded)`:

```
err: Get "http://127.0.0.1:56778": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
errors.Is(err, context.DeadlineExceeded) = true
errors.Is(err, context.Canceled)         = false
```

**Pool sizing (this PR).** `MaxConnsPerHost` now matches the fan-out sharing each client: `MaxConcurrency` for the indexer's ledger client on both the primary and secondary paths, and the resolved backfill concurrency for `backfill_device_link_latency.go` and `backfill_internet_metro_latency.go` — which also meant moving that resolution above the client it now sizes.

I chose the pool over `RequestTimeout` deliberately. A longer timeout would preserve the queueing and just tolerate it, and the SDK's own comment warns why that's worse: retry multiplies a hang by MaxAttempts, so a long per-attempt bound amplifies an endpoint stall rather than containing it. Sizing the pool makes the 10s budget mean what it says — a bound on service time, not on queue depth.

Worth stating explicitly since it cuts against the other thread on #753: **this does not worsen our rate-limit exposure.** Those caps are counted per method per source IP over a rolling 10s window, so 64 requests land in the window whether they arrive in one wave or eight. Only the burst shape changes, and the window count is what's measured.

**On the test.** I wrote the `httptest` version you specified and it passed — then found it didn't actually discriminate. At a 150ms handler delay, 64 requests through 9 connections is ~8 waves ≈ 1.2s, comfortably inside the 10s budget, so *zero timeouts either way*. Asserting on timeouts needs a service time above ~1.25s and therefore a test that runs over ten seconds.

So it asserts peak server-side concurrency instead, which always differs and is exact:

```
sized pool:    peak server concurrency 64, elapsed 155ms  -> pass
default pool:  peak server concurrency  9, elapsed 1.22s  -> fail
```

Confirmed by swapping the constructor back to `NewWithRetries` and watching it fail with that message. Timeout and non-timeout counts are still asserted as secondary guards.

**Merge order matters here and wasn't written down anywhere.** #757 lowers the timeout to 10s, which is what activates the misclassification; #756 is what stops a timeout being read as shutdown and silently discarding collected samples. **#756 should land first, or with it.** If #757 lands alone, a slow endpoint silently drops telemetry samples with every health signal green — worse than the status quo. I'll note the same on #756.

ben-dz (approved):
Good now. One thing left, non-blocking, take it or leave it: `permissionEventsRawRPC` is the one ledger client the sizing commit skipped.

- `permissionEventsRawRPC` (`indexer/cmd/indexer/main.go:356` and `:1021`) is the one ledger client the pool-sizing commit skipped, and it belongs to the component this PR was written for. It still uses `rpc.NewWithRetries` (pool 9) while its view runs `maxConcurrentFetches = 10` concurrent account drains (`indexer/pkg/dz/serviceability/permissionevents/view.go:254`) plus up to 10 in-flight `getTransaction` gated by `decodeSem` (`:118`, `:603`) — so 10–20 concurrent requests through 9 connections, the same amplification the other three clients were just fixed for. Fix: `rpc.New(url, rpc.Options{MaxConnsPerHost: ...})` sized to that view's fan-out, matching what the others now do.

bgm-malbeclabs:
Taken, in `04ba0a1`. Thanks for pushing on it, and your fan-out number is better than mine.

I had found this client in a sweep of the others and then backed the fix out, because exporting a constant from `view.go` produced the only merge conflict across the six branches (#753 rewrites that file). Putting the constant in `rpc.go` instead avoids that entirely. Verified clean against #753, #754, #755, #758 and main.

**On the number: 20, not 10.** I had assumed `maxConcurrentFetches` was the whole story. It isn't, because `fetchAccountSignatures` is not gated by `decodeSem` — confirmed, zero references to it in that function. So up to 10 accounts drain concurrently, each paginating `getSignaturesForAddress` on its own, while `decodeSem` separately caps in-flight `getTransaction` at 10 across the view. The two overlap, so the peak is 20 against a pool of 9.

```go
// MaxConcurrentRPCRequests is the peak number of in-flight HTTP requests this view
// makes, so a client serving it can size its connection pool to match.
const MaxConcurrentRPCRequests = 2 * maxConcurrentFetches
```

Derived rather than written as a literal, so changing either limit carries the pool with it. Both call sites use it.

**One thing to flag before this lands: #756 still reads as OPEN.** `mergedAt` is null and `origin/main` is still at `fdb78e2`, so the ordering constraint may not be satisfied yet. If this merges first, the 10s timeout is live while a timeout is still read as shutdown, and a slow endpoint silently discards collected samples with every health signal green. Worth confirming #756 is actually in before merging this.

PR #758, opened 2026-08-06
indexer: allow overriding the primary-env DZ ledger RPC endpoint

bgm-malbeclabs:
## Summary

- The SDK's `LedgerPublicRPCURL` is a shared RPCPool endpoint whose API key ships as a **public constant in the SDK**, so every consumer draws on one quota. lake can neither attribute its own usage nor defend its share — which is why the 4.5h `getTransaction` throttle on 2026-08-06 was undiagnosable from our side.
- Secondary envs could already be redirected via `DZ_LEDGER_RPC_URL_<ENV>`. The **primary env had no equivalent** and was hardwired to the shared constant.
- Adds `--dz-ledger-rpc-url` / `DZ_LEDGER_RPC_URL`, following the existing flag-then-env-override convention in this file.
- **No-op until an endpoint is supplied.** Landing it ahead of provisioning means a dedicated key becomes a secret change rather than a code change, review, and deploy.

## Scope of the redirect

The override is applied to `networkConfig` before any client is constructed, so it covers **every** DZ ledger read for that env, not just the audit indexer:

| consumer | reads |
|---|---|
| `dzRPCClient` | serviceability, geolocation |
| `permissionEventsRawRPC` | permission-events audit history |
| `shredsRPCURL` | shreds program state (testnet) |

All three read the same `networkConfig.LedgerPublicRPCURL` field. A dedicated endpoint should serve all of lake's ledger traffic, and this is the same assignment the secondary-env path already makes. If you'd rather scope it to permission events alone, that's a different shape — say so and I'll split the field.

## Deployment

No manifest change needed here. `k8s/base/indexer.yaml` already uses `envFrom` on the `lake-secrets` secret, so the var is picked up as soon as it exists there. **It belongs in the secret, not the configMap** — these endpoints carry their API key in the URL path.

Startup logging reflects the choice as a boolean (`dedicated_ledger_rpc: true|false`) and deliberately never logs the URL, for the same reason.

## Testing Verification

- Verified by inspection that the override is applied at `main.go:286`, ahead of all three client constructions (`:356`, `:366`, `:374`) — if it were moved below any of them it would silently become a no-op for that client, which is the one way this can break.
- No test added: this is config plumbing mirroring an existing pattern, and the ordering property above is not reachable without extracting the wiring out of `main`. Flagging it rather than papering over it.
- Builds clean; `golangci-lint` 0 issues.

## Context

Last of six PRs from the permission-events throttle investigation. #757 (SDK bump, working retry classification) is closest to the cause; #753 is the resilience layer; #754/#755/#756 are unrelated findings from auditing every batch path in the indexer. This one is the enabler for the actual fix — a dedicated key, now being requested from Triton.

Two things worth capturing when that conversation happens, since they turn guesses into derived values: the **per-method rate limit on `getTransaction`** (which would justify `defaultFetchesPerSecond` in #753 rather than leaving it a calibration guess), and whether the limit is **per-second, per-minute, or a burst bucket** (which determines whether `rate.Limiter`'s burst parameter is the right shape at all).


bgm-malbeclabs:
## Premise correction: a dedicated key may not be the fix

This PR's description calls it "the enabler for the actual fix — a dedicated key." That premise is now in question and should not be relied on in review.

Triton appears to rate limit **by source IP**, not by API key. Everything in the `doublezero-shared` k8s cluster egresses through one NAT'd IP, so a dedicated key would land in the same bucket. The RPC retry metrics confirm the co-tenancy — these all carry `k8s_cluster_name: doublezero-shared` alongside `lake-prod`:

- `dz-mainnet-beta/telemetry-state-ingest`
- `dz-testnet/telemetry-state-ingest`
- `dz-devnet/telemetry-flow-enricher`
- `dz-testnet/telemetry-flow-enricher`
- `device-health-oracle`

(`devices`, `doublezero_client`, `controller`, `monitor`, `funder`, `geoprobe_agent` and `internet_latency_collector` have no cluster label — they are outside k8s and do not share our egress.)

Following that further: the error is `"Too many requests for a specific RPC call"` — per-**method** — and no `getTransaction` series exists from any service in the fleet. lake is the only `getTransaction` caller. If the limit is per-method *and* per-IP, then neither a dedicated key nor even a dedicated egress IP helps, because we would be alone on it hitting the same per-method ceiling by ourselves. The only lever that works is reducing our own request rate, which is #753's limiter.

## Why keep this PR anyway

- Hardcoding a URL that embeds an API key is wrong independent of the quota question.
- It is a no-op until an endpoint is supplied, so it carries no risk while the policy question is open.
- If the limit turns out to be per-key after all, this is live again with no further work.
- Secondary envs already have this; the primary env lacking it was an inconsistency regardless.

## Open question that decides it

> Is the `getTransaction` limit enforced per source IP, per API key, or both — and what is the actual ceiling?

Being asked of Triton now. **Per-IP** → this stays dormant, and #753's pacing is the real fix. **Per-key** → provision one and set `DZ_LEDGER_RPC_URL`.

indexer/cmd/indexer/main.go
```diff
@@ -175,6 +181,9 @@ func run() error {
 	if envDZEnv := os.Getenv("DZ_ENV"); envDZEnv != "" {
 		*dzEnvFlag = envDZEnv
 	}
+	if envDZLedgerRPCURL := os.Getenv("DZ_LEDGER_RPC_URL"); envDZLedgerRPCURL != "" {
```
ben-dz: Setting this env var will also redirect the devnet and testnet indexers and stop them ingesting. The pinned SDK's `NetworkConfigForEnv` applies `DZ_LEDGER_RPC_URL` after its per-env switch (`config/env.go:188-191`), so `startSecondaryNetwork`'s call at `main.go:982` picks it up for both secondary envs; they then query the mainnet endpoint with devnet/testnet program IDs, get zero accounts, and stall on the `refusing to write snapshot` guard at `indexer/pkg/dz/serviceability/view.go:397`. The per-env overrides at `main.go:777-786` only rescue an env if `DZ_LEDGER_RPC_URL_DEVNET`/`_TESTNET` are also set, and neither is in any manifest; `api/handlers/ledger.go:59` reads the same var from the same shared secret, so the API's `/ledger` page moves too.

Fix: drop this block and resolve the primary env through the name the secondary path uses — `os.Getenv("DZ_LEDGER_RPC_URL_" + strings.ToUpper(strings.ReplaceAll(*dzEnvFlag, "-", "_")))` — so each env is independent and the SDK's global hook never fires. That also makes the flag at line 287 decisive rather than pre-empted. Note this means the env-var half of the diff is currently a no-op: `DZ_LEDGER_RPC_URL` already worked via the SDK.

indexer/cmd/indexer/main.go
```diff
@@ -103,6 +103,12 @@ func run() error {
 
 	// Indexer configuration
 	dzEnvFlag := flag.String("dz-env", config.EnvMainnetBeta, "DZ ledger environment (devnet, testnet, mainnet-beta)")
+	// The SDK's per-env LedgerPublicRPCURL is a shared endpoint whose API key ships
+	// as a public constant, so every SDK consumer draws on one quota and lake cannot
+	// attribute or defend its own share. Point this at a dedicated endpoint to fix
+	// that. Empty keeps the SDK default. Secondary envs already have the equivalent
+	// via DZ_LEDGER_RPC_URL_<ENV>; this is the missing primary-env case.
```
ben-dz: This comment states the opposite of what the code does, which is the part a future maintainer pays for. `DZ_LEDGER_RPC_URL` is applied by the SDK for every env including the primary one (`config/env.go:188-191`), and `indexer/README.md:142` already documents it — so the primary env was not "the missing case" and was never hardwired to the shared constant. Rewrite to say that the SDK applies the bare name globally and that this flag (or a per-env name) is what scopes the redirect to one env.

indexer/cmd/indexer/main.go
```diff
@@ -270,6 +279,14 @@ func run() error {
 		return fmt.Errorf("failed to get network config: %w", err)
 	}
 
+	// Redirect every DZ ledger read for this env, not just one client: dzRPCClient
+	// (serviceability, geolocation), permissionEventsRawRPC, and the shreds client
+	// all read this one field, and a dedicated endpoint should serve all of lake's
+	// ledger traffic. Same assignment the secondary-env path makes.
+	if *dzLedgerRPCURLFlag != "" {
+		networkConfig.LedgerPublicRPCURL = *dzLedgerRPCURLFlag
```
ben-dz, line 301: Once this URL carries a real API key, that key gets written unredacted into a ClickHouse column readable through the public SQL and MCP endpoints. solana-go embeds the full endpoint URL in its error text (`rpc/jsonrpc/jsonrpc.go:552,613,619`), and `indexer/pkg/ingestionlog/writer.go:121` stores `err.Error()` verbatim into `log_ingestion_runs.error_message` (90-day TTL); the restricted read-only user behind `/api/sql/query` (`api/main.go:704`, under `OptionalAuth`, not `RequireAuth`) can read it — verified in production, where 7 of 2585 error rows already embed an endpoint URL from the SDK's public constant.

Fix: redact at the write in `ingestionlog.buildRecord`, reusing the package the slog handler already applies (`utils/pkg/logger/logger.go:37-42`) — `msg := redact.Error(err)` — with a regression test asserting a `*url.Error`-shaped RPC error is stored redacted. Restricting the public user's grants on `log_*` is a useful extra control but does not cover internal ClickHouse readers.

ben-dz (changes requested):
Two things need to change before the rollout this PR enables.

**The env var name is already owned by the SDK and applies to every env, not just the primary one.** `config.NetworkConfigForEnv` reads `DZ_LEDGER_RPC_URL` after its per-env switch (`config/env.go:188-191` in the pinned SDK), so `startSecondaryNetwork`'s call at `main.go:982` picks it up for devnet and testnet too. Setting it in `lake-secrets` will point those indexers at the mainnet endpoint with their own program IDs, returning zero accounts and stalling `lake_devnet`/`lake_testnet` behind the `refusing to write snapshot` guard. The same var is read by `api/handlers/ledger.go:59` in a pod sharing that secret. This also means the primary env was never hardwired — `DZ_LEDGER_RPC_URL` already worked, and `indexer/README.md:142` documents it; only the new flag is new behavior.

**The key will land unredacted in ClickHouse.** solana-go puts the full endpoint URL in its error text, and `indexer/pkg/ingestionlog/writer.go:121` stores `err.Error()` verbatim in `log_ingestion_runs.error_message`, which the restricted read-only user behind `/api/sql/query` and the hosted MCP can read (verified: 7 of 2585 current error rows already embed an endpoint URL). The PR defends the log sink and misses this one.

bgm-malbeclabs:
Both findings were right, and I checked them rather than taking them on trust.

**The premise of this PR was wrong.** I have rewritten the description to say so plainly. `NetworkConfigForEnv` reads `DZ_LEDGER_RPC_URL` unconditionally after the per-env switch, so the primary env was never hardwired, the README already documented it, and the env-var half of my diff was a no-op that implied a scoping the name cannot provide. The override now resolves from `DZ_LEDGER_RPC_URL_<ENV>`, matching the secondary path, so each env is independent and the SDK's global hook never fires.

Also rebased onto main, which picks up #756, #757, #760 and #761. The diff is 20 lines on one file now.

**The leak is real and live, and it is split into #764.** I verified the row counts myself rather than trusting the number:

```
total error rows               : 2543
rows with api-key=<real value> :    6
rows with api-key=redacted     :    0
rows with a keyed rpcpool path :    1
rows with rpcpool.com/redacted :    0
```

Seven rows, none redacted. Your framing was exactly right: the redaction existed and was wired into the slog handler only, so logs were clean while the table was not.

One thing worth passing on from writing that test. My first fixture used `b4dc0ffee-dead-beef-1234-567890abcdef`, which has **nine** hex characters in the first group. `redact`'s `tokenSegment` matches `8-4-4-4-12`, so it was not redacted and I briefly thought the package had a gap for path-embedded keys. It does not — the fixture was malformed. I have noted the shape requirement in the test, because that is a failure mode where the test passes for the wrong reason.

`envVarSuffix` is tested for the same class of reason: a wrong mapping means the variable is never found, the override silently never applies, and nothing reports it.

indexer/cmd/indexer/main.go
```diff
@@ -266,6 +290,14 @@ func run() error {
...
```
ben-dz, line 301: Configuring a key-bearing endpoint before #764 lands writes the key into a ClickHouse column that the public SQL and MCP query user can read. This line is inert until `DZ_LEDGER_RPC_URL_MAINNET_BETA` or `--dz-ledger-rpc-url` is set, but at that point solana-go's error text carries the full URL into `log_ingestion_runs.error_message`, which `indexer/pkg/ingestionlog/writer.go:121` still stores verbatim via `err.Error()` on both this branch and main. Merge #764 first, or hold endpoint provisioning until it lands.

PR #759, opened 2026-08-07
api,web: customer route latency page

bgm-malbeclabs:
## Summary of Changes
- Adds `/performance/routes`, a shareable page showing measured DoubleZero vs public-internet latency (mean, p95, jitter). Pick a set of locations and the page builds the **full mesh between them as a matrix**; the whole view lives in the URL so a link reproduces it exactly. Built for a prospect evaluating DoubleZero on nine specific colo-to-colo routes.
- **Matrix cells shade by improvement, not by internet RTT.** 25 metros carry activated links and the graph is connected, so a DoubleZero path exists for roughly 300 of the 435 measured pairs — "DoubleZero carries this route" is the norm, and shading by internet RTT would spend the colour range on something constant. Four exceptions stay separately worded rather than collapsing into one grey cell: no DoubleZero path, no internet samples, figures withheld, and load failed.
- **Click a cell for the full breakdown; modifier-click to compare several at once.** Cmd, Ctrl or Shift toggles a cell into a comparison set (capped at 10, batched into one series request), and the selected routes render side by side. Mirrored cells fold to one route, since the matrix is symmetric.
- `metro-path-latency` now reports **measured** RTT/p95/jitter summed along the chosen path, alongside the existing contracted figure. Routes where any hop lacks recent samples are flagged `partiallyCommitted` and fall back to the contracted figure rather than silently understating.
- Adds `GET /api/topology/route-series` — 7 days of hourly DoubleZero and internet RTT for up to 10 metro pairs, feeding an inline SVG sparkline that breaks at gaps instead of drawing through them.
- Public-internet p95 and jitter are now surfaced (`quantileExact(0.95)` and mean |IPDV|), so both columns carry all three statistics instead of DoubleZero-only.
- Models endpoints with no DoubleZero presence. Ohio (AWS us-east-2) is **anchored**: both the DoubleZero and internet figures are taken at the same on-ramp metro so the customer's access leg is common-mode and cancels, with the on-ramp user-selectable. Zurich has no committed coverage and renders **N/A** — no substitute metro, labelled or otherwise. Nothing outside the DoubleZero network is measured or modelled.
- **Behaviour change:** `/api/topology/route-series` returns 5xx on backend failure where the initial implementation returned 200 with an error body. Only the new page consumes it, but an HTTP error-rate alert would see a new source.
- `improvementPct` keeps its contracted basis so the existing `/performance/path-latency` page stays consistent; the measured-basis figure is a separate `measuredImprovementPct` field. `path-latency-page.tsx` is unchanged.

## Diff Breakdown
| Category     | Files | Lines (+/-)   | Net   |
|--------------|-------|---------------|-------|
| Core logic   |     6 | +1357 / -38   | +1319 |
| Scaffolding  |     4 | +44   / -1    |   +43 |
| Tests        |     5 | +396  / -0    |  +396 |
| **Total**    |    15 | +1797 / -39   | +1758 |

Mostly one new page and one new endpoint; the API-side changes are additive, and the only edits to shared code are the measured-latency fields and a deterministic path tie-break.

<details>

- `web/src/components/routes-page.tsx` — the page: metro/off-net pickers, facility drill-down, anchor selection, URL state, and a single `routeFigures` policy that decides what is a real measurement and what renders absent
- `api/handlers/route_series.go` — 7-day hourly series endpoint, UTC-pinned buckets, zeroes a bucket where any hop is missing rather than reporting a partial sum
- `api/handlers/isis.go` — measured per-hop RTT/p95/jitter summed along the path, path metro list, internet p95/jitter, split improvement fields
- `web/src/components/shared/sparkline.tsx` — inline SVG sparkline rendering one segment per contiguous run so a gap breaks the line
- `web/src/lib/route-anchors.ts` — off-net endpoint model and route-token parsing; rejects malformed tokens outright so a corrupted link cannot render as a plausible-but-wrong route
- `web/src/lib/api.ts` — new response types and the series client
- `api/handlers/isis_ksp.go` — deterministic device ordering so tied equal-cost paths resolve the same way on every call

</details>

## Open questions
- ~~The us-east-2 on-ramp defaults to Chicago and needs confirmation.~~ **Confirmed by Nikesh: Chicago, specifically `dz-ch2-sw01`** (activated, metro `chi`). The default anchor was already correct, so the Tokyo→Ohio headline stands at roughly 25% rather than the 12% a Pittsburgh on-ramp would have given. The picker stays user-selectable.
- The page renders inside the full app shell, so the sidebar travels with the shared link. Only geoloc, hyperliquid and permission-audit are gated on `is_internal_user` — Incidents, Ops, the Agent and `/performance/path-latency` are one click away for anyone holding the URL. Worth a deliberate decision before the link goes to a prospect.
- **Columbus lands ~14 Aug, and that retires the Ohio entry.** Once a Columbus metro exists, Ohio should stop being an off-net endpoint anchored at Chicago — otherwise the page keeps serving anchored figures while real measurements exist, which is the stale-substitute failure the Zurich N/A rule exists to prevent. Deliberately **not** automated: the retirement needs the metro code Columbus will get, which is unknown today, and a wrong guess fails silently. It is a one-line deletion of the `OFF_NET_ENDPOINTS` Ohio entry in `web/src/lib/route-anchors.ts`, and it deserves a human comparing the anchored figure against the real one rather than a silent swap under an already-shared link.
- **Nothing has been checked in a browser.** No dev server was available during implementation, so the colour ramp, cell contrast, modifier-click behaviour, the modified-Enter guard, the responsive card grid and inert-cell focus are all reasoned rather than observed. Two specific things to look at under the preview: the millisecond line uses muted text over the strongest green (weakest contrast on the page, and it lands on the best-performing cells), and `hover:brightness-110` filters cell text along with the background, which reduces contrast in light theme on the cell being pointed at.

## Testing Verification
- Verified the figures against production ClickHouse before wiring them: `lon↔tyo` internet is 259.76 ms mean / 272.05 ms p95 / 2.56 ms jitter, against sub-0.1 ms jitter on DoubleZero links — the determinism gap is the point of the page for this audience.
- Switched `quantile` to `quantileExact` after confirming the approximate variant is non-deterministic run to run; at ~300 samples/pair the exact aggregate costs ~13 ms over 1.16M rows.
- Unit tests pin the rules that a browser check would pass by luck: path orientation (the wrong direction was a per-load coin flip), sparkline index preservation across a gap, incomplete-bucket-to-zero, `partiallyCommitted` suppression, and malformed-token rejection.
- Confirmed a stale `page_cache` payload predating the new fields renders as absent rather than throwing — the worker writes that cache from a separate process, so the first deploy serves old JSON to the new page.
- Confirmed loading, request failure, and genuinely-no-path are three distinct states. All three previously rendered as "No DoubleZero path between these metros", which is a false claim about coverage on a page a customer is reading.
- `go test -race ./api/...` and 234 frontend tests pass; `golangci-lint` clean on `./api/...`.

### DIFF
--- a/api/handlers/isis.go
+++ b/api/handlers/isis.go
@@ -1687,15 +1687,20 @@ func (a *API) GetMetroConnectivity(w http.ResponseWriter, r *http.Request) {
 
 // MetroPathLatency represents path-based latency between two metros
 type MetroPathLatency struct {
-	FromMetroPK       string   `json:"fromMetroPK"`
-	FromMetroCode     string   `json:"fromMetroCode"`
-	ToMetroPK         string   `json:"toMetroPK"`
-	ToMetroCode       string   `json:"toMetroCode"`
-	PathLatencyMs     float64  `json:"pathLatencyMs"`     // Sum of link metrics along path (in ms)
-	HopCount          int      `json:"hopCount"`          // Number of hops
-	BottleneckBwGbps  float64  `json:"bottleneckBwGbps"`  // Min bandwidth along path (Gbps)
-	InternetLatencyMs float64  `json:"internetLatencyMs"` // Internet latency for comparison (0 if not available)
-	ImprovementPct    *float64 `json:"improvementPct"`    // Improvement vs internet (nil if no internet data)
+	FromMetroPK        string   `json:"fromMetroPK"`
+	FromMetroCode      string   `json:"fromMetroCode"`
+	ToMetroPK          string   `json:"toMetroPK"`
+	ToMetroCode        string   `json:"toMetroCode"`
+	PathLatencyMs      float64  `json:"pathLatencyMs"`      // Contracted: sum of link metrics along path (ms)
+	MeasuredLatencyMs  float64  `json:"measuredLatencyMs"`  // Observed: sum of per-hop rollup RTT (ms)
+	MeasuredP95Ms      float64  `json:"measuredP95Ms"`      // Sum of per-hop p95 (conservative; percentiles are not additive)
+	MeasuredJitterMs   float64  `json:"measuredJitterMs"`   // Sum of per-hop avg jitter (ms)
+	PartiallyCommitted bool     `json:"partiallyCommitted"` // true when a hop lacked samples and measured fell back to contracted
+	PathMetros         []string `json:"pathMetros"`         // metros traversed, e.g. ["tyo","fra","lon"]
+	HopCount           int      `json:"hopCount"`
+	BottleneckBwGbps   float64  `json:"bottleneckBwGbps"`
+	InternetLatencyMs  float64  `json:"internetLatencyMs"`
+	ImprovementPct     *float64 `json:"improvementPct"` // vs internet, computed from measured latency
 }
 
 // MetroPathLatencyResponse is the response for the metro path latency endpoint
@@ -1737,7 +1742,7 @@ func (a *API) GetMetroPathLatency(w http.ResponseWriter, r *http.Request) {
 	ctx, cancel := context.WithTimeout(r.Context(), 30*time.Second)
 	defer cancel()
 
-	response, err := a.FetchMetroPathLatencyData(ctx, optimize)
+	response, err := a.FetchMetroPathLatencyData(ctx, optimize, 0)
 	if err != nil {
 		logError("metro path latency error", "error", err)
 		writeJSON(w, MetroPathLatencyResponse{Optimize: optimize, Paths: []MetroPathLatency{}, Error: err.Error()})
@@ -1749,7 +1754,7 @@ func (a *API) GetMetroPathLatency(w http.ResponseWriter, r *http.Request) {
 
 // FetchMetroPathLatencyData fetches metro path latency data for the given optimization strategy.
 // Used by both the handler and the cache.
-func (a *API) FetchMetroPathLatencyData(ctx context.Context, optimize string) (*MetroPathLatencyResponse, error) {
+func (a *API) FetchMetroPathLatencyData(ctx context.Context, optimize string, window time.Duration) (*MetroPathLatencyResponse, error) {
 	start := time.Now()
 
 	g, err := a.loadTopologyGraph(ctx, "")
@@ -1765,31 +1770,52 @@ func (a *API) FetchMetroPathLatencyData(ctx context.Context, optimize string) (*
 	// Compute best paths between all metro pairs using in-memory Dijkstra
 	metroPaths := computeMetroPairPaths(g)
 
+	measured, err := a.linkMeasuredMap(ctx, window)
+	if err != nil {
+		return nil, fmt.Errorf("loading measured link latency: %w", err)
+	}
+
 	// Build map of metro paths
 	pathMap := make(map[string]*MetroPathLatency)
 	for _, mp := range metroPaths {
+		mRtt, mP95, mJitter, partial := sumMeasuredAlongPath(mp.Path.Nodes, measured, mp.Path.TotalMetric)
+		metros := pathMetroCodes(mp.Path.Nodes, g)
 		path := &MetroPathLatency{
-			FromMetroPK:      mp.FromMetroPK,
-			FromMetroCode:    mp.FromMetroCode,
-			ToMetroPK:        mp.ToMetroPK,
-			ToMetroCode:      mp.ToMetroCode,
-			PathLatencyMs:    float64(mp.Path.TotalMetric) / 1000.0, // Convert microseconds to milliseconds
-			HopCount:         mp.HopCount,
-			BottleneckBwGbps: float64(mp.BottleneckBps) / 1e9, // Convert bps to Gbps
+			FromMetroPK:        mp.FromMetroPK,
+			FromMetroCode:      mp.FromMetroCode,
+			ToMetroPK:          mp.ToMetroPK,
+			ToMetroCode:        mp.ToMetroCode,
+			PathLatencyMs:      float64(mp.Path.TotalMetric) / 1000.0, // Convert microseconds to milliseconds
+			MeasuredLatencyMs:  mRtt,
+			MeasuredP95Ms:      mP95,
+			MeasuredJitterMs:   mJitter,
+			PartiallyCommitted: partial,
+			PathMetros:         metros,
+			HopCount:           mp.HopCount,
+			BottleneckBwGbps:   float64(mp.BottleneckBps) / 1e9, // Convert bps to Gbps
 		}
 
 		// Store in map for both directions
 		key1 := mp.FromMetroCode + ":" + mp.ToMetroCode
 		key2 := mp.ToMetroCode + ":" + mp.FromMetroCode
 		pathMap[key1] = path
+		reversed := make([]string, len(metros))
+		for i, c := range metros {
+			reversed[len(metros)-1-i] = c
+		}
 		pathMap[key2] = &MetroPathLatency{
-			FromMetroPK:      mp.ToMetroPK,
-			FromMetroCode:    mp.ToMetroCode,
-			ToMetroPK:        mp.FromMetroPK,
-			ToMetroCode:      mp.FromMetroCode,
-			PathLatencyMs:    path.PathLatencyMs,
-			HopCount:         path.HopCount,
-			BottleneckBwGbps: path.BottleneckBwGbps,
+			FromMetroPK:        mp.ToMetroPK,
+			FromMetroCode:      mp.ToMetroCode,
+			ToMetroPK:          mp.FromMetroPK,
+			ToMetroCode:        mp.FromMetroCode,
+			PathLatencyMs:      path.PathLatencyMs,
+			MeasuredLatencyMs:  path.MeasuredLatencyMs,
+			MeasuredP95Ms:      path.MeasuredP95Ms,
+			MeasuredJitterMs:   path.MeasuredJitterMs,
+			PartiallyCommitted: path.PartiallyCommitted,
+			PathMetros:         reversed,
+			HopCount:           path.HopCount,
+			BottleneckBwGbps:   path.BottleneckBwGbps,
 		}
 	}
 
@@ -1824,15 +1850,23 @@ func (a *API) FetchMetroPathLatencyData(ctx context.Context, optimize string) (*
 			key2 := metro2 + ":" + metro1
 			if p, ok := pathMap[key1]; ok {
 				p.InternetLatencyMs = avgRttMs
-				if avgRttMs > 0 && p.PathLatencyMs > 0 {
-					pct := (avgRttMs - p.PathLatencyMs) / avgRttMs * 100
+				basis := p.MeasuredLatencyMs
+				if basis <= 0 {
+					basis = p.PathLatencyMs
+				}
+				if avgRttMs > 0 && basis > 0 {
+					pct := (avgRttMs - basis) / avgRttMs * 100
 					p.ImprovementPct = &pct
 				}
 			}
 			if p, ok := pathMap[key2]; ok {
 				p.InternetLatencyMs = avgRttMs
-				if avgRttMs > 0 && p.PathLatencyMs > 0 {
-					pct := (avgRttMs - p.PathLatencyMs) / avgRttMs * 100
+				basis := p.MeasuredLatencyMs
+				if basis <= 0 {
+					basis = p.PathLatencyMs
+				}
+				if avgRttMs > 0 && basis > 0 {
+					pct := (avgRttMs - basis) / avgRttMs * 100
 					p.ImprovementPct = &pct
 				}
 			}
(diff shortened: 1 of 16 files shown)


api/handlers/isis.go
```diff
@@ -3176,3 +3222,97 @@ func (a *API) GetMetroDevicePaths(w http.ResponseWriter, r *http.Request) {
...
+// in both directions, over the given window (0 defaults to 24h).
+func (a *API) linkMeasuredMap(ctx context.Context, window time.Duration) (map[string]linkMeasured, error) {
+	if window <= 0 {
+		window = 24 * time.Hour
+	}
+	query := `
+		SELECT
+			l.side_a_pk,
+			l.side_z_pk,
+			round(sum(r.a_avg_rtt_us * r.a_samples + r.z_avg_rtt_us * r.z_samples) / greatest(sum(r.a_samples + r.z_samples), 1) / 1000.0, 3) AS avg_rtt_ms,
```
ben-malbeclabs: `a_avg_rtt_us` and `a_p95_rtt_us` in `link_rollup_5m` include loss probes, which carry `rtt_us = 0`. The rollup's own query defines a loss probe as `f.loss OR f.rtt_us = 0` (`indexer/pkg/rollup/activities.go:134`) but never excludes them from the RTT aggregates, so every zero is averaged in as a very fast sample.

This is the bug `20250316000001_fix_latency_comparison_exclude_loss.sql` fixed in `dz_vs_internet_latency_comparison`, with the comment "dfw→mia showed 6.8ms instead of the real 30ms". The rollup tables landed six days later and did not carry the fix forward.

It matters more here than on an internal page for two reasons. `fact_dz_internet_metro_latency` has no `loss` column, so the internet side has no equivalent deflation and the bias is one-directional in DoubleZero's favour. And the deflation grows exactly when a link is having a bad hour, so the number flatters us most when the network is doing worst, on a page a customer is reading.

`a_loss_pct` is already in the rollup. Options, cheapest first:

- suppress or annotate a hop whose `loss_pct` is above some threshold, so a bad link stops contributing a flattering number silently
- de-bias with the loss rate we already store, `avg / (1 - loss_pct/100)`
- read raw `fact_dz_device_link_latency` with `loss = false AND rtt_us > 0` for this page only
- add loss-excluded aggregates to `link_rollup_5m` (correct, but changes the rollup for every consumer)

Whichever way, worth a comment recording that the figure is loss-inclusive, so the next reader does not have to rediscover it.

web/src/components/routes-page.tsx
```diff
@@ -0,0 +1,1452 @@
...
+      <div className="px-6 py-4">
+        <PageHeader icon={RouteIcon} title="Route Latency" />
+
+        <p className="mt-2 text-sm text-muted-foreground">
+          DoubleZero against the public internet, across your locations. Pick the locations you
+          care about and every route between them is compared.
+        </p>
+        <p className="mt-1 text-xs text-muted-foreground max-w-4xl">
+          Figures cover the last 24 hours; a route opened as a card also shows the last 7 days by
+          the hour. The public-internet figures are measured end to end. The DoubleZero p95 and jitter
```
ben-malbeclabs: This paragraph is careful about the p95 and jitter approximations but never says which link set the paths were computed over. `FetchMetroPathLatencyData` calls `loadTopologyGraph(ctx, "")`, which is algo 0: every activated link, i.e. the multicast basis.

That is the right basis for a multicast/Edge prospect, so this is a labelling fix rather than a computation one. But nothing on the page says so, and three activated links sit outside the unicast topology today, which moves 33 of 378 metro pairs by up to 77 ms. A reader who assumes these are unicast figures is reading numbers that are optimistic by that much for their use, and there is nothing here to correct them. I made exactly that mistake reviewing this PR.

Two sentences would close it:

- name the basis, e.g. "Paths follow algo 0, the link set multicast uses."
- say the figures are round-trip. Both columns are RTT, so the improvement percentage is unaffected, but a multicast reader thinks in one-way publisher-to-subscriber terms and "ms" alone invites halving or not halving at random.

Worth a note that the internet column is unicast probe RTT, since there is no public-internet multicast. That is the right comparison for this pitch, because it is what a customer replicating over unicast pays today, but a technical reader will ask.

#765 adds `service` to the API response on the reasoning that "a figure is meaningless without it", and gives `/performance/path-latency` a toggle. This page is the one where it matters most and is currently the only one that says nothing.

web/src/App.tsx
```diff
@@ -659,6 +660,7 @@ function AppContent() {
             <Route path="/performance/dz-vs-internet" element={<DzVsInternetPage />} />
             <Route path="/performance/link-latency" element={<LinkLatencyPage />} />
             <Route path="/performance/path-latency" element={<PathLatencyPage />} />
+            <Route path="/performance/routes" element={<RoutesPage />} />
```
ben-malbeclabs, line 663: The route is registered here but `sidebar.tsx` is untouched by this PR, so `/performance/routes` has no nav entry and is reachable only by someone holding the URL. Worth making that an explicit decision rather than an omission, because it cuts both ways: the open question in the description worries about the app shell travelling with a shared link, and no nav entry is one answer to that. But it also means nobody internally will find the page.

If it should be navigable, it belongs beside Path Latency in the Performance sub-nav (`sidebar.tsx:452-473`), and needs three things:

- an `isPerformanceRoutes` check alongside the existing `isPerformancePathLatency`
- the `hasNeo4j` gate that Path Latency already has, since `metro-path-latency` and `route-series` both sit behind `RequireNeo4jMiddleware` and the page is dead without them
- an icon that is not `Route`, which Path Latency already uses

One quirk worth knowing before deciding: the Performance sub-items only render when `isPerformanceRoute` is already true, so a new entry is invisible until you are on a `/performance/*` page. Adding the link does not by itself make the page discoverable from a cold start.

api/handlers/isis.go
bgm-malbeclabs: Fixed in 471038c. `linkMeasuredMap` now reads the raw samples filtered to probes that arrived, rather than the rollup.

The last option on your list, not the cheapest, for two reasons.

De-biasing with `avg / (1 - loss_pct/100)` is exact for the mean. The lost probes contribute 0 to the sum, so the sum is already the good-sample sum and only the divisor is wrong. It is not exact for p95, and it cannot express a link that loses every probe, where the divisor is zero.

That last case turned out to matter more than the deflation. Three links were losing every probe and so reported 0 ms, not a deflated figure. `sumMeasuredAlongPath` only checks `SampleCount == 0`, and a dead link has plenty of samples, all zero. A dead hop was being added to paths as a free one. Filtering the probes leaves those links out of the map entirely, so the path is reported as partial.

Reading raw costs 40 ms against 138 ms for the rollup query, over 2.9M rows for a 24h window and 38 ms for 7 days. There was no reason to model anything.

Measured on production before and after:

| link | rollup | raw, loss excluded | loss |
|---|---|---|---|
| `was001-dz002:fra001-dz002` | 78.463 | **88.469** | 11.3% |
| `dfw001-dz001:chi001-dz001` | 18.660 | **21.924** | 14.9% |
| `cherlita:swerry` | 10.502 | **13.093** | 19.8% |
| `dz-sxb-01:dz-muc-01` | 0 | dropped | 100% |

p95 barely moves (21.966 against 21.971 on the dfw-chi link), because the zeros sit at the bottom of the distribution. The mean is where the whole bias lives.

This is one function with one caller, so it fixes `/performance/path-latency` too. Two ClickHouse-backed tests cover it: lost probes excluded from the average, and a fully-lost link absent from the map. The rollup itself still has the bug for every other consumer, which is a separate change.

web/src/components/routes-page.tsx
bgm-malbeclabs: Fixed in 055374a. A second paragraph now says:

> DoubleZero routes here follow every activated link, which is the set multicast forwards over. Unicast follows a smaller set and is slower on some routes. The public-internet column is a unicast probe, because the public internet carries no multicast, and it is what you pay today to reach the same place.

The first paragraph now opens "Figures cover the last 24 hours and every one of them is round-trip".

I kept "every activated link" rather than "algo 0". This page goes to prospects, and the plain wording carries the same fact. The exact term stays on `/performance/path-latency`, where the toggle is.

web/src/App.tsx
bgm-malbeclabs: Added in d556f27, with all three things you listed: an `isPerformanceRoutes` check, the `hasNeo4j` gate, and `Waypoints` rather than `Route`.

Making it navigable is the right call. The shared-link worry is about the app shell travelling with the URL, and hiding the page from the sidebar does nothing about that. It only hides the page from us.

Your quirk holds. The entry is invisible until you are already on a `/performance/*` page, so this makes the page findable for someone who knows roughly where to look, and not from a cold start. Same as Path Latency, so leaving it is fine.

bgm-malbeclabs:
All three addressed, in three commits.

- 471038c `api: exclude lost probes from measured path latency`
- 055374a `web: name the link set and the direction on route latency`
- d556f27 `web: add route latency to the performance nav`

The loss one grew on inspection. Alongside the deflation you found, three links were losing every probe and reporting 0 ms, so a dead hop was being added to paths as a free one. Those links now drop out and the path reads partial. Details in the thread.

Note for anyone checking the exclusion dates on a preview: a preview environment bootstraps its own `dim_dz_links_history`, so every link's first snapshot is the same moment and nothing has ever been tagged inside the window. Those dates only mean anything against production.

api/handlers/isis.go
ben-malbeclabs: Agreed on all of it, and the fully-lost case is a better find than the one I reported. A dead link with plenty of zero samples sailing past `SampleCount == 0` is worse than the deflation, and raw-over-de-biasing is clearly right given `avg / (1 - loss_pct/100)` cannot express a divisor of zero.

One thing to pull back into scope before this closes:

> This is one function with one caller, so it fixes `/performance/path-latency` too. [...] The rollup itself still has the bug for every other consumer, which is a separate change.

Both true in isolation, and deferring the rollup fix for status and incident views is the right call. But `FetchRouteSeries` in `route_series.go` is not another consumer in that sense. It is a second copy of this query, in this PR, feeding the same card on the same page:

```sql
sum(r.a_avg_rtt_us * r.a_samples + r.z_avg_rtt_us * r.z_samples) / greatest(...)
FROM dz_links_current l
JOIN link_rollup_5m r FINAL ON l.pk = r.link_pk
```

So the sparkline still averages lost probes in, and the fix has made the card disagree with the chart directly beneath it. On the preview:

```
lax-mia, before 471038c   card 25.54 ms            sparkline median 25.54 ms
lax-mia, after            card 53.16 ms, withheld  sparkline median 25.54 ms, min 0.13 ms
chi-dfw, after            card 21.51 ms            sparkline median 21.51 ms   (low loss, agrees)
```

Before the commit both halves were wrong together. Now the card correctly refuses to state a measured figure while the line under it asserts one at less than half the contracted value, dipping to 0.13 ms. That is more misleading than the original deflation, because a reader sees both at once.

My comment was anchored here and never named `route_series.go`, so the scoping followed what I pointed at. The gap is partly mine.

Two smaller things on the new query itself:

**`avg(abs(f.ipdv_us))` needs the `ifNull` the internet query already has.** `ipdv_us` is `Nullable(Int64)`, so `avg` over an all-NULL group returns NULL, `round` propagates it, and the scan into `AvgJitterMs float64` fails, taking the endpoint and the page-cache refresh with it. The internet query in this file already guards this, with a comment explaining exactly why. Filtering to probes that arrived makes it more reachable rather than less: a link down to a single surviving probe has no IPDV to compute from.

**`quantile(0.95)` against `quantileExact` in 813b4a4.** That commit moved the internet p95 to the exact variant because the approximate one is not reproducible run to run. The old DoubleZero p95 came from a precomputed rollup column, so it was at least stable across reads; recomputing it per request with `quantile` is not. The two p95 figures in one row now carry different guarantees, and at 40 ms for the raw read there is room for the exact aggregate.

bgm-malbeclabs: All three fixed in 47f1010. You are right that I drew the line in the wrong place. `FetchRouteSeries` is a second copy of the query, added in this PR, feeding the same card. Calling it "another consumer" was me reading my own sentence rather than the code.

Worse than a miss, the half-fix made it louder. Before, both halves were wrong together and agreed. After, the card withheld while the line asserted 25.54 ms and dipped to 0.13 ms. A reader seeing one number is misled; a reader seeing two contradicting numbers stops trusting the page.

Both queries now share one filter, so this cannot drift again:

```go
// probeArrived filters raw latency samples down to the probes that came back,
// for a query that aliases fact_dz_device_link_latency as f. Every measured
// figure the route page shows must apply it. The card and the chart under it
// describe the same route, so a lost probe counted in one and not the other puts
// two contradicting numbers in front of one reader.
const probeArrived = `NOT f.loss AND f.rtt_us > 0`
```

New test asserts the card and the sparkline report the same number for the same route, on the same seeded probes. It fails at 48 ms against 60 ms if either side stops filtering.

**`ifNull`.** Fixed, with the same guard and reasoning as the internet query. Confirmed `ipdv_us` is `Nullable(Int64)` and `rtt_us` is plain `Int64`, so jitter is the only column that needs it. Your reachability point is the sharp end of it, and there is now a test seeding a probe that arrived with a NULL `ipdv_us` and asserting the scan survives. That test fails without the guard.

**`quantileExact`.** Switched, for the reason in 813b4a4. Two p95 figures in one row carrying different reproducibility guarantees is the kind of thing nobody discovers until they diff two screenshots. It measures at 40 ms, the same as the approximate variant, so there was nothing to weigh.

One consumer left on the rollup that I am deliberately not touching: `enrichPathsWithMeasuredLatency` at `isis.go:992`, behind `/api/topology/paths`. Same deflation, and a dead link will read 0 ms there too. It is internal, it is not in this PR, and it already shows `loss_pct` beside the figure, so a reader has the context the route page did not. Flagging it rather than widening this PR.

Merged `main` in and re-ran everything: build, lint, the ClickHouse-backed handler tests, 285 frontend tests.

